Issue #71, opened 2026-01-14
Prefix offset implementation in pi05 denoise_step


akshay18iitg:
### System Info

```Shell
using latest opentau release
```

### Information

- [x] One of the scripts in the src/opentau/scripts/ folder of OpenTau
- [ ] My own task or dataset (give details below)

### Reproduction

clone the latest opentau release

### Expected behavior

The calculation of prefix_offset in denoise_step removes the last self.discrete_action_max_length tokens. https://github.com/TensorAuto/OpenTau/blob/4ae6ca7e4dfcc8c94622b76e9fe54df669f729b0/src/opentau/policies/pi05/modeling_pi05.py#L1239.

However, while inference we dont have discrete action tokens in embed_prefix, so actually we are removing langauge tokens.

PR #72, opened 2026-01-14
Fixing calculation of prefix_offsets in denoise step of pi05 policy

akshay18iitg:
## What this does
close https://github.com/TensorAuto/OpenTau/issues/71

## How it was tested
Ran the pi05_integration gpu pytest after the change and it still passes.


shuheng-liu (approved):
Looks correct to me.

WilliamYue37 (approved):

PR #161, opened 2026-04-13
Maintaining consistency for pi05 with paligemma pretraining 

akshay18iitg:
## What this does
Adds configuration for continuous and discrete state in original pi05. 
Switched to new prompt style consistent with paligemma and pi pretraining.

## How it was tested
Continuous state libero run (99%) - https://wandb.ai/wyautox-autox/pi05/runs/j9jk5re1?nw=nwuserashah18 
Discrete state libero run (96%) -  https://wandb.ai/wyautox-autox/pi05/runs/y9kcm6y8?nw=nwuserashah18

## Checklist

- [X] My PR includes policy-related changes.
  - [X] If the above is checked: I have run the GPU pytests (pytest -m "gpu") and regression tests.


shuheng-liu (changes requested):
@akshay18iitg Is this [continuous state checkpoint](https://huggingface.co/TensorAuto/pi05_libero_continuous_state) loadable into pi05? We need to update the HuggingFace checkpoint and also update README and relevant docs.

akshay18iitg:
> @akshay18iitg Is this [continuous state checkpoint](https://huggingface.co/TensorAuto/pi05_libero_continuous_state) loadable into pi05? We need to update the HuggingFace checkpoint and also update README and relevant docs.

Should be. The state_proj weights will be trained from scratch

src/opentau/policies/pi05/modeling_pi05.py
```diff
@@ -1031,6 +1078,16 @@ def embed_prefix(
...
+        if self.config.state_type == "continuous" and state is not None:
+            state_emb = self.state_proj(state.to(dtype=_preferred_dtype()))
+            state_emb = rearrange(state_emb, "b d -> b 1 d")
+            state_mask = torch.ones(bsize, 1, dtype=torch.bool, device=state.device)
+
+            embs.append(state_emb)
+            pad_masks.append(state_mask)
+            att_masks += [0]  # full attention with images and language
+
         lang_emb = self.paligemma_with_expert.embed_language_tokens(lang_tokens)
```
shuheng-liu, line 1079: Is there a reason we insert the state token before the language tokens? Usually, the sequence is `[images, language, state, actions]`

src/opentau/policies/pi05/modeling_pi05.py
```diff
@@ -663,13 +666,39 @@ def forward(
...
+        Args:
+            batch: Batch of data containing the "state" tensor of shape (batch_size, state_dim).
+
+        Returns:
+            A tensor of shape (batch_size, max_state_dim).
+
+        Raises:
+            ValueError: If the state dimension exceeds max_state_dim.
+        """
+        state = batch["state"]
```
shuheng-liu, line 693: `assert self.config.state_type == "continuous"` here for safety

src/opentau/policies/factory.py
```diff
@@ -94,8 +89,6 @@ def make_policy_config(policy_type: str, **kwargs) -> PreTrainedConfig:
         return PI0Config(**kwargs)
     elif policy_type == "pi05":
         return PI05Config(**kwargs)
-    elif policy_type == "pi05_continuous_state":
```
shuheng-liu, line 97: For backward compatibility, how about we keep the old `pi05_continuous_state` , throw a deprecation warning, and return a `PI05Config` with `state_type` set to `continuous`?

shuheng-liu (changes requested):
A few comments. I'm mostly concerned about the ordering of tokens

src/opentau/policies/pi05/modeling_pi05.py
akshay18iitg: This is intentional. 
In discrete mode : [images, language, state, actions]
In continuous mode : [images, state, language , actions]

Major problem: in language we have "`<eos>`Actions" string appended in prompt for the model to understand when to start predicting actions. However, if we add state after language , then we will have state after this phrase and then actions, which is not ideal

src/opentau/policies/factory.py
akshay18iitg: I think so, there is no test for pi05_continuous_state model . All gpu test passes with the above code

akshay18iitg: Still, sounds ideal. Will add deprecation warning

src/opentau/policies/pi05/modeling_pi05.py
shuheng-liu: In that case let's break the language into 2 parts, similar to how William handled it here

https://github.com/TensorAuto/OpenTau/blob/9a3c83eff915611814bf683f9975c2473c43e153/src/opentau/policies/pi05_continuous_state/modeling_pi05.py#L982-L1001

akshay18iitg: The state is not part of language for continuous mode. The implementation is identical

shuheng-liu: On second thought, I just realized there is always exactly 1 state token with continuous state projection. The model should be able to figure out the start of discrete actions by itself.

akshay18iitg: I think so in william's code, `<eos>`Actions appear before state.
I think he copied the pi0 code, which does not have this problem

akshay18iitg: Ya. But its not good while inference.The model may not learn to extract appropriate info from state token as most of the focus will be on predicting first action token.
Whereas, in my implemtation, last token of 'Actions" will be responsible for predicting first action token, which is good, as its is fixed token seen during training and inference and there is no info in 'Action', just an indicator

shuheng-liu: Ok, let's do it in either
* William's old way, with `[..., "Actions:", continuous_state, *discrete_actions]`; or
* A new, cleaner way, with `[..., continuous_state, "Actions:", *discrete_actions]`.

I'm leaning toward the former because it's less work, backward compatible with his checkpoint, and should be easy enough for VLA to learn.

Including the state token between the image and the language is just weird to me – we might as do `[state, image, language, discrete_actions]` if we're going with this approach.

shuheng-liu: > Ya. But its not good while inference.The model may not learn to extract appropriate info from state token as most of the focus will be on predicting first action token.

We don't predict discrete actions during inference. We only predict continuous actions with the action diffusion model. Am I missing something?

akshay18iitg: Ya. We dont predict discrete actions during inference. But, we still pass output from vlm to action decoder. 

So, that output should have all the necessary info from states , but it will have lot of info about first discrete token as it was train like that.


Also, position of image, state and language doen't matter as its bidirectional attention between them. Only thing changes is pos emb which should'nt be a big deal.

I am ok with    [state, image, language, discrete_actions]  or [..., continuous_state, "Actions:", *discrete_actions]

shuheng-liu: OK, let's go with `[..., continuous_state, "Actions:", *discrete_actions]` then.

src/opentau/policies/factory.py
```diff
@@ -61,10 +61,6 @@ def get_policy_class(name: str) -> type[PreTrainedPolicy]:
         from opentau.policies.pi05.modeling_pi05 import PI05Policy
 
         return PI05Policy
-    elif name == "pi05_continuous_state":
-        from opentau.policies.pi05_continuous_state.modeling_pi05 import PI05ContinuousStatePolicy
-
-        return PI05ContinuousStatePolicy
```
shuheng-liu, line 67: Raise a `DeprecationWarning` and return `PI05Policy` here.

shuheng-liu (approved):

PR #166, opened 2026-04-21
perf: parallelize ffmpeg trims in segment_lerobot_dataset

bananaSnail:
## What this does

Parallelize the ffmpeg video-trim step in `segment_lerobot_dataset.py` so that
many short clips / many camera views can be trimmed concurrently instead of one
at a time. ⚡️ Performance — **5.48× end-to-end speedup** measured on a real
dataset (see "How it was tested"). Output dataset layout and per-clip trim
behavior are byte-equivalent to before; only the scheduling changes.

Previously, `_trim_video_segment` was called inline inside the per-segment loop,
so ffmpeg invocations ran strictly one-at-a-time and became the dominant
bottleneck for datasets with many short segments / many camera views.

This PR changes the flow to:

1. During the main loop, collect each `(src_video_path, dst_video_path, start, end)`
   trim into a `ffmpeg_jobs` list instead of executing immediately. Parquet
   writing, stats, and metadata are unchanged and still written in-order.
2. After all parquet/stats/metadata are finalized, dispatch the collected trims
   through a `ThreadPoolExecutor` and show a `tqdm` progress bar over
   `as_completed(...)`. Threads are sufficient because `_trim_video_segment`
   only invokes `subprocess.run(ffmpeg, ...)` (I/O + external-process bound),
   so the GIL is not an issue.
3. Worker count is picked by `_ffmpeg_trim_max_workers()`:
   - If env var `TUNER_FFMPEG_MAX_WORKERS` is set to an integer in `[1, 64]`,
     use it.
   - Otherwise default to `min(16, (os.cpu_count() or 4) * 2)`.

### Benchmark

Tested on a real dataset, 2 source episodes with 1 AV1 1080p camera stream each
(`observation.images.camera`), segmented into 20 non-overlapping ranges per
episode for a total of **40 output episodes / 40 ffmpeg trims**.

**Environment**

- CPU: 12th Gen Intel(R) Core(TM) i7-12700K, 20 cores / 20 threads
- ffmpeg 4.2.7 (Ubuntu 20.04)
- Input: LeRobot v2.1, 2 episodes, 12,523 frames total, AV1 @ 1920×1080, 50 fps
- Segment plan: episode 0 (7,723 frames) → 20 segments of 386–387 frames;
  episode 1 (4,800 frames) → 20 segments of 240 frames

**Wall-clock**

| Run | `TUNER_FFMPEG_MAX_WORKERS` | ffmpeg-trim wall-clock | Total wall-clock |
|---|---:|---:|---:|
| Serial (baseline) | `1` | 44m 11s | **44m 20s** (2,660s) |
| Parallel (default) | `16` (auto = `min(16, cpu*2)`) | 7m 59s | **8m 05s** (485s) |

**Speedup: 5.48× end-to-end** on this workload.

The speedup is below the ideal 16× for two CPU-bound reasons, neither of which
is caused by this PR: (1) AV1 software decoding fully saturates all 20 cores
once more than ~8 ffmpeg processes run concurrently, so the bottleneck shifts
from per-trim I/O waits to shared decoder CPU; (2) the existing
`trim=start_frame:end_frame` filter decodes the source from frame 0 up to
`start_frame` for each segment (the 20th serial clip alone took 119s), and that
decoding work does not shrink under parallelism — it is only amortized across
cores. 5.48× is consistent with what we'd expect when the workload is dominated
by shared AV1 decode CPU.

**Correctness**

- Both runs produced **40 parquet files + 40 mp4 files** under
  `data/chunk-000/` and `videos/chunk-000/observation.images.camera/`.
- `ffprobe -count_frames` on a sample (episodes 0, 10, 20, 30, 39) returned
  identical frame counts between the serial and parallel outputs, and they
  matched the plan exactly (386, 386, 240, 240, 240).
- `episodes.jsonl` lengths match the ffprobe-reported frame counts.

```bash
git fetch origin paralle_ffmpeg_video_trimming
git checkout paralle_ffmpeg_video_trimming

# Parallel (new default)
python src/opentau/scripts/segment_lerobot_dataset.py \
    <input_dataset> <output_dataset> <segments.json>

# Serial baseline, for comparison
TUNER_FFMPEG_MAX_WORKERS=1 python src/opentau/scripts/segment_lerobot_dataset.py \
    <input_dataset> <output_dataset> <segments.json>
```

## Checklist

### DIFF
--- a/src/opentau/scripts/segment_lerobot_dataset.py
+++ b/src/opentau/scripts/segment_lerobot_dataset.py
@@ -29,6 +29,7 @@ An example of segments.json can be found in `configs/examples/segments.json`.
 
 import argparse
 import json
+import logging
 import math
 import os
 import shutil
@@ -54,6 +55,10 @@ from opentau.datasets.utils import (
     write_json,
 )
 
+logger = logging.getLogger(__name__)
+
+FFMPEG_MAX_WORKERS_ENV_VAR = "OPENTAU_FFMPEG_MAX_WORKERS"
+
 
 def parse_args() -> argparse.Namespace:
     """Parse command-line arguments for dataset segmentation.
@@ -69,6 +74,17 @@ def parse_args() -> argparse.Namespace:
     parser.add_argument("input_root", type=Path, help="Path to source LeRobot dataset root.")
     parser.add_argument("output_root", type=Path, help="Path to output dataset root (must not exist).")
     parser.add_argument("segments_json", type=Path, help="Path to JSON segmentation plan.")
+    parser.add_argument(
+        "--ffmpeg-max-workers",
+        type=int,
+        default=None,
+        help=(
+            "Override the worker count used to dispatch parallel ffmpeg trim jobs. "
+            "Must be an integer in [1, 64]. When unset, falls back to the "
+            f"{FFMPEG_MAX_WORKERS_ENV_VAR} environment variable, then to "
+            "min(16, (os.cpu_count() or 4) * 2)."
+        ),
+    )
     return parser.parse_args()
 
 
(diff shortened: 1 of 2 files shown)


shuheng-liu (commented):
## Review

Nice, self-contained perf win — the pattern (collect jobs during the in-order main loop, then drain them through a `ThreadPoolExecutor` after stats/parquet are written) is clean and the correctness argument is convincing. The benchmark table is helpful. Only one non-nit below; the rest are suggestions.

### Should fix before merge

- **Env var name breaks project convention.** `TUNER_FFMPEG_MAX_WORKERS` uses a `TUNER_` prefix that appears nowhere else in the repo. Existing env vars follow `OPENTAU_*` / `HF_OPENTAU_*` (`src/opentau/constants.py:52` → `HF_OPENTAU_HOME`; `tests/utils.py:31` → `OPENTAU_TEST_DEVICE`). Please rename to something like `OPENTAU_FFMPEG_MAX_WORKERS` so users can discover it and it matches surrounding conventions. Since this is a brand-new user-facing knob, renaming now is free.

- **Silent fallback on invalid values is user-hostile.** `_ffmpeg_trim_max_workers` accepts only `raw.isdigit()` in `[1, 64]` and otherwise silently uses the default. So `OPENTAU_FFMPEG_MAX_WORKERS=0`, `=100`, `=-1`, `=abc`, or a typo'd value all get silently ignored. A user setting the knob and seeing "Trimming videos (16 workers)" will assume their override applied. At minimum, emit a warning (or raise `ValueError`) when the env var is set but not accepted. Same applies to leading `+` signs, which `isdigit()` rejects.

### Suggestions

- **Prefer a CLI flag to an env var.** The script already uses `argparse` (`parse_args`), and other parallel code paths in this codebase expose worker count as a normal parameter (`src/opentau/datasets/v21/convert_stats.py:81` → `num_workers: int = 0`, `src/opentau/scripts/eval.py:804` → config-driven `max_parallel_tasks`). A `--ffmpeg-max-workers` argument would be more discoverable, show up in `--help`, and be consistent with the rest of the repo. The env var can remain as a secondary override if you want CI/scripting control.

- **`_trim_video_segment_job` is an unnecessary indirection.** The docstring says the adapter is needed "so the trim calls can be submitted to a pool without capturing local variables from the main loop", but `pool.submit(_trim_video_segment, *job)` would work identically and with no closure capture. I'd drop the wrapper (and its docstring) and submit directly:

  ```python
  futures = [pool.submit(_trim_video_segment, *job) for job in ffmpeg_jobs]
  ```

  That also shrinks the diff and removes a call layer from stack traces.

- **Docstring justification contradicts the benchmark.** `_ffmpeg_trim_max_workers` explains the `* 2` over-subscription by claiming workers are "I/O bound" and "spend most of [their] time waiting on an external ffmpeg subprocess". But the PR description's own analysis says the real workload is CPU-bound (AV1 software decode saturates cores, 5.48× vs. ideal 16×). For stream-copy trims the I/O-bound framing holds, but for the decode-heavy path here it doesn't. Consider rewording to "each worker is an external subprocess, so Python threads are fine (GIL is released); the `* 2` and cap of 16 are a pragmatic default — tune via `OPENTAU_FFMPEG_MAX_WORKERS` for CPU-heavy codecs."

  (Related: with 16 concurrent ffmpegs each spinning up their own internal decode threads, you will over-subscribe any box. Worth a future follow-up to pass `-threads N` per-invocation when scaling via the pool, but out of scope here.)

- **Failure semantics on the first bad trim.** When `fut.result()` raises inside the `as_completed` loop, the `with ThreadPoolExecutor(...)` block still waits for every in-flight ffmpeg to finish before propagating (threads can't be interrupted). This means after an error the user can sit for a minute or two watching the progress bar before the traceback appears. If you want to fail faster, you can `pool.shutdown(wait=False, cancel_futures=True)` in an `except` branch before re-raising — it cancels the queued futures while still waiting on the in-flight ones. Optional; current behavior is correct, just slow-to-fail.

### Test coverage

- No new tests for the added code paths. `_ffmpeg_trim_max_workers` in particular is trivial to unit-test and exercises every branch of the env-var parsing (valid, out-of-range, non-digit, unset, `os.cpu_count() is None`). The existing `test_trim_video_segment_uses_frame_range_filter` still passes but doesn't cover parallel dispatch; a small test that mocks `_trim_video_segment` and asserts `segment_dataset` submits N jobs and calls `fut.result()` on each would catch regressions where someone forgets the `fut.result()` (which is what surfaces worker exceptions today).

### Things that look right

- Deferring trims until after parquet/stats/metadata finalize — correct and keeps the in-order invariants intact. `episodes.jsonl` is still appended in-order after the pool drains.
- `dst_video_path.parent.mkdir(...)` happens in the main loop, not inside workers, so no concurrent-mkdir race.
- `_trim_video_segment` already includes `src_video_path` and the frame range in its `RuntimeError`, so the "which segment?" context isn't lost despite the delayed failure.
- `subprocess.run` is thread-safe and each call has its own pipes — no interleaved stderr.
- Jobs list memory is trivial even at 100k segments.

Overall, happy to approve once the env-var naming is fixed (or replaced by a CLI flag); everything else is polish.

---

bananaSnail:
Thanks for the thorough review — pushed a follow-up commit addressing every point.

**Must-fix**
- **Env var renamed** to `OPENTAU_FFMPEG_MAX_WORKERS`. Exported as `FFMPEG_MAX_WORKERS_ENV_VAR` so tests and --help share the same source of truth.
- **No more silent fallback.** The env-var path now parses with `int()` (which also fixes the `+4`-rejected-by-`isdigit()` case you flagged) and logs a `WARNING` on either non-integer or out-of-range values. The new `--ffmpeg-max-workers` CLI flag is stricter: invalid overrides raise `ValueError` rather than falling back, since a user-specified number should never be silently dropped.

**Suggestions applied**
- **CLI flag added.** `--ffmpeg-max-workers INT` is plumbed through `segment_dataset(ffmpeg_max_workers=...)` and shows up in `--help`. Env var remains as a secondary override for CI.
- **_trim_video_segment_job removed.** The pool now does `pool.submit(_trim_video_segment, *job)` directly.
- **Docstring rewritten.** No longer claims workers are I/O-bound; explains that each worker is an external subprocess (GIL released), `*2` / cap 16 is a pragmatic default, and CPU-heavy codecs (AV1) are the tuning case via the flag / env var.
- **Fail-fast on first error.** Wrapped the `as_completed` loop in try/except BaseException: `pool.shutdown(wait=False, cancel_futures=True); `raise so queued trims are cancelled immediately; in-flight ffmpeg subprocesses still finish (unavoidable — threads can't interrupt subprocesses).

**Follow-up (out of scope here, per your note)**
- Per-invocation `-threads` N to prevent ffmpeg internal threads from over-subscribing the box when scaling via the pool. Follow-up tracked in #172 

**Tests**
Added 10 new test functions (29 parametrized cases total) in
`tests/datasets/test_segment_lerobot_dataset.py`:

- `_ffmpeg_trim_max_workers` branch coverage (7 functions): valid env values
  including `+4`; invalid values (non-integer, out-of-range, empty/whitespace)
  fall back with a logged warning; unset → CPU default; cap at 16; `os.cpu_count()`
  returning `None`; override precedence over env var; invalid overrides raise
  `ValueError`.
- `segment_dataset` parallel dispatch (3 functions): asserts N jobs submitted
  (one per segment × video_key), worker exceptions propagate via `fut.result()`
  (regression guard), and `ffmpeg_max_workers` reaches `ThreadPoolExecutor(max_workers=...)`
  via a spy subclass.

​```
$ pytest tests/datasets/test_segment_lerobot_dataset.py -q
37 passed, 172 warnings in 0.22s
​```

PTAL.

shuheng-liu (approved):

PR #168, opened 2026-04-22
feat: attach metadata and expose optional standard data format keys

shuheng-liu:
## What this does

Adds segment-level metadata annotation for LeRobot v2.1 datasets plus six new optional keys on the standard data format.

- New `opentau.scripts.attach_metadata` — a copy-only CLI that annotates a LeRobot v2.1 dataset with per-frame `response` / `memory` / `mistake` parquet columns and episode-level `quality` / `segments` fields in `episodes.jsonl`. Annotation JSON is validated fail-fast before any file is written; memory is generated via `pi_mem_data_generator` (or filled with a per-frame placeholder under `--skip-memory`).
- `__getitem__` now emits `memory`, `next_memory`, `speed`, `mistake`, `quality`, and `subgoal0..N-1`, each with an `{key}_is_pad` sibling so the default PyTorch collate handles everything. Six new `DatasetMixtureConfig` probabilities (`history_state_drop_prob`, `subgoal_drop_prob`, `subgoal_end_of_segment_prob`, `response_drop_prob`, `metadata_drop_all_prob`, `metadata_drop_each_prob`) drive training-time dropout. Legacy datasets emit every new key with `_is_pad=True`, so nothing regresses.
- Adds a new tutorial (`docs/source/tutorials/attach_metadata.rst`), extends the concepts doc with the optional-keys section, and ships a runnable example annotation at `configs/examples/attach_metadata_annotations.json`.

## How it was tested

- Added `tests/scripts/test_attach_metadata.py` — 20 unit tests covering validation, per-frame column builders, `info.json` / `episodes.jsonl` updates, plus a `@pytest.mark.slow` integration test that downloads `lerobot/droid_100@v2.1`, annotates a copy via `--skip-memory`, reloads via `make_dataset`, and asserts every optional key has the expected shape and value range.
- Added `tests/datasets/test_optional_keys.py` — 10 tests for `BaseDataset._emit_optional_keys` covering the legacy (no-annotations) path, all-probs-zero, and forced-drop paths (including "subgoal drop implies no response drop").
- Added `TestDatasetMixtureOptionalKeyDropProbs` in `tests/datasets/test_dataset_mixture.py` — 19 parametrized tests covering defaults, bounds, and boundary values.
- `pytest -m "not gpu and not slow" tests/` → 516 pass; no existing tests regress.
- `pytest -m slow tests/scripts/test_attach_metadata.py::test_attach_metadata_end_to_end_droid_100` → passes locally against the real Hub.
- `pre-commit run --files ...` on every changed file → clean.

```bash
# Fast unit + config tests (no network)
pytest -sx tests/scripts/test_attach_metadata.py \
          tests/datasets/test_optional_keys.py \
          tests/datasets/test_dataset_mixture.py::TestDatasetMixtureOptionalKeyDropProbs

# Slow end-to-end (downloads lerobot/droid_100@v2.1)
pytest -sx -m slow tests/scripts/test_attach_metadata.py::test_attach_metadata_end_to_end_droid_100

# Manual smoke: annotate an existing v2.1 dataset with the bundled example
python -m opentau.scripts.attach_metadata \
    --dataset   /path/to/source/dataset \
    --annotations configs/examples/attach_metadata_annotations.json \
    --copy-to   /tmp/dataset_annotated \
    --skip-memory
```

## Checklist


shuheng-liu (commented):
## Code Review

### Overview
Adds a copy-only CLI (`opentau.scripts.attach_metadata`) that annotates a LeRobot v2.1 dataset with per-segment `response`/`memory`/`mistake` parquet columns plus episode-level `quality`/`segments` in `episodes.jsonl`. On the consumer side, `BaseDataset.__getitem__` now emits six new optional keys (`memory`, `next_memory`, `speed`, `mistake`, `quality`, `subgoal0..N-1`), each with an `{key}_is_pad` sibling, gated by six new `DatasetMixtureConfig` dropout probabilities. Legacy datasets continue to load with every new key marked `_is_pad=True`.

### Strengths
- **Strict fail-fast annotation validator** covers schema, coverage, monotonicity, bool-vs-int distinctions (nice touch rejecting float `start` and stringy `success`). Test coverage is thorough (20 validation tests + 10 emit tests + 19 drop-prob tests + a slow end-to-end against `lerobot/droid_100@v2.1`).
- `_is_pad` convention lets the default PyTorch collate handle mixed-provenance batches — `TestDefaultCollate` proves this.
- Validation-subset dropout is suppressed by default and made explicit via the shared `enable_optional_key_dropout` flag.
- The `monkeypatch` fix in `tests/datasets/test_delta_timestamps.py` catches a real global-state leak from the previous `type(dataset).fps = property(...)` pattern — good find.
- Tutorial doc, runnable example JSON, and an updated concepts section all ship together.

### Issues & Suggestions

#### High priority

1. **Subgoal video decoding happens before the drop roll.** `LeRobotDataset.__getitem__` calls `_load_subgoal_frames` (which does `_query_videos` per camera slot) *before* `_emit_optional_keys` rolls `subgoal_drop_prob`. With the default `subgoal_drop_prob=0.75`, ~75% of decodes are thrown away. Roll the drop first (in `__getitem__` or at the top of `_load_subgoal_frames`) and short-circuit when masked. See `src/opentau/datasets/lerobot_dataset.py:1682` / `:1786`.

2. **`assert not (self.root / "subgoals").is_dir()` is too aggressive.** This runs in `LeRobotDataset.__post_init__` and will hard-fail any dataset that happens to have a `subgoals/` folder for unrelated reasons. Either (a) downgrade to a warning, (b) reserve a namespaced name (`.opentau-subgoals/`), or (c) only check when a future config flag opts in. `src/opentau/datasets/lerobot_dataset.py:1253`.

3. **`history_state_drop_prob=0.3` and `subgoal_drop_prob=0.75` are aggressive defaults.** Dropping `state` on 30% of samples (and zeroing historical camera frames when `n_obs_history>1`) is a large behavior change buried in config defaults — any policy training that doesn't explicitly set these will see materially different inputs. Consider lowering defaults to 0 and requiring per-policy opt-in, or flag this migration in the PR description. `src/opentau/configs/default.py:205-210`.

#### Medium priority

4. **`speed` semantics are surprising and inconsistent.** The field is named "speed" but computed as `int(round(episode_length / 500) * 500)` — an episode-length bucket in frames. Short episodes (<250 frames) round to 0. Consider renaming to `episode_length_bucket` or documenting the formula/unit prominently. `src/opentau/datasets/lerobot_dataset.py:1785`, `docs/source/concepts.rst:135`.

5. **`speed` is never padded in `LeRobotDataset`.** Unlike `mistake`/`quality`/`memory`, it's derived purely from `episode_length` and populated unconditionally. The concepts doc frames it as an "optional key when the dataset has been enriched with segment metadata" — but it's emitted non-pad for legacy datasets too. Clarify in the doc.

6. **`BaseDataset.__init__` sets `enable_optional_key_dropout = True` by default.** Any code that instantiates a dataset outside `make_dataset`'s train/val split (inference, eval harnesses, notebooks) will silently do dropout. Either default to `False` and have the train factory flip it on for the train subset, or document this as a contract on `BaseDataset`. `src/opentau/datasets/lerobot_dataset.py:661`.

7. **Memory column preload loads the entire column.** `self.hf_dataset.with_format("arrow").select_columns(["memory"]).to_pylist()` materializes every row into a Python string at init, even though the only rows we need are at segment starts. For multi-million-frame datasets this is wasted RAM. Use `hf_dataset.select(indices)` (or `pa.Table.take`) to read only segment-start rows. `src/opentau/datasets/lerobot_dataset.py:1279-1284`.

8. **`test_subgoal_end_of_segment_roll_stays_active` doesn't verify its claim.** With `num_cams=0`, `_load_subgoal_frames` returns `{}` at line 1683 *before* any `torch.rand` call. The test only proves the early-return path, not that `subgoal_end_of_segment_prob` is read when dropout is disabled. A stronger test would patch `torch.rand` and assert it's called, or use `num_cams>=1` with a stubbed `_query_videos`. `tests/datasets/test_optional_keys.py:278-296`.

9. **`val_underlying = copy.copy(dataset)` is fragile.** A shallow copy relies on all non-divergent state being reachable by reference (meta, hf_dataset, episode_data_index). If a future refactor adds an attribute that *should* diverge, this silently shares it. Consider a helper like `dataset.with_dropout_disabled()` that makes the contract explicit, or at least a `# SHARED_STATE:` comment listing what must stay shared. `src/opentau/datasets/factory.py:231-233`.

#### Low priority

10. **Asymmetric cache population.** `episode_lengths` and `segment_starts_by_episode` are built from `self.meta.episodes.items()` (all episodes), but `segment_memories_by_episode` only populates `self.episodes` (the filtered subset). Not a bug (only `self.episodes` is reachable via `__getitem__`), but inconsistent. Pick one and apply it uniformly. `src/opentau/datasets/lerobot_dataset.py:1261-1288`.

11. **`DATA_FEATURES_NAME_MAPPING[_TEST_MAPPING_KEY] = {...}` at module scope** in `test_optional_keys.py` mutates a global dict without teardown. Consider a fixture + `monkeypatch.setitem` so runs are order-independent. `tests/datasets/test_optional_keys.py:32`.

12. **`_sample_subgoal_frame` edge cases aren't unit-tested.** The slow integration test covers the happy path. A focused test for: frame at segment end, frame at episode end, window_frames beyond segment end, would catch regressions cheaply.

13. **Validator doesn't reject empty `subtask` strings.** `{"subtask": ""}` passes validation. Probably fine, but worth an explicit decision — an empty subtask propagates into the `response` column.

#### Nits
- Tutorial CLI table says `--model` is "overridable via `OPENAI_MODEL`" but I don't see that env-var lookup in `parse_args`. Either wire it up or drop the claim. `docs/source/tutorials/attach_metadata.rst:141`.
- Section-marker comments in `_emit_optional_keys` are mostly helpful, but some (`# change data naming to standard data format` above) could be trimmed.

### Summary
Correctness-wise this is safe to merge (legacy datasets flow through the `_is_pad=True` path, and tests cover that). The real concerns are (a) decoding subgoal videos that get dropped (#1) and (b) aggressive default drop probabilities silently changing downstream training runs (#3). I'd hold on those two before merging.

---

src/opentau/datasets/lerobot_dataset.py
```diff
@@ -758,13 +792,124 @@ def _to_standard_data_format(self, item: dict) -> dict:
...
+            raw = item.get(raw_key)
+            drop_this = drop_meta_all or _roll(self.metadata_drop_each_prob)
+            if raw is None or drop_this:
+                zero = False if dtype is torch.bool else 0
+                standard_item[key] = torch.tensor(zero, dtype=dtype)
+                standard_item[f"{key}_is_pad"] = torch.tensor(True)
+            else:
+                value = bool(raw) if dtype is torch.bool else int(raw)
+                standard_item[key] = torch.tensor(value, dtype=dtype)
+                standard_item[f"{key}_is_pad"] = torch.tensor(False)
```
akshay18iitg, line 925: create a single key for metadata in standard_item

src/opentau/datasets/lerobot_dataset.py
```diff
@@ -1102,6 +1247,46 @@ def __init__(
         ep_data_index_np = {k: t.numpy() for k, t in self.episode_data_index.items()}
         check_timestamps_sync(timestamps, episode_indices, ep_data_index_np, self.fps, self.tolerance_s)
 
+        # Subgoal images attached from a dedicated subfolder are a future feature;
+        # for now we refuse to silently ignore one so users can't be confused by
+        # stale subgoals that don't influence the loaded sample.
+        assert not (self.root / "subgoals").is_dir(), (
+            f"'subgoals/' subfolder found at {self.root / 'subgoals'}. Loading subgoals "
+            "from disk is not implemented yet; please remove or rename the folder."
+        )
```
akshay18iitg: The subgoals image paths should be grabbed from metadata/info.json file.

src/opentau/datasets/lerobot_dataset.py
```diff
@@ -1431,6 +1631,101 @@ def _query_videos(self, query_timestamps: dict[str, np.ndarray], ep_idx: int) ->
...
+        ts = subgoal_frame / self.fps
+        out: dict[str, torch.Tensor] = {}
+        for k in range(self.num_cams):
+            vid_key = name_map.get(f"camera{k}")
+            if vid_key is None or vid_key not in self.meta.video_keys:
+                continue
+            frames = self._query_videos({vid_key: np.array([ts])}, ep_idx)
+            out[f"subgoal{k}_raw"] = frames[vid_key]
+        return out
+
```
akshay18iitg, line 1728: This function should load subgoal images when available in dataset, else i should load from camera videos.

Rn , its just loading from camera videos

shuheng-liu: I haven't implemented loading world-model-generated subgoals yet.

akshay18iitg (changes requested):

src/opentau/datasets/lerobot_dataset.py
akshay18iitg: ok

akshay18iitg (approved):

PR #169, opened 2026-04-22
feat: per-dataset validation loss reporting

shuheng-liu:
## What this does

Implements per-dataset validation loss reporting so training runs with multiple datasets can track each dataset's validation metrics independently rather than only the aggregate.

- Adds `WeightedDatasetMixture.get_per_dataset_dataloaders()`, which returns one sequential `DataLoader` per underlying dataset (empty datasets skipped).
- In `train.py`, the validation step now iterates each per-dataset loader in sequence, updating a per-dataset `MetricsTracker` and an aggregate tracker at the same time. Each step emits both `Validation/<name>/Loss` (plus MSE/CE/L1/Accuracy) and the existing `Validation/Loss` aggregate scalars.
- Uses each `DatasetConfig`'s `repo_id` / `vqa` identifier as the logged dataset name (with `#<idx>` disambiguation when duplicated), falling back to the old `ClassName_i` scheme when the config list cannot be lined up with the datasets (e.g. unit tests that construct a mixture directly).

This changes the validation semantics from "sample `len(concat)` times via hierarchical weighted sampling with replacement" to "fully iterate each val subset once", which is the conventional choice for validation and is what makes per-dataset reporting meaningful.

## How it was tested

- Verified the name-derivation logic in isolation across unique, duplicate, VQA, mismatched-length, and missing-config cases.
- `python -c "import ast; ast.parse(...)"` on both modified files.
- Existing `tests/datasets/test_dataset_mixture.py` assertions on `len(mixture.dataset_names)` still hold (the public shape is unchanged).
- Full GPU validation in a real run is still pending (no GPU available in this environment).

```bash
pytest -sx tests/datasets/test_dataset_mixture.py
```

Then launch any training config with `val_freq > 0` and multiple datasets in `dataset_mixture.datasets`, and confirm wandb shows both the aggregate `Validation/*` scalars and one `Validation/<repo_id>/*` group per dataset.

## Checklist

https://claude.ai/code/session_019zJCa5a31uULbXPrFjtiTr

### DIFF
--- a/src/opentau/scripts/train.py
+++ b/src/opentau/scripts/train.py
@@ -327,7 +327,7 @@ def train(cfg: TrainPipelineConfig):
         if is_val_step:
             policy.eval()
 
-            def _make_val_tracker() -> MetricsTracker:
+            def _make_val_tracker(current_step: int = step) -> MetricsTracker:
                 return MetricsTracker(
                     cfg.batch_size * accelerator.num_processes,
                     {
@@ -337,7 +337,7 @@ def train(cfg: TrainPipelineConfig):
                         "l1_loss": AverageMeter("val_l1_loss", ":.3f"),
                         "accuracy": AverageMeter("val_accuracy", ":.3f"),
                     },
-                    initial_step=step,
+                    initial_step=current_step,
                 )
 
             agg_tracker = _make_val_tracker()
@@ -373,10 +373,7 @@ def train(cfg: TrainPipelineConfig):
                             .item()
                         )
                         ce_loss = (
-                            accelerator.gather_for_metrics(losses["CE"])
-                            .to(dtype=torch.float32)
-                            .mean()
-                            .item()
+                            accelerator.gather_for_metrics(losses["CE"]).to(dtype=torch.float32).mean().item()
                         )
                         l1_loss = (
                             accelerator.gather_for_metrics(losses.get("L1", zero))
--- a/src/opentau/scripts/train.py
+++ b/src/opentau/scripts/train.py
@@ -211,10 +211,19 @@ def train(cfg: TrainPipelineConfig):
 
     if cfg.val_freq > 0:
         train_dataloader = train_dataset.get_dataloader()
-        val_dataloader = val_dataset.get_dataloader()
-        policy, optimizer, train_dataloader, val_dataloader, lr_scheduler = accelerator.prepare(
-            policy, optimizer, train_dataloader, val_dataloader, lr_scheduler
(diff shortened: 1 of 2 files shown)


src/opentau/scripts/train.py
```diff
@@ -317,65 +326,93 @@ def train(cfg: TrainPipelineConfig):
...
+                            .item()
+                        )
+
+                        if accelerator.is_main_process:
+                            for tracker in (ds_tracker, agg_tracker):
+                                tracker.loss = loss
+                                tracker.mse_loss = mse_loss
+                                tracker.ce_loss = ce_loss
+                                tracker.l1_loss = l1_loss
+                                tracker.accuracy = accuracy
```
akshay18iitg, line 397: This increases the validation time as now we have to validate for each dataset. Eventualy , lead to longer training time for same number of steps.

Can we divide the overall batch size of validation into equal parts for each datasets. This should roughly keep the validation time same as before.

akshay18iitg (changes requested):

src/opentau/scripts/train.py
shuheng-liu: Thanks for the review! I actually don't think this changes the validation cost meaningfully — the previous code wasn't doing a "small sample" of the val set either.

`WeightedDatasetMixture.get_dataloader()` sets `num_samples = len(self.concatenated_dataset) = sum(len(ds) for ds in self.datasets)` on the `HierarchicalSampler` ([dataset_mixture.py:487](https://github.com/TensorAuto/OpenTau/blob/main/src/opentau/datasets/dataset_mixture.py#L487)). So the old validation already iterated that many samples — just drawn with replacement, weighted by `dataset_weights`.

In the new code each per-dataset loader iterates its dataset fully, and the totals add up to the same `sum(len(ds))`. So the number of batches — and therefore wall-clock time — is approximately unchanged. What *does* change is the distribution: before, a high-weight / small dataset was heavily oversampled while a low-weight / large dataset was under-seen; now every val sample in every dataset is visited exactly once, which is what makes per-dataset losses meaningful.

Re: dividing the overall batch into equal parts per dataset — I'd push back on that. With N datasets that scheme would give each per-dataset metric only `len(concat) / N` samples (and for the aggregate we'd see `N×` fewer val samples than before), which would make per-dataset losses quite noisy. Happy to iterate if you think that tradeoff is worth it, but I lean toward keeping the full pass.

---

akshay18iitg: ok. looks good.

akshay18iitg (approved):

Issue #170, opened 2026-04-22
V-JEPA2 video encoder receives unnormalized [0,1] pixels — expects ImageNet normalization


shuheng-liu:
## Summary

The V-JEPA2 video encoder in `src/opentau/policies/pi05_mem` receives raw `[0, 1]` pixel values during training and inference, but V-JEPA2 expects ImageNet-normalized inputs. This is a silent numerical mismatch — the encoder still runs, but its activation statistics are far from what the pretrained weights were exposed to, likely degrading downstream task performance.

## Repro / Evidence

`VJEPA2VideoProcessor` from the same repo the encoder loads weights from reports ImageNet normalization as standard:

```python
>>> from transformers import VJEPA2VideoProcessor
>>> proc = VJEPA2VideoProcessor.from_pretrained("facebook/vjepa2-vitl-fpc64-256")
>>> proc.do_normalize, proc.image_mean, proc.image_std
(True, [0.485, 0.456, 0.406], [0.229, 0.224, 0.225])
```

But nothing in the pi05_mem pipeline applies this normalization:

- `src/opentau/policies/pi05_mem/configuration_pi05.py:95` sets `"VISUAL": NormalizationMode.IDENTITY`.
- `src/opentau/policies/normalize.py:213-214` — `IDENTITY` is a no-op (`continue`).
- `src/opentau/policies/pi05_mem/modeling_pi05.py:696-704` — `prepare_videos` only resizes and zeroes padded slots.
- `src/opentau/policies/pi05_mem/video_encoder.py:99` — docstring (pre-migration) claimed "pixel values, ImageNet-normalized", but no caller in the repo enforces this. `factory.py:212-216` registers IMAGENET_STATS on `dataset.meta.stats`, however these are only consumed by `Normalize` when the corresponding feature's `NormalizationMode` is `MEAN_STD` — which it isn't for `VISUAL`.

So V-JEPA2 was processing inputs in `[0, 1]` range instead of roughly `(x - mean) / std` (≈ `[-2.12, 2.64]`).

## Impact

The encoder was frozen by default (`freeze_vision_encoder=True`), so its pretrained statistics governed feature quality. Feeding data outside the distribution the weights were trained on degrades representations and likely suppresses any benefit the V-JEPA2 encoder could provide over a from-scratch vision tower.

## Status

[...]

PR #171, opened 2026-04-22
feat: replace V-JEPA2 with space-time SigLIP video encoder in pi05_mem

shuheng-liu:
## What this does

Rewrites the `pi05_mem` low-level memory video encoder to follow the MEM paper's actual architecture (Torne, Pertsch, Walke et al., "MEM: Multi-Scale Embodied Memory for Vision Language Action Models", Section III-C + Appendix C): a standard SigLIP ViT (from PaliGemma) extended with space-time separable attention every 4th layer. The prior V-JEPA2 + Perceiver implementation is replaced.

**Architectural highlights**
- **No new learnable parameters.** The temporal sublayer at every 4th of the 27 SigLIP layers (indices `[3, 7, 11, 15, 19, 23]`) reuses that layer's own `q_proj`/`k_proj`/`v_proj`/`out_proj` weights — applied over the `T` axis with an SDPA causal mask.
- **Shared SigLIP weights.** The video encoder holds `paligemma_with_expert.paligemma.vision_tower` and `multi_modal_projector` **by reference** (wrapped in a list so `nn.Module` doesn't re-register params). No duplicate download, half the memory, and pi05 checkpoints load with `strict=True` and **bit-identical state_dict keys**. `SpaceTimeEncoderLayerWrapper` adopts the underlying `SiglipEncoderLayer`'s submodules (`self_attn`, `layer_norm1`, `layer_norm2`, `mlp`) as its own — a vanilla pi05 checkpoint's keys load into the wrapped layer untouched.
- **Fixed sinusoidal temporal PE** with `e(current) = 0`. The current frame lives at `t = T-1`, verified against [factory.py:136](src/opentau/datasets/factory.py#L136) (delta_timestamps) and `_build_history_batch` (FIFO deque with newest frame last). `obs_history_is_pad[:, -1]` is always `False` by construction.
- **Past-timestep tokens are dropped** after the encoder; only the current frame's 256 tokens are returned, matching a single-frame VLA's image-token count.
- **Single-frame invariance.** With `T=1` the wrapper short-circuits to the unmodified `SiglipEncoderLayer`, and the encoder reduces **byte-exactly** to `PaliGemmaModel.get_image_features` on the same frame (verified at `atol=0`, not just under tolerance). This holds because at T=1 causal temporal attention is not an identity (it returns `out_proj(v_proj(LN1(x)))`) — so `e(0)=0` alone is insufficient; the block itself must also be bypassed, and it is.

**Config changes**
- `vjepa2_*` fields replaced by a single `spacetime_layer_stride: int = 4`. No more `paligemma_model_name` / `num_video_tokens` / `video_encoder_dtype` — the encoder shares with `paligemma_with_expert`, so it inherits the same weights, device, and dtype.
- `policy.n_obs_history` **removed** — it was a duplicate of `n_obs_steps` (both specify T, both must equal `dataset_mixture.n_obs_history`). Use `n_obs_steps` everywhere. `train.py` sync logic is rewritten to validate `policy.n_obs_steps == dataset_mixture.n_obs_history` (treating unset as 1), applied uniformly to every policy — not just pi05_mem. `history_interval` is still synced for policies that carry it.
- `freeze_vision_encoder=True` freezes SigLIP only; `multi_modal_projector` stays trainable (matches `pi05_continuous_state`). All freezing is handled by `PaliGemmaWithExpertModel.set_requires_grad` — one code path, no duplication.
- The `multi_modal_projector` is applied, but the `/ sqrt(hidden_size)` scaling is intentionally omitted to match the repo's [patched `PaliGemmaModel.get_image_features`](src/opentau/utils/transformers_patch.py) (which strips that scaling). `PaliGemmaForConditionalGeneration` is re-exported from `transformers_patch` so tests import through a path that guarantees the patch is active.

**Side find**: V-JEPA2 was silently receiving unnormalized `[0, 1]` pixels instead of ImageNet-normalized inputs. Filed as #170. The new SigLIP encoder sidesteps it by rescaling `[-1, 1]` in its own forward.

**Unrelated bundled fix** (#175, commit `1f7a7f6`): Sync DeepSpeed `gradient_accumulation_steps` from `TrainPipelineConfig` so that config is the single source of truth. Previously the pipeline raised if the Accelerate YAML's `deepspeed_config.gradient_accumulation_steps` diverged from `cfg.gradient_accumulation_steps`; now the pipeline logs a warning and overrides the YAML value on every rank before `accelerator.prepare()`. Not related to the MEM video encoder work, but it landed on this branch and is shipped with this PR.

### CPU tests (Apple Silicon, every commit)

```
uv run pytest tests/policies/test_pi05_mem.py -m "not gpu"
# 46 passed in ~63s
```

Coverage:
- `TestPI05MemConfig` — `spacetime_layer_stride` + `n_obs_steps` validation.
- `TestTemporalSinusoidalPE` — `e(T-1)=0` boundary, sinusoidal rows elsewhere, `T=1` produces all zeros.
- `TestSpaceTimeSiglipVideoEncoder`:
  - `test_forward_shape` / `test_wrong_num_frames_raises` / `test_wrong_ndim_raises`
  - `test_wrapped_layer_count` — every 4th layer gets wrapped (`[3, 7, 11, 15, 19, 23]`)
  - `test_state_dict_keys_match_vanilla_siglip` — locks in the "pi05 checkpoint loads directly" guarantee
  - `test_single_frame_invariance_structural` — byte-exact T=1 invariance without pretrained weights by sharing SigLIP state_dict between a stride-999 (no wrapping) encoder and a stride-4 wrapped encoder
  - `test_temporal_pe_on_base_layer_device` + `test_forward_after_external_dtype_cast` — regression tests for the GPU-only device-mismatch bug fixed in `a47588a` (see below)

### GPU tests — **run on a real CUDA machine (NVIDIA RTX 5090), all 6 passing**

```
uv run pytest tests/policies/test_pi05_mem_gpu.py -m gpu -v -s
```

The initial GPU run surfaced one real bug: the temporal PE buffer was constructed on CPU by default, and when PaliGemma was moved to CUDA **before** the encoder wrapped its layers (the pretrained-load flow in the test helper), the PE never got moved along with the rest. The wrapper is inserted AFTER `.to("cuda")` and is held in the vision_tower via a ModuleList mutation, but the encoder holds vision_tower in a list to avoid state_dict key duplication — so `encoder.to("cuda")` doesn't traverse into the wrapper.

Why training didn't hit this: accelerate's `prepare()` calls `.to(device)` **after** construction, which walks the full module tree (including the wrapped layers) and moves the PE correctly. Only the direct `from_pretrained(...).to("cuda")` → construct-encoder flow triggers the mismatch.

Fix in `a47588a`:
- `SpaceTimeEncoderLayerWrapper.__init__` now builds the PE on the base layer's current `device` and `dtype` (read from `base_layer.self_attn.q_proj.weight`).
- `forward` also casts the PE with `device=x.device` as a cheap no-op safety net in case something moves the tensors between construction and forward.
- Two new CPU regression tests (above) reproduce the pattern without needing CUDA.

After the fix, all six GPU tests pass:

1. `test_forward_shape_and_dtype_cuda` — B=2, T=8, bf16 → `(2, 256, 2048)`.
2. `test_single_frame_invariance_cuda` — matches the patched `PaliGemmaModel.get_image_features` to `atol=rtol=1e-3` (passed even before the fix because T=1 short-circuits the affected code path).
3. `test_gradient_freeze_cuda` — confirms `vision_tower` is frozen when externally requested, `multi_modal_projector` stays trainable.
4. `test_gradient_unfreeze_cuda` — without freezing, gradients reach SigLIP.
5. `test_causality_smoke_cuda` — perturbing the current frame dominates perturbing the oldest frame by ≥ 2× in L2 norm.
6. `test_policy_end_to_end_cuda` — constructs `PI05MemPolicy(init_strategy="no_init")` with finite `dataset_stats` and runs one forward + backward.

### Validation loader uses the same history stacking

The validation dataloader produces `(B, T, C, H, W)` camera tensors with the same `T = dataset_mixture.n_obs_history` as training. [factory.py:218-224](src/opentau/datasets/factory.py#L218-L224) constructs a single `LeRobotDataset` with `delta_timestamps` already encoding history, then `torch.utils.data.random_split` partitions indices; both train and val subsets are `Subset` views over the same underlying dataset and share the same `delta_timestamps`. `obs_history_is_pad` behaves identically — leading slots are clamped when the sampled index falls near the start of an episode. The encoder sees the same shape in both phases; there is no validation-path T-mismatch.

### Earlier MPS verification (Apple Silicon)

Before the GPU machine was available, an ephemeral MPS-adapted copy of the GPU suite ran locally on Apple Silicon to de-risk. That run passed at `atol=0` on T=1 invariance at fp32. Three other bugs surfaced then were fixed in `aa07bc3` (T=1 short-circuit, state_dict param duplication, causality threshold, E2E test setup).

CPU-side:

```bash
uv run pytest tests/policies/test_pi05_mem.py -m "not gpu"
```

GPU-side (requires CUDA + access to `google/paligemma-3b-pt-224`):

```bash
uv run pytest tests/policies/test_pi05_mem_gpu.py -m gpu -v -s
```

Sanity-check the encoder shape & wrapping structure without an HF download:

```bash
uv run python - <<'PY'
import torch
from transformers import PaliGemmaConfig, SiglipVisionConfig, SiglipVisionModel
from transformers.models.paligemma.modeling_paligemma import PaliGemmaMultiModalProjector
from opentau.policies.pi05_mem.video_encoder import (
    SpaceTimeSiglipVideoEncoder, SpaceTimeEncoderLayerWrapper,
)
vision_cfg = {"hidden_size":1152,"intermediate_size":4304,"model_type":"siglip_vision_model",
  "num_attention_heads":16,"num_hidden_layers":27,"num_image_tokens":256,
  "patch_size":14,"image_size":224,"projection_dim":2048,
  "projector_hidden_act":"gelu_fast","vision_use_head":False}
vt = SiglipVisionModel(SiglipVisionConfig(**vision_cfg))
proj = PaliGemmaMultiModalProjector(PaliGemmaConfig(
  vision_config=vision_cfg, text_config={"hidden_size":2048,"model_type":"gemma","vocab_size":257152}))
enc = SpaceTimeSiglipVideoEncoder(vision_tower=vt, multi_modal_projector=proj,
                                   num_frames=2, spacetime_layer_stride=4).eval()
with torch.no_grad():
    out = enc(torch.rand(1, 2, 3, 224, 224))
print("out:", tuple(out.shape), out.dtype)
print("wrapped indices:", [i for i, l in enumerate(vt.vision_model.encoder.layers)
                           if isinstance(l, SpaceTimeEncoderLayerWrapper)])
PY
```

## Checklist

> GPU tests were run on a real NVIDIA RTX 5090 CUDA machine. All 6 tests in `test_pi05_mem_gpu.py` pass after the device-mismatch fix in `a47588a`. Initial run surfaced one real bug (fixed, with two CPU regression tests added to lock it in).


shuheng-liu:
LIBERO finetuning off the tPi0.5-libero checkpoint verified this works with 96.875% success rate
image

shuheng-liu:
Training from TensorAuto/pi05_base on libero using  [866b66c](https://github.com/TensorAuto/OpenTau/pull/171/commits/866b66c152b81e7e5e24470b74ed6a562d9800ae):

image

akshay18iitg (approved):

PR #174, opened 2026-04-22
fix: indicator embedding scaling, padding, and backward compat

shuheng-liu:
## What this does

Fixes several issues found during review of #161:

1. **Missing embedding scaling on indicator tokens (bug).** The indicator block (`<eos>Actions:` / `<eos>Response:`) is embedded via `embed_language_tokens` but was not scaled by `sqrt(hidden_dim)`, unlike the language and response embeddings. This makes the indicator ~1/45x the magnitude of surrounding tokens in the embedding sequence, effectively invisible to attention. Adds the missing `* math.sqrt(indicator_emb_dim)` scaling.

2. **Indicator padding side (design-goal fix).** The indicator was tokenized with `padding_side="right"`, which places `<pad>` tokens at the end of the block. Since `<eos>Actions:` likely tokenizes to fewer than 4 tokens, the *last* position of the indicator block (the one whose hidden state produces the first discrete-action logit) would be a pad token — defeating the design goal of having `":"` absorb that prediction pressure. Switches to `padding_side="left"` so pads go at the front and `":"` is always the rightmost token.

3. **`state_proj` key filter too narrow.** `_fix_pytorch_state_dict_keys` used `key.startswith("state_proj.")`, which only matches top-level keys from openpi-format checkpoints. In-tree checkpoints use `model.state_proj.*`. Widens to `"state_proj." in key`. Also downgrades the log level from `warning` to `info` since this is an expected skip, not an error.

4. **Draccus registry break for `pi05_continuous_state`.** The factory-level deprecation alias in #161 handles programmatic `get_policy_class("pi05_continuous_state")` calls, but `PreTrainedConfig` uses `draccus.ChoiceRegistry` for config deserialization, and the old `@PreTrainedConfig.register_subclass("pi05_continuous_state")` decorator was removed with the deleted file. Adds a thin `PI05ContinuousStateConfig(PI05Config)` shim that re-registers the old name, defaults `state_type="continuous"`, and emits a `DeprecationWarning`. Re-exports it from `policies/__init__.py` so `from opentau.policies import PI05ContinuousStateConfig` still works.

5. **Stale comments.** Removes two comments that referenced the old "state before prompt" layout which no longer applies.

## How it was tested

- Pre-commit hooks pass (ruff, ruff-format, bandit, typos, license headers, etc.)
- Visual inspection of the embedding scaling against the language and response paths

```bash
# Verify the indicator scaling matches language/response:
grep -A3 "indicator_emb = self.paligemma" src/opentau/policies/pi05/modeling_pi05.py

# Verify draccus shim registers correctly:
python -c "from opentau.policies import PI05ContinuousStateConfig; print(PI05ContinuousStateConfig())"

# Verify state_proj filter catches nested keys:
grep "state_proj" src/opentau/policies/pi05/modeling_pi05.py
```

## Checklist

https://claude.ai/code/session_01EKmrD6GJtFu4MV6WJqRXqi

### DIFF
--- a/src/opentau/policies/__init__.py
+++ b/src/opentau/policies/__init__.py
@@ -21,5 +21,6 @@ such as PI0, PI05, and Value policy.
 
 from .pi0.configuration_pi0 import PI0Config as PI0Config
 from .pi05.configuration_pi05 import PI05Config as PI05Config
+from .pi05.configuration_pi05 import PI05ContinuousStateConfig as PI05ContinuousStateConfig
 from .pi05_mem.configuration_pi05 import PI05MemConfig as PI05MemConfig
 from .value.configuration_value import ValueConfig as ValueConfig
--- a/src/opentau/policies/pi05/modeling_pi05.py
+++ b/src/opentau/policies/pi05/modeling_pi05.py
@@ -470,8 +470,8 @@ class PI05Policy(PreTrainedPolicy):
                 new_key = key.replace("action_time_mlp_in.", "time_mlp_in.")
             elif key.startswith("action_time_mlp_out."):
                 new_key = key.replace("action_time_mlp_out.", "time_mlp_out.")
-            if key.startswith("state_proj.") and model_config.state_type == "discrete":
-                logging.warning(f"Skipping state_proj key in discrete state mode: {key}")
+            if "state_proj." in key and model_config.state_type == "discrete":
+                logging.info(f"Skipping state_proj key in discrete state mode: {key}")
                 continue
 
             # Handle vision tower embedding layer potential differences
@@ -837,7 +837,6 @@ class PI05Policy(PreTrainedPolicy):
         tasks = batch["prompt"]
 
         if self.config.state_type == "continuous":
-            # the below prompt <eos>Actions is added assuming state will arrive before prompt in continuous mode
             prompt = [f"Task: {task}" for task in tasks]
         else:
             # add state to the prompt
@@ -873,7 +872,7 @@ class PI05Policy(PreTrainedPolicy):
             tokenized_indicator = self.language_tokenizer.__call__(
                 indicator,
                 padding="max_length",
-                padding_side="right",
+                padding_side="left",
                 max_length=4,
                 return_tensors="pt",
                 truncation=True,
@@ -1124,7 +1123,6 @@ class PI05FlowMatching(nn.Module):
         num_lang_embs = lang_emb.shape[1]
         att_masks += [0] * num_lang_embs
 
-        # adds continuous state to the embedding if it is provided before prompt
         if self.config.state_type == "continuous" and state is not None:
             state_emb = self.state_proj(state.to(dtype=_preferred_dtype()))
             state_emb = rearrange(state_emb, "b d -> b 1 d")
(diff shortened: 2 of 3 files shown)


src/opentau/policies/__init__.py
```diff
@@ -21,5 +21,6 @@
 
 from .pi0.configuration_pi0 import PI0Config as PI0Config
 from .pi05.configuration_pi05 import PI05Config as PI05Config
+from .pi05.configuration_pi05 import PI05ContinuousStateConfig as PI05ContinuousStateConfig
 from .pi05_mem.configuration_pi05 import PI05MemConfig as PI05MemConfig
 from .value.configuration_value import ValueConfig as ValueConfig
```
shuheng-liu, line 26: @claude What's the deal with these "import xxx as xxx"?

src/opentau/policies/pi05/modeling_pi05.py
```diff
@@ -470,8 +470,8 @@ def _fix_pytorch_state_dict_keys(
                 new_key = key.replace("action_time_mlp_in.", "time_mlp_in.")
             elif key.startswith("action_time_mlp_out."):
                 new_key = key.replace("action_time_mlp_out.", "time_mlp_out.")
-            if key.startswith("state_proj.") and model_config.state_type == "discrete":
-                logging.warning(f"Skipping state_proj key in discrete state mode: {key}")
+            if "state_proj." in key and model_config.state_type == "discrete":
+                logging.info(f"Skipping state_proj key in discrete state mode: {key}")
```
shuheng-liu: @claude I'd keep the `.startswith()` pattern. It was intentional.

src/opentau/policies/pi05/modeling_pi05.py
```diff
@@ -873,7 +872,7 @@ def prepare_language(self, batch: dict[str, Tensor]) -> tuple[Tensor, Tensor]:
             tokenized_indicator = self.language_tokenizer.__call__(
                 indicator,
                 padding="max_length",
-                padding_side="right",
+                padding_side="left",
```
shuheng-liu, line 875: @claude Why are we changing the padding side?

shuheng-liu (commented):
@claude Fix the issues in the comments

src/opentau/policies/__init__.py
shuheng-liu: The `import X as X` pattern is a Python convention that tells type checkers (mypy, pyright) "this is an intentional public re-export". Without the `as X`, some type checkers treat the import as private and warn downstream consumers who try to import it from this module.

I just followed the existing pattern — every other import in this file already uses `as` for the same reason (PI0Config, PI05Config, PI05MemConfig, ValueConfig).

---

src/opentau/policies/pi05/modeling_pi05.py
shuheng-liu: Reverted in 2277f2c.

---

src/opentau/policies/pi05/modeling_pi05.py
shuheng-liu: The indicator is tokenized with `max_length=4` and `padding_side="right"`. If `"<eos>Actions:"` tokenizes to fewer than 4 tokens (likely 3: `[<eos>, Actions, :]`), right-padding puts a `<pad>` at the end of the block:

```
[<eos>, Actions, :, <pad>]
                    ^^^^^^ this is position -N-1 (last prefix token before discrete actions)
```

That last position is the one whose hidden state produces the logit for the first discrete action token (via `prefix_out[:, discrete_token_start - 1]`). With right-padding, that position is a `<pad>` — so the design goal of having `:` absorb the first-action prediction pressure is quietly defeated.

With `padding_side="left"`, pads go at the front:

```
[<pad>, <eos>, Actions, :]
                        ^^ now ":" is at position -N-1
```

So the `:` token is always the rightmost position of the indicator block regardless of how many tokens the tokenizer produces.

---

akshay18iitg: Should'nt matter much as these pad tokens are not attended by any discrete action tokens.

We can still keep the paddings to left.

akshay18iitg (approved):

PR #176, opened 2026-04-22
perf(pi05): 2.9x LIBERO training throughput via DDP + fused AdamW

shuheng-liu:
## What this does

Net **2.90× training throughput** for the pi05 / LIBERO recipe on fixed 8×A100 hardware: **28.7 → 83.1 global samples/s**. Per-sample GPU-hours drop from **~77.4 → ~26.7** per 1M samples; wall-clock per sample drops **~2.89×**. Same model, same loss, no hardware change.

Label: ⚡️ Performance

This PR is the consolidated outcome of the investigation tracked in #177. Three classes of change land, each justified by a controlled benchmark on the same 8×A100 box:

1. **Switch the libero distributed default from DeepSpeed ZeRO-2 to plain DDP** (the single biggest contributor — 2.64× by itself). DeepSpeed's per-parameter Python hook overhead was the wall on this workload; single-GPU `bwd` is 641 ms (healthy ~1.7× forward, compute-bound), vs. 3442 ms under 8-rank ZeRO-2 at bs=14 — the ~2.8 s delta is host-side hook execution, not GPU compute and not NCCL. See [#177](https://github.com/TensorAuto/OpenTau/issues/177) for the full chain of evidence.
2. **Drop ~263 M dead params** (`gemma_expert.lm_head`) from the pi05 / pi0 model graph. The action expert is used as an encoder; its output flows through `action_out_proj` / `da_head`, never through a token head. Removing the unused `lm_head` lets DDP run with `find_unused_parameters=False` (no per-step graph walk) and frees enough memory to bump per-rank batch from 11 → 12.
3. **Default to fused AdamW** (`torch.optim.AdamW(fused=True)`). One-line change in the optimizer config; cuts `optimizer.step()` from **99.6 ms → 39.7 ms** (60% reduction) on this model size.

Plus diagnostic infrastructure (`profile_step.py`, `profile_dataloader.py`, `find_unused_params.py`) and a config / docs cleanup.

### Throughput progression (fixed 8×A100, bf16, same model, same LIBERO training config)

Each row is `profile_step.py` with warmup=20 + measured=200 steps, batch size maxed for that configuration (verified via `nvidia-smi`). Memory and wall-clock are per-rank / per-step averages:

| State | bs/rank | total ms | mem (GiB) | **samples/s** | vs baseline | Attribution |
|---|---:|---:|---:|---:|---:|---|
| **Pre-PR: ZeRO-2 8× @ bs=14** | 14 | 3909 | 79.7 | **28.7** | **1.00×** | 8-rank DeepSpeed ZeRO-2, same as pre-PR libero config would give on 8 GPUs |
| ZeRO-1 8× @ bs=14 *(control)* | 14 | 3799 | 79.7 | 29.5 | 1.03× | Same Python-hook overhead as ZeRO-2 |
| → DDP 8× @ bs=11 *(fup=true)* | 11 | 1160 | 78.0 | **75.8** | **2.64×** | Backend swap, biggest contributor |
| → + drop `gemma_expert.lm_head`, `fup=false`, bs=12 | 12 | 1216 | 79.2 | **78.9** | **2.75×** | Memory headroom + per-step graph walk removed |
| → + fused AdamW | 12 | 1155 | 79.2 | **83.1** | **2.90×** | `optim_step` 99.6 → 39.7 ms |

For context / decoupling the hardware axis, single-GPU at bs=13 gets **10.4 samples/s** (scaling efficiency from 1 → 8 GPUs: 83.1 / 10.4 = 7.99× on 8× hardware — near-ideal DDP scaling). A user moving from the old 4-GPU libero config to 8 GPUs gets an additional ~2× hardware scaling on top of the 2.90× software win, but that's a separate decision from this PR.

### Economic picture (per 1M optimizer steps, 8×A100)

| Config | step time | wall-clock | GPU-hours | per-sample cost |
|---|---:|---:|---:|---:|
| Pre-PR: ZeRO-2 8× @ bs=14 | 3909 ms | **1086 h (45 d)** | **8688** | 77.6 GPU-h / 1M samples |
| Post-PR: DDP 8× + fused AdamW @ bs=12 | 1155 ms | **321 h (13 d)** | **2568** | **26.8 GPU-h / 1M samples** |

**~2.89× cheaper per sample**, ~3.38× faster wall-clock per optimizer step at a slightly smaller global batch (96 vs 112).

### What changed (file-level)

**Production code** (4 files):

- `src/opentau/policies/pi05/paligemma_with_expert.py` — `self.gemma_expert.lm_head = None` mirroring the existing `embed_tokens = None` pattern at L234.
- `src/opentau/policies/pi0/paligemma_with_expert.py` — same fix; pi0 has identical structure. (`pi05_mem` and `pi05_continuous_state` import the same `PaliGemmaWithExpertModel` class so they inherit the fix transitively, no per-policy code change needed; verified by grep.)
- `src/opentau/optim/optimizers.py` — `fused: bool = True` default on `AdamConfig` and `AdamWConfig`. Guard in `build()` falls back to `foreach` on CPU-only hosts so tests / single-process debug don't break. State-dict layout unchanged; existing checkpoints load.
- `src/opentau/scripts/train.py` — `find_unused_parameters` gated behind a `FIND_UNUSED_PARAMS` env var (default `True` for safety with unaudited policies). `find_unused_params.py` is the audit tool.

**Diagnostic scripts** (3 new files under `src/opentau/scripts/`):

- `profile_step.py` — splits per-step wall-clock into `dataload_wait / forward / bwd / unscale_clip / optim_step / zero_grad_sched / sync_gather` phases. Env-var overrides (`FUSED_ADAMW`, `FIND_UNUSED_PARAMS`, `PROFILE_STEPS`, `PROFILE_NO_OPTIM`) for benchmark A/B without touching production configs.
- `profile_dataloader.py` — pure dataloader throughput ceiling, no model, no collective. Used to rule out H1 (dataloader starvation) at the start of the investigation.
- `find_unused_params.py` — single-GPU: prints every parameter where `param.grad is None` after one real `loss.backward()`. Run this before flipping `FIND_UNUSED_PARAMS=false` on any new policy.

**User documentation** (new tutorial + cross-ref):

- `docs/source/tutorials/benchmarking.rst` — user-facing guide with per-script commands, annotated example output, how-to-read-the-output guidance, per-script env-var tables, and an end-to-end "diagnose a slow run" section.
- `docs/source/tutorials.rst` — wired the new tutorial into the toctree.
- `docs/source/tutorials/training.rst` — added a `.. tip::` pointing at the benchmarking guide for users investigating slow training.
- `docs/source/installation.rst`, `docs/source/tutorials/training.rst` — recommend DDP as the documented default; keep DeepSpeed as a memory-constrained alternative (still used by `regression_test.yml`).

**Config consolidation**:

- `configs/examples/accelerate_ddp_config.yaml` — replaced (was a 2-GPU `mixed_precision: 'no'` placeholder) with the proven 8-GPU + bf16 setup. Now the documented default.
- `configs/libero/reproduce_pi05_libero_accelerate_config.yaml` — switched from DeepSpeed ZeRO-2 4-GPU to DDP 8-GPU bf16. Existing `opentau-train --accelerate-config=configs/libero/...` invocations get the new behavior automatically.
- `configs/examples/accelerate_deepspeed_config.yaml` — **untouched.** Still consumed by `.github/workflows/regression_test.yml` to keep CI coverage of the DeepSpeed code path, and a valid alternative for memory-constrained scenarios (very large models, long sequences). DeepSpeed isn't broken — it's just not the throughput-optimal default for pi05.
- 9 experimental configs in `configs/libero/` deleted (the `_ddp`, `_ddp_8gpu`, `_zero1`, `_zero1_8gpu`, `_tuned`, `_tuned_8gpu`, `_single` variants and the two raw deepspeed JSONs from the investigation).

### What this does *not* change

- **No change to forward/backward math.** Same model, same losses, same gradients (modulo bf16 reassociation noise from the fused Adam kernel — bit-identical to within ~1e-7 relative).
- **No change to the public training API.** `opentau-train --config_path=...` works exactly as before; users on the existing libero accelerate config get the speedup automatically.
- **DeepSpeed code path stays supported.** CI still tests it; users with memory-constrained scenarios can opt back via `--accelerate-config=configs/examples/accelerate_deepspeed_config.yaml`.
- **No SDPA, no gradient checkpointing.** Both were tried, both were break-even or net loss for pi05 at this scale, both were reverted in commits `7a5a923` and `97be5e5`. Full benchmark data and reasoning in the [#177 follow-up comment](https://github.com/TensorAuto/OpenTau/issues/177#issuecomment-4301905944).
- **Sibling policies covered transitively.** `pi05_mem` and `pi05_continuous_state` import `PaliGemmaWithExpertModel` from `pi05/`, so the `lm_head` fix takes effect for them at construction time without per-policy code changes. `value` uses a different architecture and doesn't have the same issue.

### Automated tests (CPU CI, run on every PR)

New unit tests under `tests/scripts/` covering the pure-function helpers in each diagnostic script. All run in milliseconds under `pytest -m "not gpu"` in `.github/workflows/cpu_test.yml`:

- `tests/scripts/test_profile_step.py` — `_fmt_ms` (empty list → `"n/a"`, single value, multi-value mean / median, p95 index-based formula, sort-before-percentile, large-value ms rendering).
- `tests/scripts/test_profile_dataloader.py` — `_to_device` on CPU (tensor passthrough with object identity, dict / nested dict / list / tuple recursion, non-recursion into strings, scalar passthrough); `_batch_size_of` (tensor batch dim, 0-D tensor, dict, nested dict, list, empty dict / list, dict with only non-tensors, scalars / None / strings).
- `tests/scripts/test_find_unused_params.py` — `_module_root` (default depth=2, depth=3, depth=1, path shorter/equal to depth, depth=0 edge case).

### Integration / end-to-end (what's deliberately *not* unit tested)

The actual loop bodies of each script — `profile_step.profile()`, `profile_dataloader.profile()`, `find_unused_params.find()` — are not unit-tested because doing so meaningfully would require mocking the entire `accelerate` + `torch` + `datasets` + policy-construction stack, which is neither a useful unit test nor a stable one. The core functionality they exercise is already covered end-to-end:

- **The `accelerate launch` + `Accelerator.prepare` + dataloader + policy + optimizer path** is exercised by every run of `opentau-train`, and specifically by `.github/workflows/regression_test.yml:111` which runs a short real training every PR.
- **The fused-AdamW + DDP + `gemma_expert.lm_head`-removed production code paths** are exercised the moment anyone re-runs training with the updated configs in this PR. The benchmarks in the progression table above were all taken on the same combined path.
- **Loss / correctness preservation** is guaranteed by construction for the production changes:
  - `fused AdamW` — same Adam update equations, same state-dict shape, bit-identical up to fp32 reassociation order. Used by HF Trainer / Megatron-LM / NeMo in production for years.
  - `gemma_expert.lm_head = None` — removing a parameter that received no gradient. Cannot affect training dynamics by construction; verified empty `UNUSED` set from `find_unused_params.py` after the change.

### Live benchmarks

Every row in the progression and economics tables above was produced by `profile_step.py` warmup=20 + measured=200 on `gpu-387` (8×A100 80 GB, NV12 NVSwitch), batch size verified at the memory ceiling via `nvidia-smi`. Pre-PR baseline is the same-hardware 8-rank ZeRO-2 control, not the original 4-GPU config — hardware scaling is decoupled from this PR's claims.

### Local checks before push

Pre-commit hooks (ruff, ruff-format, pyupgrade, typos, license-header) clean on all touched files. `check-checklist` CI passes.

Diagnostic — confirm no params are unused under the new pi05:
```bash
python src/opentau/scripts/find_unused_params.py \
    --config_path=configs/libero/reproduce_pi05_libero.json
# Expected: UNUSED (0 tensors, 0 params)
```

Throughput sanity check — should see `optim_step ≈ 40 ms`, `total ≈ 1155 ms`, `samples/s ≈ 83`:
```bash
FIND_UNUSED_PARAMS=false accelerate launch \
    --config_file configs/libero/reproduce_pi05_libero_accelerate_config.yaml \
    --config_path=configs/libero/reproduce_pi05_libero.json \
    --batch_size=12
```

Fair A/B — reproduce the 2.90× claim by running the same script with the DeepSpeed config:
```bash
# Baseline (ZeRO-2 on the same 8 GPUs): ~28-29 samples/s expected
accelerate launch \
    --config_path=configs/libero/reproduce_pi05_libero.json \
    --batch_size=14
```

Unit tests (no GPU needed):
```bash
pytest -m "not gpu" -v tests/scripts/test_profile_step.py \
    tests/scripts/test_profile_dataloader.py \
    tests/scripts/test_find_unused_params.py
```

Real training (the recipe is now ~2.9× faster per sample vs. main):
```bash
opentau-train \
    --accelerate-config=configs/libero/reproduce_pi05_libero_accelerate_config.yaml \
    --config_path=configs/libero/reproduce_pi05_libero.json
```

## Checklist

(Note: this PR does touch `src/opentau/policies/pi05/paligemma_with_expert.py` and `src/opentau/policies/pi0/paligemma_with_expert.py` — both changes drop a parameter that DDP confirmed receives no gradient under any training config, so there is no behavioral change to gate behind a GPU regression run. Happy to flip the box and add a regression test if reviewers prefer.)

Tracking: #177
Branch: `claude/debug-gpu-utilization-WiDcZ`

https://claude.ai/code/session_0129LtYYua8s3sJ4gfcjn1VB


shuheng-liu (commented):
## Review summary

Consolidates an 8×A100 throughput investigation (tracked in #177) into three independently-justified changes plus diagnostics. The 2.90× claim is backed by a clear, controlled progression table with same-hardware baselines; A/B methodology is sound. No math/loss change — bit-identical up to bf16 fused-Adam reassociation noise.

### Code correctness
- **`gemma_expert.lm_head = None` is safe.** Verified by grep across `src/opentau/policies/**`: the only `lm_head` references in modeling code are to `paligemma.lm_head` (the VLM side). `gemma_expert` is consumed via `.model` at `pi05/paligemma_with_expert.py:387` and `pi0/paligemma_with_expert.py:325`, never via a token head. `pi05_mem` / `pi05_continuous_state` reuse the same class and inherit the fix transitively.
- **`fused=True` default is well-guarded.** `build()` falls back when `not torch.cuda.is_available()` so CPU-only tests keep working. `torch>=2.7.1` (from `pyproject.toml`) comfortably supports `fused`. Existing optimizer JSONs don't specify `fused`, so dataclass defaults apply — no config-schema break. State_dict layout unchanged, so existing checkpoints keep loading.
- **`FIND_UNUSED_PARAMS` default stays `true`.** Opt-in env var; existing users get no behavioral change until they deliberately audit and flip it. Good migration posture.

### Minor nits (non-blocking)
- **Hardcoded `"pi05 parameter audit"` in `src/opentau/scripts/find_unused_params.py:154`.** The script is actually policy-agnostic (driven by `cfg.policy`); the header is misleading when run against pi0 or value. Consider `f"# {cfg.policy.type} parameter audit"` or drop the name.
- **`torch.tensor(s).prod()` in `_summarize`** (`find_unused_params.py:158,169`) is a heavy way to compute a shape product — `math.prod(s)` avoids allocating a CPU tensor per parameter.
- **`_to_device` tuple reconstruction** (`profile_dataloader.py:77`, `find_unused_params.py:42`) does `type(obj)(out)`, which would break on a namedtuple batch (namedtuples don't accept a single iterable in their ctor). Unlikely in practice with dict/tensor dataloaders, but worth a note.
- **`profile_step.py` FUSED_ADAMW monkey-patching** (L148–159) wraps `optimizer.step` with a synthetic `_step_count`. Author calls this out explicitly as benchmark-only; scoping is fine but it's fragile (depends on `_LRScheduler` internals). Acceptable in a diagnostic script.
- **`kwargs_handlers` set unconditionally in `profile_step.py`.** Under DeepSpeed the `DistributedDataParallelKwargs` is silently ignored (docstring says so), but constructing it when the backend is DeepSpeed is wasted. Harmless.
- **`AdamConfig.fused` docstring** cites a specific 60% win for "large models (~3B+ trainable params)" — this is representative of the pi05 benchmark, not universal. Consider softening.

### Test coverage
- Unit tests cover all the pure helpers (`_fmt_ms`, `_to_device`, `_batch_size_of`, `_module_root`). Reasonable given the rest is exercised end-to-end by `regression_test.yml`.
- **Gap:** no test asserts that `AdamConfig(fused=True).build([cpu_param])` falls back to `fused=False` on a CPU-only host. That's the branch most likely to regress silently. A two-line test (`fused=True` + `torch.cuda.is_available() is False`) would cement it.
- **Gap:** no test for `FIND_UNUSED_PARAMS` env parsing in `train.py:130`. Easy with monkeypatch.
- Policy `lm_head = None` changes aren't behind a GPU regression run; given `find_unused_params.py` mechanically proves the removal is invariant, I'd accept as-is.

### Performance / risk
- DeepSpeed ZeRO-2 path still exercised via `regression_test.yml`, so `fused=True` will run through it there — any DeepSpeed/fused incompat will surface in CI.
- Config cleanup deletes 9 experimental libero configs — worth a final check that nothing in tests or docs references the deleted variants.

### Security
No security-relevant changes. Env vars read and compared as strings; no shell eval, no network, no deserialization of external input.

### Recommendation
Approve with two small asks:
1. Fix the hardcoded `"pi05"` string in `find_unused_params.py`.
2. Add a one-liner test for `AdamWConfig.build()`'s CPU fallback branch (`fused=True` + no CUDA → `fused=False`).

Everything else is either cosmetic or adequately justified. Well-scoped PR, thorough evidence, and the opt-in design for `FIND_UNUSED_PARAMS` keeps the blast radius small.

_(Automated review; a human should make the final merge call.)_

---

akshay18iitg (changes requested):
Lets do a quick libero run for pi05

shuheng-liu:
@akshay18iitg This PR doesn't change the policy behavior at all. It only removed an unused parameter, which would have errored out if that parameter is ever-used. I have proven in my experiments it runs bug-free.

akshay18iitg (approved):

PR #180, opened 2026-04-24
fix(train): re-seed new ranks deterministically when resuming on more GPUs

shuheng-liu:
## Summary

When resuming DDP training with `--resume` on a **different number of GPUs** than the checkpoint was saved with, Accelerate's `load_state` behavior is asymmetric and silent — correct for scale-down but lossy for scale-up. This PR adds `reseed_new_ranks_on_resume` (called inside `cfg.resume` in `train.py`) that detects the mismatch by counting `random_states_*.pkl` files in the checkpoint, re-seeds new ranks deterministically in the scale-up case, and surfaces the change to operators so they know to adjust `gradient_accumulation_steps`.

## Background: what Accelerate does today

`accelerator.save_state` writes one `random_states_<process_index>.pkl` per rank (plus the consolidated `model.safetensors`, `optimizer.bin`, `scheduler.bin`). On load, `accelerate/checkpointing.py:287-309` only opens **the single file matching the current rank's `process_index`**, inside a `try/except Exception` that falls through to `logger.info("Could not load random states")`. It never enumerates or cross-checks file count vs. `num_processes`.

I verified this behavior against `accelerate==1.12.0` (OpenTau's pin) with a standalone smoke test — see the "Test plan" section.

### Scale-down: 8 GPUs → 4 GPUs

- **Per-rank RNG files 4–7 are silently ignored.** Accelerate only reads the file matching the current rank (0–3), so the extras are never opened. Even a corrupt `random_states_7.pkl` wouldn't be noticed. Surviving ranks (0–3) correctly restore their per-rank RNG from the checkpoint.
- **Model / optimizer / scheduler / training step** all load correctly — they're consolidated single-file artifacts, rank-count-independent.
- **OpenTau's consolidated `rng_state.safetensors`** (via `random_utils.py:load_rng_state`) is loaded by all 4 ranks and gives them identical Python/NumPy/Torch CPU RNG.
- **Global batch size halves.** `train.py:269` computes `cfg.batch_size * accelerator.num_processes`, so going 8→4 cuts effective batch in half and changes the optimizer-step semantics (same LR schedule now covers half the data per step).
- **DataLoader re-shards** from 8-way to 4-way stripe under `accelerator.prepare`; sampler position is not checkpointed today, so data ordering diverges from the original run either way.
- **No warning or error is emitted** by anything in the current code path — operators have no signal that the effective batch size changed.

### Scale-up: 4 GPUs → 8 GPUs

- **Ranks 0–3 restore their per-rank RNG correctly** from existing files.
- **Ranks 4–7 have no matching file.** Accelerate's `try/except` catches the `FileNotFoundError`, logs `"Could not load random states"` at INFO (easy to miss), and those ranks continue with whatever default RNG the process happened to start with — correlated across ranks and non-reproducible.
- **Consolidated `rng_state.safetensors`** gives all 8 ranks the same CPU RNG state; CUDA/dropout/augmentation RNG on the new ranks is undefined.
- **Global batch size doubles**, same caveat as above but in the opposite direction.

## What this PR changes

`reseed_new_ranks_on_resume(checkpoint_dir, accelerator, seed)`:

- Counts `random_states_*.pkl` in the checkpoint to infer the saved world size.
- **Scale-up** (`current > saved`): emits a `logging.warning` naming both rank counts and the global-batch scale factor, and re-seeds each new rank (`process_index >= saved`) via `set_seed(cfg.seed, accelerator)`, which yields `cfg.seed + process_index * 12345` per rank — deterministic and decorrelated. Surviving ranks are not touched.
- **Scale-down** (`current < saved`): emits a `logging.info` naming both rank counts and the global-batch scale factor. No RNG mutation — Accelerate's silent-ignore of extra files is already correct; this PR just makes the change visible in the log.
- **Same world size**: silent no-op.
- **`cfg.seed is None`**: emits a secondary warning instead of re-seeding (no deterministic seed to use).

In both scale-up and scale-down cases, the log message includes the phrase _"global batch size has changed by a factor of Nx; consider scaling cfg.gradient_accumulation_steps to compensate"_ so operators have a clear cue.

## Out of scope

- **Auto-adjusting `gradient_accumulation_steps` / `batch_size`** — left as an operator decision; the warning surfaces it. Halving grad-accum when going 4→8 (or doubling it when going 8→4) preserves global batch and LR-schedule semantics; `batch_size` scaling would additionally change per-forward BN-like behavior, so grad-accum is usually the right knob.
- **Dataloader sampler state** — OpenTau does not checkpoint `DistributedSampler` position today, so data re-sharding on rank-count change is pre-existing behavior in both directions.
- **FSDP / DeepSpeed paths** — optimizer state is sharded per rank and handled separately by those frameworks; this PR only touches the standard DDP resume path.

## Checklist

## Test plan

- [x] Verified Accelerate 1.12.0 behavior via a standalone smoke test driving `load_accelerator_state` directly:
  - Missing `random_states_<rank>.pkl` (scale-up case) → `try/except Exception` in `accelerate/checkpointing.py:287-309` silently logs INFO and returns; confirmed.
  - Extra `random_states_<rank>.pkl` files (scale-down case) are never opened; corrupt extras are ignored; confirmed.
- [x] Smoke-tested the helper logic across 5 cases (scale-up new rank, scale-up surviving rank, scale-down, same size, `seed=None`) — all pass; new rank's RNG matches the deterministic `seed + process_index * 12345` reconstruction exactly.
- [x] Ran `pre-commit run --files src/opentau/scripts/train.py src/opentau/utils/train_utils.py` locally — all hooks pass.
- [ ] End-to-end test on a real multi-GPU node: save on 4 GPUs, resume on 8; inspect the new `WARNING` and verify loss trajectory.
- [ ] End-to-end test the reverse: save on 8, resume on 4; verify the `INFO` log and that global-batch guidance is acted on.

https://claude.ai/code/session_016YyNbCipf9hgp7jcPUxepn

---
_Generated by [Claude Code](https://claude.ai/code/session_016YyNbCipf9hgp7jcPUxepn)_

### DIFF
--- a/src/opentau/scripts/train.py
+++ b/src/opentau/scripts/train.py
@@ -50,6 +50,7 @@ from opentau.utils.train_utils import (
     load_training_state,
     load_training_step,
     prune_old_checkpoints,
+    reseed_new_ranks_on_resume,
     save_checkpoint,
 )
 from opentau.utils.utils import (
@@ -248,6 +249,11 @@ def train(cfg: TrainPipelineConfig):
         # load accelerator state
         # This will load the model, optimizer, and lr_scheduler state
         accelerator.load_state(cfg.checkpoint_path)
+
+        # Per-rank random_states_<rank>.pkl is loaded by accelerator.load_state above,
+        # but missing files (when resuming on more GPUs than were saved) are silently
+        # skipped. Re-seed those new ranks deterministically.
+        reseed_new_ranks_on_resume(cfg.checkpoint_path, accelerator, cfg.seed)
 
         # all processes should load the step & rng states
         step = load_training_state(cfg.checkpoint_path)
--- a/src/opentau/scripts/train.py
+++ b/src/opentau/scripts/train.py
@@ -400,11 +400,6 @@ def train(cfg: TrainPipelineConfig):
         # This will load the model, optimizer, and lr_scheduler state
         accelerator.load_state(cfg.checkpoint_path)
 
-        # Per-rank random_states_<rank>.pkl is loaded by accelerator.load_state above,
-        # but missing files (when resuming on more GPUs than were saved) are silently
-        # skipped. Re-seed those new ranks deterministically.
-        reseed_new_ranks_on_resume(cfg.checkpoint_path, accelerator, cfg.seed)
-
         # When the master-weights wrapper is in use, the live bf16 weights
         # have just been overwritten by ``accelerator.load_state``. Rebuild
         # the fp32 masters from those weights so the inner optimizer's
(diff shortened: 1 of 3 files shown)


shuheng-liu:
@claude fix merge conflicts.

shuheng-liu:
@claude fix per review if they make sense. Push back if they don't.

WilliamYue37 (approved):

Issue #177, opened 2026-04-23
perf: ~2.9× throughput by switching from DeepSpeed to DDP


shuheng-liu:
## TL;DR

The default distributed backend for pi05 LIBERO training (`distributed_type: DEEPSPEED, zero_stage: 2`) leaves **~65% of GPU time on the table at fixed hardware**. Comparing **same-hardware** 8×A100 configurations, switching the training script to plain DDP (`distributed_type: MULTI_GPU`) combined with two small follow-on fixes (drop an unused `gemma_expert.lm_head` parameter, default fused AdamW) gives **2.90× more samples/s**, ~3.4× faster wall-clock per optimizer step, and **~2.9× fewer GPU-hours per sample** — without changing any model code or loss.

> **Distinguishing hardware scaling from software gains.** The pre-investigation libero accelerate config had `num_processes: 4`. A user upgrading from that 4-GPU config to the new 8-GPU DDP config observes `13.8 → 83.1 samples/s` = **~6.0× end-to-end**, but roughly half of that comes from doubling the hardware (DDP scales near-linearly 1→8 GPUs on this model, see the per-rank throughput discussion below). **The software-attributable win is 2.90×.** This issue focuses on the software axis.

Root cause is a known DeepSpeed ZeRO pathology for models with many small parameter tensors (PaliGemma + Gemma expert): per-parameter Python hooks installed on every backward pass dominate step time and cannot be reduced by any communication-overlap knob. Single-rank profiling shows the model is healthy and compute-bound; the overhead only appears once DeepSpeed is in the loop.

## What we ruled out

We worked through the obvious suspects in order. Each one is conclusively *not* the cause:

1. **Dataloader starvation.** `profile_dataloader.py` (no model, no collective) ran at **597 cluster-wide samples/s** while training consumed only ~29 samples/s under ZeRO-2. `dataload_wait` per step is **1.14 ms** in the actual training loop. Despite the suggestive square-wave pattern in the `nvidia-smi` chart, the input pipeline is roughly 20× ahead of the GPUs.
2. **Per-step host/device sync (5× `gather_for_metrics(...).item()` in `update_policy`).** Measured at **3 ms / 0.1%** of step time. Real, but immaterial.
[...]

PR #167, opened 2026-04-21
Feat/pi07 paligemma

akshay18iitg:
## What this does
Adds the complete policy implementation for the π07 PaliGemma hierarchical architecture, consisting of a high-level planner and a low-level planner.

Low-level planner (pi07_paligemma/low_level_planner/)
V-JEPA2 video encoder replaces SigLIP — compresses each camera's temporal video into 256 tokens via a Perceiver cross-attention reducer.
Temporal state sequences (B, T, D) are projected per-timestep into separate VLM tokens (supports observation history via n_obs_history).
Subtask response conditioning — accepts an optional subtask string from the high-level planner, embedded as bidirectional tokens in the prefix.
Subgoal image conditioning — derives subgoal keys from image_features (camera{k} → subgoal{k}) following the LeRobot dataset convention; uses subgoal_is_pad to mask out dropped/absent subgoals.
Structured metadata conditioning — assembles metadata from per-sample speed, quality, mistake fields (each with _is_pad flags) into a tokenized string.
Flow matching with Knowledge Insulation — VLM KV-cache is detached before being passed to the Gemma action expert for continuous action generation.
Discrete (FAST) action prediction via the VLM backbone in parallel with continuous actions.
Prefix token layout: [video | language | response | state | subgoal images | metadata | discrete actions], with bidirectional attention for the core prefix, new bidirectional blocks for subgoals and metadata, and causal attention for discrete actions.
High-level planner (pi07_paligemma/high_level_planner/)
Uses SigLIP (via PaliGemma) for image encoding.
Autoregressive generation of updated memory tokens and subtask (response) tokens.
Structured metadata conditioning — same speed/quality/mistake assembly as low-level.
Prefix token layout: [images | language | metadata | memory | response], with bidirectional attention for images+language, a new bidirectional block for metadata, and causal attention for memory and response.
metadata_max_length added to PI07HighLevelPlannerConfig.
Configuration & types
Added metadata_max_length to both PI07lowlevelPlannerConfig and PI07HighLevelPlannerConfig.
Removed FeatureType.SUBGOAL — subgoal features are now derived from image_features at runtime.
Fixed subgoal_image_features property logic (was an impossible VISUAL and SUBGOAL check).
Docstrings
Google-style docstrings added/updated across all module, class, and method levels for both planners.
Architecture ASCII diagrams updated to reflect metadata and subgoal inputs.
## How it was tested
[...]

PR #182, opened 2026-04-24
fix(pi05): DDP fp32 Adam state + SDPA / grad-ckpt opt-ins (#181, #177)

### LINKED ISSUE #181: DDP vs Deepspeed
### System Info

```Shell
Main branch
```

### Information

- [x] One of the scripts in the src/opentau/scripts/ folder of OpenTau
- [ ] My own task or dataset (give details below)

### Reproduction

Run original pi05 code on libero.

### Expected behavior

Ran 2 exp on libero with same pi05 codebase.

1. Uses deepspeed with 4 gpus - success rate 94%
2. Uses ddp with 4 gpus - success rate 0%

Follow these 2 runs:

ddp - https://wandb.ai/wyautox-autox/pi05/runs/slqmot3k?nw=nwuserashah18
deepspeed - https://wandb.ai/wyautox-autox/pi05/runs/8e7yx2hr?nw=nwuserashah18

shuheng-liu:
## What this does

Combines the **correctness fix for the DDP bf16 Adam state bug** (issue #181, primary) with **two opt-in throughput levers** (SDPA attention backend + per-layer gradient checkpointing) and the **alignment of `profile_step.py`** with `train.py`'s master-weights wrap path so the post-fix benchmark numbers below are reproducible from main after merge.

Originally landed as correctness-only. Expanded to include the SDPA + grad-ckpt opt-ins after a re-bench against the corrected baseline showed the [#177 Phase 3.3 / 3.4 negative results](https://github.com/TensorAuto/OpenTau/issues/177#issuecomment-4301905944) had been measured under the bf16 Adam state bug — both verdicts reverse on the corrected fp32-master baseline. See the [post-fix retrospective on #177](https://github.com/TensorAuto/OpenTau/issues/177#issuecomment-4318533005) and [the loss-equivalence comment on this PR](https://github.com/TensorAuto/OpenTau/pull/182#issuecomment-4318601215) for the full data.

### Bug 0 — bf16 Adam state under DDP (issue #181, primary)

`src/opentau/scripts/train.py` cast the policy to bf16 and then constructed `torch.optim.AdamW` directly over those bf16 parameters. Under DeepSpeed ZeRO-2 this was harmless — ZeRO replaces the optimizer with `BF16_Optimizer`, which keeps fp32 master copies and runs Adam in fp32. Under pure DDP / single-process / FSDP, no swap happens, so `torch.zeros_like(p)` allocated Adam state (`exp_avg`, `exp_avg_sq`) in bf16 and the update ran in bf16 with `eps=1e-8`. Over many steps the state accumulates rounding error and training silently diverges — on pi05 / libero, DDP 4-GPU hit 0% success while DeepSpeed ZeRO-2 4-GPU hit 94%.

`MasterWeightOptimizer` (`src/opentau/optim/master_weights.py`) mirrors DeepSpeed's `BF16_Optimizer` design under every non-DeepSpeed backend:

- **Model stays in bf16** (forward / backward / activation memory / bf16 compute) regardless of backend.
- **Optimizer holds fp32 master copies** of the bf16 params; the underlying `torch.optim.AdamW` operates over those masters, so `exp_avg` / `exp_avg_sq` are fp32 by construction.
- Each `step()` upcasts bf16 grads into fp32 master grads, runs the inner step in fp32, and downcasts the updated fp32 masters back into the live bf16 params.
- `clip_grad_norm_(...)` performs the upcast first and clips on fp32 master grads, returning the fp32 norm; the subsequent `step()` skips a redundant upcast.

| Backend | Model (fwd/bwd) | Optimizer state | Clip point |
|---|---|---|---|
| DeepSpeed ZeRO-2 | bf16 | fp32 (DS BF16_Optimizer master) | fp32 (DS post-upcast) |
| DDP / single / FSDP (pre-fix) | bf16 | **bf16** | bf16 |
| DDP / single / FSDP (post-fix) | bf16 | fp32 (`MasterWeightOptimizer`) | fp32 (post-upcast, matches DS) |

### Bug 1 — LR scheduler frozen at initial value (8be2cd1)

`make_optimizer_and_scheduler` binds the scheduler to the original `torch.optim.AdamW`. `MasterWeightOptimizer.from_existing` discards that AdamW and creates a fresh inner with new `param_groups` dicts. Without rebinding `scheduler.optimizer = wrapped`, `lr_scheduler.step()` mutates the orphan's `param_groups[i]['lr']` and the inner AdamW's lr stays at the construction value forever — wandb shows a flat LR line that looks intentional.

**Fix**: rebind `lr_scheduler.optimizer = wrapped` immediately after `from_existing` (in both `train.py` and `profile_step.py`). Pinned by `test_lr_scheduler_reaches_inner_after_from_existing_with_rebind` (positive case) and `test_lr_scheduler_does_not_reach_inner_without_rebind` (negative control documenting the failure mode if a future change drops the rebind line).

### Bug 2 — accelerate's `isinstance` check skipped the wrapper (8be2cd1)

`Accelerator._prepare_one` (accelerate `accelerator.py:1403`) is a strict `isinstance(obj, torch.optim.Optimizer)` check. The wrapper deliberately wasn't a subclass, so accelerate returned it **unchanged** from `prepare()` — never wrapped in `AcceleratedOptimizer`. Consequences: gradient-accumulation gating bypassed (`grad_accum > 1` would have silently collapsed to effective 1 — invisible on libero where `accum=1`, but a hidden landmine); wrapper missing from `Accelerator._optimizers`, so `prepare_scheduler`'s identity match failed and `AcceleratedScheduler` got an empty optimizer list.

**Fix**: subclass `torch.optim.Optimizer` without calling `super().__init__()`. `param_groups` / `state` / `defaults` already proxy to `self.inner` via existing `@property` accessors. `AcceleratedOptimizer` itself uses the same skip-super pattern. Pinned by `test_wrapper_is_torch_optimizer_subclass`.

### Bug 3 — fp32 masters left on CPU after `accelerator.prepare` (dd1154e)

`MasterWeightOptimizer.from_existing` clones fp32 masters from the live bf16 params at wrap-time, but the wrap fires **before** `accelerator.prepare(...)` (which is when the policy is moved CPU → GPU). The masters were therefore allocated on CPU and stayed there. Adam ran on CPU master tensors with a CPU↔GPU memcpy on every grad upcast and param downcast — functional but very slow under DDP, and the fp32 master state never appeared in nvidia-smi accounting (which is what fooled an earlier round of benchmarks into reporting 33 GiB peaks at bs=4 instead of the correct ~73 GiB).

**Fix**: `rebuild_masters_from_live` re-points `master.data` to live's current device (re-assignment, not in-place `copy_`); migrates Adam state to the new device if populated. Called from `train.py` and `profile_step.py` once after `accelerator.prepare`. Three GPU-marked tests cover the fresh-run / resume-with-state / param-identity cases.

### SDPA attention backend

`PI05Config.attention_implementation: str = "eager"` was a dead flag (the validator accepted `"eager"` or `"fa2"` but `get_attention_interface()` always returned eager). This PR wires it to dispatch to `torch.nn.functional.scaled_dot_product_attention`, which on A100 + bf16 routes to FlashAttention-2 (no S×S materialization in forward, recompute in backward). Default stays `"eager"`; set `"sdpa"` to opt in. **Loss-equivalence A/B passes** at bs=6 / 1020 steps / same seed: median |relative delta| = 0.82%, end-of-run loss eager 1.0356 vs sdpa 1.0099 — equivalent within bf16 reassociation noise.

### Gradient checkpointing

`PI05Config.gradient_checkpointing: bool = False` (new). When True, each transformer-layer body (`_run_layer`, extracted from the inlined per-layer loop) is wrapped in `torch.utils.checkpoint.checkpoint(..., use_reentrant=False)` during training. `train.py` adds a strict startup guard refusing the flag under unsupported backends (ZeRO-3, FSDP) — silent incorrectness is worse than a failed job start. Per-layer mutation of `past_key_values` is idempotent across recompute (each layer writes a unique key).

### `profile_step.py` alignment

`profile_step.py` now mirrors `train.py`'s prefix exactly (master-weights wrap, clip dispatch, post-prepare master migration, LR rebind), so the benchmark numbers below reproduce from main after merge instead of silently using the buggy raw-bf16 path.

## Numbers (4×A100 80GB, libero / pi05, post-fix)

Same hardware, same data, fp32 Adam state, `PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True`, `accum=1`. Full Stage 4 table + verdicts on [#177](https://github.com/TensorAuto/OpenTau/issues/177#issuecomment-4318533005); SDPA + grad-ckpt kernel-level analysis on [this PR's correction comment](https://github.com/TensorAuto/OpenTau/pull/182#issuecomment-4318605785).

| Run | Backend / config | bs/rank | Peak | Throughput | vs B1 |
|---|---|---:|---:|---:|---:|
| B1 | **DDP fp32, eager**, FUP=true | 6 | 78.45 GiB | 30.2 s/s | baseline |
| B3 | DeepSpeed ZeRO-2, gas=1 | 14 | 78.75 GiB | 15.4 s/s | DDP/B3 = **1.96×** |
| R2 | DDP, **SDPA**, FUP=false | 7 | 77.96 GiB | 31.6 s/s | +4.6% |
| R3 | DDP, **grad-ckpt**, FUP=false | 18 | 78.15 GiB | 32.9 s/s | +8.9% |
| **R4** | DDP, **SDPA + grad-ckpt**, FUP=false | 19 | 79.22 GiB | **33.3 s/s** | **+10.3%** |

PR #176's headline 2.90× was measured at 8 GPUs under the bf16 Adam state bug at bs=12 — a config only feasible because of this issue's bug. The corrected same-hardware DDP-over-ZeRO-2 ratio at 4 GPUs is **~1.96×**, with another **+10.3%** on top from SDPA + grad-ckpt opt-ins (R4, also at 4 GPUs).

## Files changed

- `src/opentau/optim/master_weights.py` (**new**, ~430 lines): `MasterWeightOptimizer` — fp32 masters, subclass `Optimizer`, `from_existing`, `rebuild_masters_from_live` with device migration, `step` / `zero_grad` / `clip_grad_norm_` / `state_dict` / `load_state_dict` / `add_param_group`.
- `src/opentau/scripts/train.py`: optimizer wrap + LR rebind + post-prepare master migration + grad-ckpt unsupported-backend startup guard + clip-grad dispatch.
- `src/opentau/scripts/profile_step.py`: full prefix alignment with `train.py` so post-merge benchmarks reproduce.
- `src/opentau/policies/pi05/paligemma_with_expert.py`: SDPA backend dispatch + `_run_layer` extraction for grad-ckpt.
- `src/opentau/policies/pi05/configuration_pi05.py`: `gradient_checkpointing` flag (default False); `attention_implementation` now wired (default `"eager"`).
- `src/opentau/policies/pi05/modeling_pi05.py`: thread `gradient_checkpointing` into the model config.
- `tests/optim/test_master_weights.py`: 13 CPU tests + 3 GPU tests (device migration).

Not touched: any config file, any other policy, `src/opentau/optim/optimizers.py`.

## Testing

- `pytest -m "not gpu" tests/optim/test_master_weights.py`: 13/13 passed locally in <1s.
- Full Stage 4 matrix re-run on `gpu-387` (4×A100 80GB) post-cherry-pick: nvidia-smi peaks identical to pre-cherrypick (78.45 / 78.75 / 78.43 / 77.96 / 78.15 / 79.22 / 77.96 GiB across B1/B3/R1/R2/R3/R4/R5); throughput within ±1% noise. The new commits don't regress baselines.
- SDPA loss-equivalence A/B (post-cherrypick, 1020 steps each, bs=6, same seed): median |relative delta| 0.82% — PASS 2% tolerance. End-of-run loss eager 1.0356 vs sdpa 1.0099 (equivalent within bf16 noise).
- `pre-commit run --all-files`: clean (ruff / ruff-format / pyupgrade / typos / bandit / gitleaks / zizmor / license-header / eof / trailing-whitespace).

## How to checkout & try

```bash
git fetch origin claude/verify-precision-issue-SjtJf
git checkout claude/verify-precision-issue-SjtJf

# Unit tests (CPU, <1s):
pytest -m "not gpu" tests/optim/test_master_weights.py

# End-to-end: existing libero recipe still works as-is (eager attention + no ckpt).
opentau-train \
    --accelerate-config=configs/libero/reproduce_pi05_libero_accelerate_config.yaml \
    --config_path=configs/libero/reproduce_pi05_libero.json

# Opt into SDPA + grad-ckpt for the +10.3% throughput stack:
# Edit configs/libero/reproduce_pi05_libero.json under "policy":
#   "attention_implementation": "sdpa",
#   "gradient_checkpointing": true,
# Then re-run with --batch_size=19 --dataloader_batch_size=19.
```

## Checklist

- [x] My PR includes policy-related changes (SDPA backend + grad-ckpt flag on pi05's `PaliGemmaWithExpertModel`; the `MasterWeightOptimizer` wrap also affects every policy that goes through `train.py`).
  - [x] GPU pytests added (`@pytest.mark.gpu` device-migration regressions in `tests/optim/test_master_weights.py`); 4-GPU Stage 4 matrix on cluster confirms throughput parity within ±1% and memory peaks exact across all configurations; loss-equivalence A/B verifies SDPA produces equivalent loss curves to eager.

Refs: #181, #177, #176


shuheng-liu:
### Pre-flight audit before flipping out of draft

Re-audited the integration paths against the current diff. Two latent bugs found and fixed in `8be2cd1`; two lower-severity items documented for awareness.

#### Bug 1 — CRITICAL (fixed): LR schedule never reaches the inner optimizer

`make_optimizer_and_scheduler` (`src/opentau/optim/factory.py:32-33`) builds the scheduler with `cfg.scheduler.build(optimizer, ...)`, so `lr_scheduler.optimizer` permanently points at the original `torch.optim.AdamW`. `MasterWeightOptimizer.from_existing` discards that AdamW and builds a fresh inner with new `param_groups` dicts.

`torch.optim.lr_scheduler.LRScheduler.step` (`lr_scheduler.py:296`):
```python
for param_group, lr in zip(self.optimizer.param_groups, values, strict=True):
    param_group["lr"] = lr
```
…mutates the **orphaned** AdamW's groups. Our wrapper's inner AdamW reads its own `param_groups[i]["lr"]` on every step and never sees the schedule. `accelerate.scheduler.AcceleratedScheduler` does **not** rebind `scheduler.optimizer`. Effective LR was frozen at the construction value forever; `train_metrics.lr = optimizer.param_groups[0]["lr"]` read the wrapper's correct (frozen) value, so wandb would show a flat line that looks intentional. The DeepSpeed path was unaffected (we don't wrap there).

**Fix.** Rebind `lr_scheduler.optimizer = wrapped` immediately after `from_existing`, before `accelerator.prepare`. The wrapper's `param_groups` property proxies to the inner, so scheduler mutations land on the right dicts. `accelerate.Accelerator.prepare_scheduler` (`accelerator.py:2806-2810`) then matches the rebound optimizer against `_optimizers` for `AcceleratedScheduler.step_was_skipped` — this is also the moment Bug 2 has to be fixed for the match to succeed.

#### Bug 2 — HIGH, latent (fixed): wrapper wasn't recognised as a torch Optimizer

`MasterWeightOptimizer` deliberately did **not** subclass `torch.optim.Optimizer` (the previous class docstring said as much). But `accelerate.Accelerator._prepare_one` (`accelerator.py:1403-1404`) is a strict `isinstance` check:
```python
elif isinstance(obj, torch.optim.Optimizer):
    optimizer = self.prepare_optimizer(obj, ...)
```
Our wrapper fell through and was returned **unchanged** from `Accelerator.prepare` — never wrapped in `AcceleratedOptimizer`. Consequences:

- **Grad accumulation broken.** `AcceleratedOptimizer.step` and `.zero_grad` both gate on `gradient_state.sync_gradients` (`optimizer.py:124-127, 161`); without the wrap, our raw step/zero_grad ran on every micro-batch, collapsing `cfg.gradient_accumulation_steps > 1` to effective 1.
- **Wrapper missing from `Accelerator._optimizers`.** `prepare_scheduler`'s identity match (`accelerator.py:2807`) couldn't find it, and `AcceleratedScheduler` was constructed with an empty optimizer list (silent no-op on `step_was_skipped`).

DDP example config has `accum=1`, so neither bug surfaced in our previous A/B; the libero recipe also uses `accum=1`. Hidden landmine.

**Fix.** Subclass `torch.optim.Optimizer` without calling `super().__init__()`. `isinstance` now passes; `param_groups`/`state`/`defaults` already proxy to `self.inner` via existing `@property` accessors. `AcceleratedOptimizer` itself uses this exact skip-super pattern in production, so the approach is proven safe.

#### Bug 3 — MEDIUM (documented only): pre-fix checkpoint resume is silently lossy

Old checkpoints (saved before this PR, when the optimizer was over bf16 params) have bf16 `exp_avg` / `exp_avg_sq`. `torch.optim.Optimizer.load_state_dict` auto-casts state tensors to match the current param dtype (now fp32) — so loads succeed, but the bf16 mantissa precision is what you get; the missing fp32 bits are gone for good. Functional, not a crash. Worth a note for anyone resuming a pi05 / libero run from a pre-#176 checkpoint.

#### Bug 4 — LOW (documented only): per-resume precision drift

Optimizer `state_dict` only saves Adam state, not param tensor values. `accelerator.load_state` restores **bf16** live weights; `rebuild_masters_from_live` then re-derives fp32 masters from those bf16 weights, losing the fp32 mantissa bits accumulated since the last checkpoint. Bounded (one round-off per resume). DeepSpeed's `BF16_Optimizer` saves the fp32 masters explicitly and avoids this; we don't, which is the documented tradeoff.

#### Tests added (CPU, fast — 13 total now, all green locally)

- `test_wrapper_is_torch_optimizer_subclass`: pins Bug 2.
- `test_lr_scheduler_reaches_inner_after_from_existing_with_rebind`: positive case for Bug 1 — runs LinearLR through the prefix that `train.py` executes and asserts the inner's lr decays.
- `test_lr_scheduler_does_not_reach_inner_without_rebind`: negative control documenting the failure mode if a future change drops the rebind line.

Pushed in `8be2cd1`. CI should pick it up shortly. Ready for review once green.

---

shuheng-liu:
## Re-bench on `gpu-387` after the cherry-picks: baselines preserved within ±1%, Stage 4 winners hold

Pulled SDPA backend, gradient-checkpointing flag, and `profile_step.py` master-weights alignment onto this branch (commits 2a42db2..5bb7f30). Re-ran the full benchmark matrix on 4×A100 to verify the new commits don't regress baseline numbers and to land post-fix Stage 4 measurements.

### Sanity check — pre-cherry-pick vs post-cherry-pick (memory peaks exact, throughput within ±1%)

| Run | bs | Peak | Throughput (pre / post) |
|---|---:|---:|---:|
| B1 (DDP eager, FUP=true) | 6 | 78.45 GiB | 30.2 / **29.9** s/s |
| B3 (ZeRO-2, gas=1) | 14 | 78.75 GiB | 15.4 / **15.3** s/s |
| R1 (FUP=false) | 6 | 78.43 GiB | 30.3 / **30.1** s/s |
| R2 (SDPA bs=7) | 7 | 77.96 GiB | 31.6 / **31.4** s/s |
| R3 (grad-ckpt bs=18) | 18 | 78.15 GiB | 32.9 / **32.6** s/s |
| R4 (SDPA + ckpt bs=19) | 19 | 79.22 GiB | 33.3 / **33.3** s/s |
| R5 (FUP=true SDPA bs=7; control) | 7 | 77.96 GiB | 31.6 / **31.4** s/s |

The `Wrapped optimizer with MasterWeightOptimizer (fp32 master weights + fp32 Adam state)` log line fires on every DDP-path run; eager B1 lands at the same 78.45 GiB peak we measured before, so the wrap is correctly allocating fp32 masters on GPU (the broken-CPU-master case showed 33 GiB at bs=4 — definitely not what we have here).

### Stage 4 verdicts (now first-class on this PR after the cherry-picks)

- **`find_unused_parameters=False` default flip — REJECT.** R1 vs B1 = +0.3% (within noise), R5 vs R2 = identical. The 10–15% claim in `train.py:127` is invalid for this configuration; the autograd-graph walk apparently overlaps with NCCL allreduce. Default stays True. Save a single-line change for a separate PR if anyone wants it.
- **SDPA backend — ACCEPT pending loss-equivalence A/B.** +4.6% throughput at bs=6→7, frees ~5.6 GiB on forward (no S×S materialization). Loss-equivalence A/B (`PROFILE_LOSS_LOG`-based) is the gating step before flipping `PI05Config.attention_implementation` default to `"sdpa"`; not run in this matrix (cancelled to save GPU time after baseline equivalence was confirmed). Default flag stays `"eager"` for now; flip happens in a follow-up after the A/B.
- **Gradient checkpointing — ACCEPT as opt-in.** +8.9% alone at bs=18; +10.3% stacked with SDPA at bs=19. Default flag stays `False` (the safe option for memory-flush configurations like 7B+ pi05 variants where the dynamics may differ); enable via `PI05Config.gradient_checkpointing: True` for libero / 4-GPU / pi05.

### Corrected DDP-over-ZeRO-2 ratio

`B1 / B3 = 30.2 / 15.4 = 1.96×` at 4 GPUs. PR #176's headline 2.90× was measured at 8 GPUs against a buggy bf16-state DDP at bs=12 — a configuration only feasible because of this issue's bug. The bf16 Adam state took ~half the memory of the fp32 state we now correctly allocate, so the per-rank batch had to drop from 12 → 6 under the corrected path, giving the corrected speedup. See [the corrective comment on issue #177](https://github.com/TensorAuto/OpenTau/issues/177#issuecomment-4318533005) for the full retrospective.

### What this PR contains after the cherry-picks

- `master_weights.py` — fp32 master weights wrapper (correctness-primary), subclass `torch.optim.Optimizer` so accelerate wraps it (8be2cd1 / Bug 2), device migration in `rebuild_masters_from_live` (dd1154e / Bug 0 follow-on).
- `train.py` — wrap optimizer + LR-scheduler rebind (8be2cd1 / Bug 1) + post-prepare master migration (dd1154e) + grad-ckpt unsupported-backend startup guard.
- `paligemma_with_expert.py` / `configuration_pi05.py` / `modeling_pi05.py` — SDPA backend opt-in + gradient-checkpointing opt-in + `_run_layer` extraction.
- `profile_step.py` — full alignment with `train.py`'s master-weights wrap path (wrap, clip dispatch, post-prepare migration, LR rebind) so the benchmark numbers above are reproducible from main after #182 merges.
- `tests/optim/test_master_weights.py` — 13 CPU tests (full coverage of construction, step, zero_grad, state_dict round-trip, clip-then-step, `from_existing` invariants, `rebuild_masters_from_live` re-sync, `Optimizer` subclass invariant, LR rebind positive + negative controls); 3 GPU tests for device migration on fresh / resume / param-identity.

CPU tests pass (13/13); pre-commit clean; CI green.

Ready to flip out of draft once you're happy with the scope expansion (correctness-only → correctness + SDPA/ckpt opt-in + benchmark alignment).

---

shuheng-liu:
## Pre-cherrypick AB-eager vs AB-sdpa (1000 measured steps, bs=6)

Throughput + peak memory from the pre-cherrypick A/B I ran this morning (the JSON files themselves were wiped during the matrix-rerun cleanup; the values are preserved in my command transcripts):

| Run | bs | Peak | Step | Throughput |
|---|---:|---:|---:|---:|
| AB-eager-bs6 | 6 | 78.43 GiB | 794 ms | 30.2 s/s |
| AB-sdpa-bs6 | 6 | 72.83 GiB | 797 ms | 30.1 s/s |
| **Δ (sdpa − eager)** | — | **−5.60 GiB** | +3 ms | **−0.3%** |

Same bs, same data, same seed. SDPA is **throughput-neutral at fixed bs** (within noise) and frees 5.6 GiB on forward — consistent with PR #176's analysis that SDPA recomputes the (S×S) scores in backward, so the kernel itself is roughly a wash; the throughput win comes from spending the freed memory on a larger bs (the R2 +4.6% headline = bs=6 eager → bs=7 SDPA, not "SDPA at the same bs is faster").

### Caveats on these numbers

1. **No loss-equivalence verdict.** Both runs above were on the pre-cherrypick branch where Bug 1 (LR scheduler frozen at initial value, fixed in 8be2cd1) was active. The bug was *symmetric* across both runs, so the eager-vs-SDPA comparison is internally valid for kernel equivalence at constant LR — but the loss curves themselves don't reflect what production would converge to under a real warmup schedule. The loss JSONs (`AB-*-losses.json`) were collected but deleted during the matrix-rerun cleanup before I could run [`compare_loss_logs.py`](https://github.com/TensorAuto/OpenTau/blob/claude/post-181-rebench-speedups/src/opentau/scripts/compare_loss_logs.py) on them.

2. **Bug 2 (subclass `Optimizer` / `AcceleratedOptimizer` wrapping, 8be2cd1) was inactive in both runs.** For libero `gradient_accumulation_steps=1` this is observably a no-op: every step is a sync step, so `AcceleratedOptimizer.step` would have called the wrapper's step path on every micro-batch anyway. Memory peaks confirm — eager bs=6 hit the same 78.43 GiB on both pre- and post-cherrypick runs.

To gate the SDPA-default flip cleanly we'd want a loss-equiv A/B on the post-cherrypick branch (~34 min cluster time: 1000 eager + 1000 SDPA at bs=6, same seed). I cancelled the queued runs per your earlier call; happy to re-launch if you want the loss curves on the record before deciding on the default.

---

shuheng-liu:
## Post-cherrypick A/B loss-equivalence — SDPA passes, default flip is safe

Re-ran the eager-vs-SDPA A/B on PR #182 HEAD (post-cherrypick, with all the fixes including the LR scheduler rebind active). Both runs: 1020 steps total (20 warmup + 1000 measured) at bs=6, same seed, same data ordering, `find_unused_parameters=false`, `accelerator.gradient_accumulation_steps=1`.

### Throughput + memory (bs=6, post-cherrypick)

| Run | Peak | Step | Throughput |
|---|---:|---:|---:|
| AB-eager-bs6 | 78.43 GiB | 799 ms | 30.0 s/s |
| AB-sdpa-bs6 | 72.83 GiB | 812 ms | 29.5 s/s |
| **Δ (sdpa − eager)** | **−5.60 GiB** | +13 ms | **−1.7%** |

Same conclusion as the pre-cherrypick A/B: at fixed bs SDPA is throughput-neutral / slightly worse (the backward recompute slightly outweighs the forward savings); the actual win comes from spending the 5.6 GiB headroom on a larger bs (R2 = bs=7 = +4.6%).

### Loss-equivalence (`compare_loss_logs.py`, 1020 measured steps each)

```
reference: AB-eager-bs6-losses.json  attn=eager  bs=6  seed=1000
candidate: AB-sdpa-bs6-losses.json   attn=sdpa   bs=6  seed=1000

ref  loss:  start=5.639005  end=1.035643  mean=1.724058
cand loss:  start=5.643604  end=1.009877  mean=1.717522

per-step delta (cand − ref):
  median   = −4.99e-03
  mean     = −6.54e-03
  stdev    =  2.02e-02
  abs.med  =  1.22e-02
  abs.max  =  9.45e-02

per-step relative delta:
  median   = −0.29%
  abs.med  =  0.82%
  abs.max  =  7.21%

per-component breakdown:
  mse  delta: median=-2.98e-03  abs.med=5.97e-03  abs.max=8.88e-02
  ce   delta: median=-1.39e-03  abs.med=8.11e-03  abs.max=5.96e-02

verdict (rel_tol=2%, abs_tol=1e-3):
  median |relative delta| = 0.82%   → PASS
  median |abs delta|      = 1.22e-2 → FAIL  (bf16 noise at small absolute losses)
  EQUIVALENT: YES
```

The median relative delta is **0.82% — well within bf16 reassociation noise** and below the 2% tolerance. The median absolute delta failing the 1e-3 tolerance is expected (the loss starts at ~5.64 and ends at ~1.0; 1e-3 absolute is too tight at that magnitude — the relative-delta gate is the right one). End-of-run loss is essentially identical (eager 1.0356, sdpa 1.0099 — sdpa slightly lower, within noise; certainly not diverging).

### Bottom line for the SDPA default flip

SDPA produces equivalent loss trajectories to eager at the same seed / data / hyperparameters. The kernel itself is throughput-neutral at fixed bs; the win is the 5.6 GiB memory headroom that lets bs grow (R2 +4.6% at bs=7, R4 +10.3% stacked with grad-ckpt at bs=19).

Default flip from `"eager"` → `"sdpa"` in `PI05Config.attention_implementation` is **safe**. Whether to actually flip is a separate question (the eager kernel is the historical baseline and removes one source of CUDA-version sensitivity); I'd suggest landing this PR with `"eager"` still default and the flag now wired up + working, so users opt in to SDPA explicitly while we accumulate field experience.

---

shuheng-liu:
## Correction: SDPA + grad-ckpt **does** outperform eager + grad-ckpt at fixed bs

In [the loss-equivalence comment](https://github.com/TensorAuto/OpenTau/pull/182#issuecomment-4318601215) I claimed the SDPA kernel is throughput-neutral at fixed bs and that the win comes only from spending freed memory on a larger bs. That's only true in the **no-ckpt** case. With grad-ckpt the dynamic flips and SDPA wins on the kernel itself.

### Per-sample latency (`step_ms / (bs × num_ranks)`, 4 ranks)

| Config | bs | Step | **ms/sample** |
|---|---:|---:|---:|
| Eager, no ckpt | 6 | 799 ms | 33.29 |
| SDPA, no ckpt | 6 | 812 ms | 33.83  *(eager wins +1.6%)* |
| Eager + ckpt | 18 | 2189 ms | 30.40 |
| SDPA + ckpt | 19 | 2281 ms | **30.01** *(SDPA wins +1.3%)* |

### Why the sign flips

- **No ckpt**: SDPA's backward recomputes the (S×S) attention scores from scratch (FlashAttention-2 doesn't save them in forward — that's where SDPA's forward memory savings come from). Eager's backward reads the saved scores. Net: SDPA backward is doing strictly more compute than eager backward, so kernel-level SDPA is ~1-2% slower per sample at fixed bs.

- **With ckpt**: both backends already recompute the full forward during backward (that's what grad-ckpt does — drop activations, redo forward in backward). SDPA's lighter forward (no S×S materialization) means there's just less to recompute. Net: SDPA+ckpt is ~1-2% faster per sample than eager+ckpt at fixed bs.

So the recommended stack is in fact **SDPA + grad-ckpt** (R4 = 33.3 s/s at bs=19, +10.3% over B1 baseline), not just SDPA on its own:
- The kernel is faster per sample (this comment).
- The memory headroom (5.6 GiB at bs=6 → bigger savings at bs=19) lets the bs grow further than eager + ckpt allowed (bs=18).

Both effects compound. Correcting the earlier framing — sorry for the misleading "throughput-neutral" claim, that was a no-ckpt-only observation.

---

shuheng-liu (commented):
## Code Review

### Overview
Three landings in one PR:

1. **Correctness (primary, #181):** New `MasterWeightOptimizer` (~440 lines) wraps any `torch.optim.Optimizer` to keep fp32 master copies + fp32 Adam state for the DDP / single / FSDP path, mirroring DeepSpeed's `BF16_Optimizer` semantics. Wired into `train.py` (gated by `distributed_type != DEEPSPEED`) with three follow-up fixes for LR-scheduler rebind, accelerate `isinstance` recognition, and post-`accelerator.prepare` device migration of the fp32 masters.
2. **Throughput levers (opt-in, #177):** A `sdpa_attention_forward` path on `PaliGemmaWithExpertModel` (default off — `attention_implementation="eager"`) and a per-layer `gradient_checkpointing` flag on `PI05Config` (default off), with a strict startup guard in `train.py` rejecting the latter on ZeRO-3 / FSDP.
3. **Bench reproducibility:** `profile_step.py` now mirrors `train.py`'s exact prefix (master-weights wrap, scheduler rebind, post-prepare device migration, clip dispatch).

### Code quality / correctness

The wrapper is well thought out and the docstrings are unusually thorough. Highlights:

- **Subclass + skip-`super().__init__()` trick** (`master_weights.py:38`, `:81-87`) is exactly the pattern accelerate's own `AcceleratedOptimizer` uses — mandatory for `Accelerator._prepare_one`'s strict `isinstance(_, torch.optim.Optimizer)` check; pinned by `test_wrapper_is_torch_optimizer_subclass`.
- **`_upcast_done_this_step` flag** correctly avoids the double-upcast when `clip_grad_norm_` runs before `step` (the typical accelerate flow). Reset in `step` (post-call), `zero_grad`, and `load_state_dict`. `test_clip_then_step_does_not_double_upcast` verifies the key path.
- **`master.data = new_data` (`:380`) preserves Parameter identity** so the inner AdamW's `param_groups[i]["params"]` references stay valid post-migration; the GPU test `test_rebuild_masters_preserves_parameter_identity_across_migration` pins this.
- **Adam-state device migration** (`:372-375`) before the `master.data` re-point is the right order — without it the next `step` would crash on a device-mismatch `addmul_`. Pinned by `test_rebuild_masters_migrates_populated_state_and_step_succeeds`.
- **Two LR-scheduler rebind tests** (positive + negative control documenting the failure mode) are excellent regression armor for a bug that's silent in wandb (flat LR line that "looks intentional").

The bug analysis in the PR description is meticulous and matches the code; the loss-equivalence A/B + post-fix benchmark numbers are reassuring evidence.

### Specific suggestions / nits

#### 1. `%%` in `logging.warning` will display literally as `%%` (cosmetic)
`paligemma_with_expert.py:230`

```python
logging.warning(
    "attention_implementation='fa2' is not implemented; falling back to 'eager'. "
    "Consider switching to 'sdpa' for ~10-15%% better throughput."
)
```

stdlib `logging` only does `%`-formatting when args are passed; with no args the message is logged verbatim, so users will see `10-15%% better`. Verified locally. Should be `%`.

#### 2. `PI05Config` docstring lags the new `"sdpa"` value
`configuration_pi05.py:64` still reads `"eager" or "fa2"`; the validator now accepts `"sdpa"` too. One-line doc fix.

#### 3. `gradient_checkpointing` guard not duplicated in `profile_step.py`
The PR comments call this out explicitly — but `profile_step.py` is precisely the script most likely to be invoked under exotic backends during perf debugging. The guard is ~12 cheap lines; I'd argue duplication is worth it (or extract a helper). Not a blocker.

#### 4. `from_existing` ordering hazard for future policies
`rebuild_masters_from_live(policy.parameters())` filters by `requires_grad` and assumes the resulting order matches the wrapper's internal `_fp32_params`. Today this is safe because `pi05.get_optim_params()` is `self.parameters()`. If a future policy returns a re-ordered subset (or splits into named groups), masters would silently misalign — the length check at `master_weights.py:354-358` won't catch it. Two cheap mitigations:
- Prefer the no-arg form `rebuild_masters_from_live()` in `train.py` / `profile_step.py` (uses captured `_live_params`, immune to ordering since `accelerator.prepare` migrates `.data` in-place rather than swapping Parameter objects), OR
- Add an `id()`-set equality check in the with-arg branch.

#### 5. `getattr(optimizer, "optimizer", optimizer)` repeated four times
`train.py` (3×) + `profile_step.py` (2×). Worth extracting to `def _unwrap(optim) -> MasterWeightOptimizer | None:` returning the wrapper if present, else None — readability + one place to maintain the `AcceleratedOptimizer` unwrap pattern.

#### 6. Grad-checkpoint + KV-cache mutation correctness is subtle
`_run_layer` writes `past_key_values[layer_idx] = {...}` and the comment claims idempotency "because each layer writes its own unique key". This holds, but the deeper reason it's safe is **autograd's topological traversal**: suffix-forward checkpoints recompute first (consuming the original prefix-forward `pkv[K]` tensors), then prefix-forward checkpoints recompute (overwriting `pkv[K]` with fresh tensors). If autograd ever batched/reordered, a prefix recompute writing `pkv[K]` before all suffix recomputes had read it would silently corrupt gradients. The empirical loss-equivalence A/B (median 0.82% relative delta) is the only thing pinning this — a focused test (e.g. forward+backward with grad_ckpt vs. without, asserting bit-identical fp32 grads at lr=0) would harden it.

#### 7. `add_param_group` is implemented but untested
`master_weights.py:422`. Since the train path doesn't call it, it's not exercised in production today, but it's part of the public optimizer protocol and accelerate-internal codepaths may invoke it. A one-shot test (add a fresh `nn.Linear`'s params to a wrapper, step, assert fp32 state) would close this gap.

#### 8. Minor style
- `_run_layer` re-derives `models = [self.paligemma.language_model, self.gemma_expert.model]` per call (`paligemma_with_expert.py:475`); it's hot but cheap. Fine.
- The 11 positional args into `torch.utils.checkpoint.checkpoint(self._run_layer, ...)` is right at the readability ceiling; a kwarg dispatch via `**` or a small dataclass would help. Not blocking.

### Performance / risks

- **Memory accounting.** Now correct: fp32 masters live on GPU (was CPU pre-`dd1154e`), so nvidia-smi peaks reflect the true Adam-state cost. The PR's 78–79 GiB peak numbers at 4×A100 are now believable.
- **Default behavior unchanged.** SDPA + grad-ckpt are both opt-in; existing configs run identically. The DDP correctness fix changes math under bf16+DDP — that's the *whole point* and is what the PR description and #181 motivate.
- **Resume path.** `accelerator.load_state` → `rebuild_masters_from_live` is the right order; tested at the unit level. End-to-end resume not tested in CI but the unit composition is covered.
- **Throughput claims.** R4 (+10.3%) and 1.96× DDP-vs-ZeRO-2 ratio sit on the post-fix baseline, which is the honest comparison after retracting #176's bf16-Adam-state-bug-induced numbers — good methodology.

### Test coverage

13 CPU + 3 GPU tests, all narrowly targeted. Strong on Adam-state correctness, master ↔ live sync, scheduler rebind (with negative control), device migration. Gaps noted above: `add_param_group`, grad-ckpt loss-equivalence pin, ordering hazard in `rebuild_masters_from_live(policy.parameters())`.

### Security

Nothing relevant — pure numeric / device-management code. No new I/O, no new deserialization paths.

### Verdict

**LGTM modulo cosmetic items #1 and #2.** Other suggestions are non-blocking and can land as follow-ups; #4 and #6 are the ones I'd most want to see hardened in a future PR. The correctness bug is real and the fix is principled; the rebind / `isinstance` / device-migration follow-ups show good debugging hygiene; the throughput levers are properly opt-in with a guard against the unsafe backends.

---

shuheng-liu:
### Triage on the remaining review items

Items #1 and #2 landed in `91c133f`. For the rest, filed four follow-up issues for the items I think are worth tracking; skipped two.

**Filed:**

- **#3 → #200** — [profile_step.py missing the gradient_checkpointing backend guard that train.py has](https://github.com/TensorAuto/OpenTau/issues/200). Reviewer's point about `profile_step.py` being precisely where someone would iterate on configs across exotic backends is right; the guard is cheap and the silent-wrong-gradient-or-hang failure mode under ZeRO-3 is exactly what we should fail loudly on.
- **#4 → #201** — [MasterWeightOptimizer.rebuild_masters_from_live: ordering hazard for future policies](https://github.com/TensorAuto/OpenTau/issues/201). Reviewer flagged this as one of the two items most worth hardening. The length check at `master_weights.py:354-358` would catch a count mismatch but not a re-ordering, and re-ordering would silently corrupt Adam updates. Recommended switching call sites to the no-arg form (uses captured `_live_params`, immune to ordering) plus optionally an `id()`-set assertion in the with-arg branch.
- **#6 → #202** — [pi05 grad-ckpt + KV cache: pin gradient-equivalence with a focused regression test](https://github.com/TensorAuto/OpenTau/issues/202). Reviewer's other "future PR" pick. The loss-equiv A/B (median 0.82% relative delta over 1020 steps) is currently the only thing pinning the KV-mutation idempotence under autograd's checkpoint recompute order. A per-tensor `assert_close(p_eager.grad, p_ckpt.grad)` regression test resolves any per-parameter mismatch that would otherwise average out at the loss-trajectory level.
- **#7 → #203** — [MasterWeightOptimizer.add_param_group: add a regression test](https://github.com/TensorAuto/OpenTau/issues/203). Public-protocol method, currently zero coverage. Trivial CPU test closes the gap.

**Skipped:**

- **#5** (extract `getattr(optimizer, "optimizer", optimizer)` to a helper). The pattern appears in five places and is a one-line idiom whose intent is clear at the call site. A helper saves four characters per call but adds an indirection that anyone debugging an `AcceleratedOptimizer` wrapping issue then has to chase. The bookkeeping (file an issue, review the PR, merge) outweighs the maintenance value here. If we ever grow more wrapping layers (FSDP-prepared optimizer, MS-AMP, etc.) a helper becomes useful — happy to file then.
- **#8** (style — `_run_layer` re-derives lookups; 11 positional args to `checkpoint`). Reviewer themselves marked these as "Fine" / "Not blocking" rather than as fix-me items. The 11-arg checkpoint call is ugly but `torch.utils.checkpoint.checkpoint`'s signature constrains us — wrapping in a dataclass means re-deriving the call site at every entry. Not worth the indirection.

Items #1–#7 represent the substantive review feedback; the four issues filed here capture every non-blocker the reviewer expressed concern about.

---

shuheng-liu:
Claude review passed

akshay18iitg (dismissed):

shuheng-liu:
@claude fix the merge conflicts

shuheng-liu:
### CPU Tests failure — pre-existing main breakage, not from this PR

The collection error from [run 25134364602 / 25134366386](https://github.com/TensorAuto/OpenTau/actions/runs/25134366386/job/73668713730):

```
ImportError while importing test module 'tests/policies/test_pi07_paligemma_low_level_planner.py'.
Traceback:
  src/opentau/policies/pi07_paligemma/low_level_planner/modeling_pi07_low_level.py:59:
    from opentau.policies.pi05_mem.video_encoder import VJEPA2VideoEncoder
ImportError: cannot import name 'VJEPA2VideoEncoder'
    from 'opentau.policies.pi05_mem.video_encoder'
```

…is not introduced by this PR. Verified on `origin/main` directly:

- `src/opentau/policies/pi07_paligemma/low_level_planner/modeling_pi07_low_level.py:59,1053` — imports / instantiates `VJEPA2VideoEncoder`.
- `src/opentau/policies/pi05_mem/video_encoder.py` — defines only `SpaceTimeSiglipVideoEncoder` (no `VJEPA2VideoEncoder`).

Main got into this state via the interleaving of two recently-merged PRs:

- `dfe8261` ([#167](https://github.com/TensorAuto/OpenTau/pull/167) — pi07 paligemma) added the import + the test file at the time `VJEPA2VideoEncoder` still existed.
- `bfc9d93` ([#171](https://github.com/TensorAuto/OpenTau/pull/171) — V-JEPA2 → space-time SigLIP) renamed the class but did not update the pi07 import site or the test file.

The tests inside the file are `@pytest.mark.gpu` so they would be deselected in CPU CI — but the **module-level import** fails before pytest reaches the markers, so collection itself fails.

`cpu_test.yml`'s deselect list does not exclude `tests/policies/test_pi07_paligemma_low_level_planner.py`. So main's CPU CI should also be red on the same error after #171 merged. (Worth checking the most recent main CI run to confirm — if it's been red for a few days, this is unblocked work.)

#### Not fixing in #182

The fix is renaming the import in `pi07_paligemma/low_level_planner/modeling_pi07_low_level.py` (and adjusting the constructor signature if it changed in #171 — needs checking). That's pi07-paligemma work, not master-weights / DDP-precision work. Out of scope for #182.

Two practical options to unblock this PR:

1. **Land a small fix on main** (rename `VJEPA2VideoEncoder` → `SpaceTimeSiglipVideoEncoder` at the two call sites in `modeling_pi07_low_level.py`, plus any ctor changes), then merge main into this branch again — CI goes green.
2. **Add `--ignore=tests/policies/test_pi07_paligemma_low_level_planner.py` to `cpu_test.yml`** as a temporary workaround. Smaller, but kicks the can.

Strongly prefer option 1. Happy to file the issue / open the small fix PR if you want — just say the word.

Everything #182-specific is green: pre-commit passes, all 13 master-weights CPU tests pass locally on the post-merge tree, no master-weights-related collection errors. Once main is unbroken this PR's CI follows.

---

akshay18iitg (approved):

PR #204, opened 2026-04-29
feat: surface robot_type and control_mode in standard data format

shuheng-liu:
## What this does

Adds two optional string fields — **`robot_type`** and **`control_mode`** — to the dict returned by every dataset's `__getitem__`. Both are sourced from `self.meta.info` and fall back to the empty-string sentinel `""` when absent, matching the existing optional-string convention used by `response`, `memory`, and `next_memory`.

* **`robot_type`** is already a first-class field in standard LeRobot v2 `meta/info.json` (e.g. `"aloha"`, `"panda"`, `"human"`); previously reachable only via `LeRobotDataset.meta.robot_type`.
* **`control_mode`** was added to `meta/info.json` in #183 with allowed values `{"joint", "ee", "mixed"}`; previously reachable only via `LeRobotDataset.control_mode` / `LeRobotDataset.meta.control_mode`.

Promoting them to per-sample fields gives multi-embodiment / control-mode-aware policies a uniform signal that survives the dataloader, dataset mixture, and collator. They are **not** subject to training-time dropout — they're dataset-level identifiers, constant per dataset; like `prompt` / `loss_type`. The existing missing-`control_mode` warning at `LeRobotDataset.__init__` and the `LeRobotDatasetMetadata.control_mode` property's `"unknown"` default are unchanged (they serve curation, not the data format).

VQA datasets emit `""` for both, since `VQADatasetMetadata.info` carries neither key — the standard format stays uniform across robot and vqa datasets.

🗃️ Feature

## How it was tested

* Added `TestRobotTypeControlMode` in `tests/datasets/test_optional_keys.py` — 5 unit tests covering: both fields present (pass-through), both absent (`""` sentinel), `robot_type=None` (allowed by the type signature), every drop probability cranked to 1.0 with dropout enabled (pass-through still holds), and Python-`str` typing (default PyTorch collate compatibility).
* Added `test_robot_type_and_control_mode_in_meta_info` in `tests/datasets/test_datasets.py` — integration check that the LeRobotDataset factory threads `info.json` values into `ds.meta.info` for both populated and missing cases.
* Extended `_DummyBaseDataset` in `tests/datasets/test_optional_keys.py` to accept a `meta_info` dict so existing direct-emission tests keep working without setup changes.
* `pytest -m "not gpu and not slow" -n auto tests/datasets/` → **310 passed, 7 skipped, 0 failed** (the 7 skipped are the existing GPU/slow markers; nothing regresses).
* `pytest -m "not gpu and not slow" -n auto tests/scripts/test_attach_metadata.py tests/configs/test_default.py` → **31 passed** (adjacent suites that exercise the standard format end-to-end).
* `pre-commit run --files <four changed files>` → clean.

Documentation: extended the **Optional Standard-Format Keys** section in `docs/source/concepts.rst` to document the two new keys, the `""` sentinel, and the dropout exemption.

```bash
pytest -sx tests/datasets/test_optional_keys.py::TestRobotTypeControlMode \
          tests/datasets/test_datasets.py::test_robot_type_and_control_mode_in_meta_info \
          tests/datasets/test_datasets.py::test_control_mode_warning_emitted_once_per_repo
```

```bash
pytest -m "not gpu and not slow" -n auto tests/datasets/
```

## Checklist

### DIFF
--- a/src/opentau/datasets/lerobot_dataset.py
+++ b/src/opentau/datasets/lerobot_dataset.py
@@ -820,15 +820,10 @@ class BaseDataset(torch.utils.data.Dataset):
         else:
             standard_item["obs_history_is_pad"] = torch.tensor([False], dtype=torch.bool)
 
-        # Emit optional keys (memory, next_memory, speed, mistake, quality, subgoalK)
-        # with their _is_pad siblings, applying training-time dropout rolls.
+        # Emit optional keys (memory, next_memory, speed, mistake, quality,
+        # robot_type, control_mode, subgoalK) with their _is_pad siblings,
+        # applying training-time dropout rolls.
         self._emit_optional_keys(item, standard_item)
-
-        # Dataset-level identifiers, surfaced per-sample so policies/collators
-        # can read them uniformly. "" signals "field absent in meta/info.json"
-        # (same convention as response / memory / next_memory).
-        standard_item["robot_type"] = self.meta.info.get("robot_type") or ""
-        standard_item["control_mode"] = self.meta.info.get("control_mode") or ""
 
         # cast all tensors in standard_item to bfloat16
         for key, value in standard_item.items():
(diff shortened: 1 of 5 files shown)


src/opentau/datasets/lerobot_dataset.py
```diff
@@ -824,6 +824,12 @@ def _to_standard_data_format(self, item: dict) -> dict:
         # with their _is_pad siblings, applying training-time dropout rolls.
         self._emit_optional_keys(item, standard_item)
 
+        # Dataset-level identifiers, surfaced per-sample so policies/collators
+        # can read them uniformly. "" signals "field absent in meta/info.json"
+        # (same convention as response / memory / next_memory).
+        standard_item["robot_type"] = self.meta.info.get("robot_type") or ""
+        standard_item["control_mode"] = self.meta.info.get("control_mode") or ""
+
```
akshay18iitg: Can we add this in self._emit_optional_keys and also have the drop rate for these fields in same format as speed and mistake

akshay18iitg (changes requested):

src/opentau/datasets/lerobot_dataset.py
shuheng-liu: Agreed.
@claude fix this.

shuheng-liu:
@claude fix this according to the code review.

akshay18iitg (approved):

PR #225, opened 2026-05-01
Deprecate DatasetConfig.val_split_ratio in favor of mixture-level

shuheng-liu:
## What this does

Backward-compatible alternative to #224.

`val_split_ratio` was declared on **both** `DatasetConfig` and `DatasetMixtureConfig`, but the per-dataset value was always silently overwritten by the mixture-level value inside `DatasetMixtureConfig.__post_init__`:

```python
# old configs/default.py
for dataset_cfg in self.datasets:
    dataset_cfg.val_split_ratio = self.val_split_ratio
```

So setting `val_split_ratio` on a `DatasetConfig` already had no effect — it was misleading dead config surface.

This PR removes the duplication **without breaking any pre-existing JSON config**:

- Keeps `val_split_ratio` as a field on `DatasetConfig` (default `0.05`, same as before) so old configs still parse — but flags it as **deprecated** in the docstring and points users at the mixture-level field.
- Replaces the silent propagation loop in `DatasetMixtureConfig.__post_init__` with a `DeprecationWarning` that fires only when a child's value diverges from the mixture's. This actually surfaces the previously-silent override behavior to the user.
- Routes `factory.make_dataset` to read `train_cfg.dataset_mixture.val_split_ratio` directly — the mixture is now the single source of truth for the split.

Behavior preserved end-to-end:
- Old JSON configs that set `val_split_ratio` only on `DatasetMixtureConfig` → unchanged.
- Old JSON configs that set `val_split_ratio` on a child `DatasetConfig` → still parse; the mixture's value still wins (as before); now also see a deprecation warning instead of silent override.
- Old JSON configs that omit it everywhere → both default to `0.05`, no warning.

Tag: 🧹 Cleanup / 📝 Documentation

## How it was tested

- `git grep val_split_ratio` confirms the field still exists on `DatasetConfig` (for parser back-compat), the docstring is updated, the warning fires on divergence, and `factory.make_dataset` reads the mixture-level value.
- No JSON config under `configs/`, no test under `tests/`, and no notebook references the per-dataset field, so existing callers see no behavior change other than the new deprecation warning when they explicitly diverge.
- Files parse cleanly (`python -c "import ast; ast.parse(...)"`) on both edited files.
- Pre-commit and pytest weren't run in this environment (no `uv` venv / no `pre-commit` binary); please re-run locally before merging:

```bash
pre-commit run --all-files
pytest -m "not gpu" -n auto
```

```bash
git fetch origin claude/deprecate-dataset-val-split-ratio
git checkout claude/deprecate-dataset-val-split-ratio
pre-commit run --all-files
pytest -m "not gpu" -n auto
```

Quick manual smoke for the deprecation warning:

```python
import warnings
from opentau.configs.default import DatasetConfig, DatasetMixtureConfig

with warnings.catch_warnings(record=True) as w:
    warnings.simplefilter("always")
    DatasetMixtureConfig(
        datasets=[DatasetConfig(repo_id="foo/bar", val_split_ratio=0.2)],
        val_split_ratio=0.1,
    )
    assert any(issubclass(x.category, DeprecationWarning) for x in w)
```

## Checklist

---
_Generated by [Claude Code](https://claude.ai/code/session_01195CkvBEEzJZxghK5HkfBp)_

### DIFF
--- a/src/opentau/configs/default.py
+++ b/src/opentau/configs/default.py
@@ -22,6 +22,7 @@ This module provides default configuration classes for:
 - Evaluation settings and parameters
 """
 
+import warnings
 from dataclasses import dataclass, field
 
 import draccus
@@ -105,9 +106,13 @@ class DatasetConfig:
     robot_type: str | None = None
     control_mode: str | None = None
 
-    # Ratio of the dataset to be used for validation. Please specify a value.
-    # If `val_freq` is set to 0, a validation dataset will not be created and this value will be ignored.
-    # Defaults to 0.05.
+    # DEPRECATED. Set `val_split_ratio` on `DatasetMixtureConfig` instead — the
+    # mixture-level value is the single source of truth and is applied uniformly
+    # to every dataset in the mixture. This per-dataset field is retained only
+    # so that pre-existing JSON configs continue to parse; setting it here has
+    # no effect on the actual split. A `DeprecationWarning` is emitted by
+    # `DatasetMixtureConfig.__post_init__` whenever a child's value diverges
+    # from the mixture's. Defaults to 0.05 (matches the mixture default).
     val_split_ratio: float = 0.05
 
     def __post_init__(self):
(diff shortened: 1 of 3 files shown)


shuheng-liu:
@claude fix per review

Issue #226, opened 2026-05-02
pi07 low-level planner forward is not byte-equivalent to pi05 when all optional inputs are dropped


shuheng-liu:
## Context

When `subgoal_drop_prob = response_drop_prob = metadata_drop_all_prob = metadata_drop_each_prob = 1.0` (e.g. the smoke config below), `PI07LowLevelPlannerFlowMatching.embed_prefix` should produce a prefix that is functionally identical to `PI05FlowMatching.embed_prefix` (continuous state, `predict_response=False`) — same weights → same KV cache → same losses. It does not. Three small divergences in the prefix layout / attention pattern make pi07 emit a different prefix even after #205 landed (which gated the *content* of the optional middle blocks but not the separators around them).

## Repro

`fix/pi07_paligemma_fixes` (currently `9528cfb`). Smoke test: instantiate both policies (`pretrained_path=None`, identical seed), copy `paligemma_with_expert.state_dict()` + the projection layers from pi05 to pi07, cast both to bf16, feed one fixed batch with the same noise / time:

```
camera0/camera1: (B=2, 3, 224, 224)  random in [0, 1]
state:           (B=2, 32)            random in [-1, 1]
actions:         (B=2, 10, 32)        random in [-1, 1]
prompt:          ["pick up the red block"] * 2
no response/subgoal/metadata keys -> all gating False
```

Result on `fix/pi07_paligemma_fixes` head:

```
pi05  MSE=81.16315460  CE=11.03499031   prefix=(2, 681, 2048)
pi07  MSE=81.89469910  CE=11.47132397   prefix=(2, 684, 2048)
diff  MSE=7.31e-01     CE=4.36e-01      +3 prefix tokens in pi07
```

Per-segment byte check (max abs diff between pi05's and pi07's prefix embeddings):

| segment              | pi05 slice  | pi07 slice  | max abs diff |
| -------------------- | ----------- | ----------- | ------------ |
| video / image (T=1)  | `[0:512]`   | `[0:512]`   | **0.000** |
| lang                 | `[512:612]` | `[512:612]` | **0.000** |
| `"State: "`          | `[612:615]` | `[612:615]` | **0.000** |
| state_emb            | `[615:616]` | `[615:616]` | **0.000** |
| pi05 `":\n"` (2)     | `[616:618]` | —           | (different token from pi07's `", "`) |
| pi07 `", "` (2)      | —           | `[616:618]` | |
[...]

Issue #210, opened 2026-04-29
pi07_paligemma low-level planner: import is broken — needs migration to SpaceTimeSiglipVideoEncoder (or restoration of VJEPA2VideoEncoder)


shuheng-liu:
## Summary

After #171 renamed `VJEPA2VideoEncoder` → `SpaceTimeSiglipVideoEncoder` in `pi05_mem/video_encoder.py`, the pi07 paligemma low-level planner was left with stale imports of the now-removed name. The module fails at import time:

```
ImportError: cannot import name 'VJEPA2VideoEncoder'
    from 'opentau.policies.pi05_mem.video_encoder'
```

This affects two sites in `src/opentau/policies/pi07_paligemma/low_level_planner/modeling_pi07_low_level.py`:

- **L59**: `from opentau.policies.pi05_mem.video_encoder import VJEPA2VideoEncoder`
- **L1053**: `self.video_encoder = VJEPA2VideoEncoder(...)` with 8 V-JEPA2-specific kwargs

PR #209 adds `--ignore=tests/policies/test_pi07_paligemma_low_level_planner.py` to `cpu_test.yml` as a tactical CI unblock — but **the production module is still broken**. Anyone trying to instantiate `PI07LowLevelPlannerPolicy` today gets the same `ImportError`. This issue tracks the proper fix.

## Why it's not a one-line rename

Old ctor — self-contained:

```python
VJEPA2VideoEncoder(
    vjepa2_model_name=...,
    num_frames=...,
    crop_size=...,
    num_video_tokens=...,
    vlm_hidden_size=...,
    perceiver_heads=...,
    freeze_encoder=...,
    encoder_dtype=...,
)
```

New ctor — caller-owned vision components:

```python
SpaceTimeSiglipVideoEncoder(
    vision_tower=...,           # passed by reference from caller's PaliGemma
    multi_modal_projector=...,  # ditto
    spacetime_layer_stride=...,
    num_video_tokens=...,
    # ... different config surface entirely
)
```

The new class **does not construct its own backbone**; it wraps and mutates the caller's PaliGemma vision tower in place. pi07 currently builds its own V-JEPA2 model independently of the PaliGemma it owns. To migrate, you have to decide:

[...]

Issue #192, opened 2026-04-27
Replace V-JEPA2 with the MEM space-time SigLIP encoder in π07 PaliGemma low-level planner


shuheng-liu:
## Context

The π07 PaliGemma low-level planner currently uses a V-JEPA2 + Perceiver-reducer video encoder:

- `src/opentau/policies/pi07_paligemma/low_level_planner/modeling_pi07_low_level.py:59` imports `VJEPA2VideoEncoder` from `opentau.policies.pi05_mem.video_encoder`.
- `:1043` constructs it from the `vjepa2_*` config fields (`vjepa2_model_name`, `vjepa2_crop_size`, `vjepa2_num_video_tokens`, `vjepa2_perceiver_heads`, `vjepa2_dtype`).
- `:1081` runs the video through it on every forward pass.

PR #171 replaced the V-JEPA2 path in `pi05_mem` with the MEM paper's actual architecture — a standard SigLIP ViT (from PaliGemma) extended with space-time separable attention every 4th layer, sharing weights with `paligemma_with_expert.paligemma.vision_tower` and `multi_modal_projector` by reference. The π07 low-level planner should follow the same pattern.

## Goal

Rewrite the π07 low-level planner's video encoder to match the architecture landed in #171, in the same manner:

- **No new learnable parameters.** Reuse the SigLIP layer's own `q_proj`/`k_proj`/`v_proj`/`out_proj` weights for the temporal sublayer at every 4th of the 27 SigLIP layers (`[3, 7, 11, 15, 19, 23]`), with an SDPA causal mask over the `T` axis.
- **Shared SigLIP weights.** The video encoder holds `paligemma_with_expert.paligemma.vision_tower` and `multi_modal_projector` by reference (wrapped in a list so `nn.Module` doesn't re-register params). No duplicate download; checkpoints load with `strict=True` and bit-identical state_dict keys.
- **Fixed sinusoidal temporal PE** with `e(current) = 0`. Verify against the π07 low-level history layout: current frame at `t = T-1`, FIFO ordering newest-last, `obs_history_is_pad[:, -1]` always `False`.
- **Past-timestep tokens dropped** after the encoder; only the current frame's 256 tokens are returned, matching a single-frame VLA's image-token count.
- **Single-frame invariance.** With `T=1` short-circuit to the unmodified `SiglipEncoderLayer`, byte-exact reduction (`atol=0`) to `PaliGemmaModel.get_image_features` on the same frame.

## Config changes

[...]

PR #228, opened 2026-05-02
fix(pi07_paligemma): backport prefix gating + cross-att length to feat/pi07 (#226)

shuheng-liu:
## What this does

Backports the combined [#205](https://github.com/TensorAuto/OpenTau/pull/205) + [#226](https://github.com/TensorAuto/OpenTau/issues/226) gating fixes from `fix/pi07_paligemma_fixes` into `feat/pi07`'s pi07_paligemma low-level planner. The full set of bugs that #205 + #226 address are present on `feat/pi07` today (only a partial earlier attempt landed via `2949e73 "Applying pi07_paligemma fixes to pi07"`).

`PI07LowLevelPlannerFlowMatching.embed_prefix` (in `src/opentau/policies/pi07_paligemma/low_level_planner/modeling_pi07_low_level.py`):

- **Content gating (#205).** Wrap the response, subgoal (header + image loop + footer), and metadata blocks in `if has_response / has_subgoal / has_metadata` guards computed at the top of `embed_prefix` from the actual masks. Without this, with `*_drop_prob=1.0` every prefix carried real (non-padded) spurious tokens with `att_masks=[1,...]` causal-block boundaries that corrupted cumsum for every following token. Subgoal header / footer pad-masks are now per-sample via `sample_has_subgoal` so mixed batches stay correct.
- **Separator gating (#226).** State-end now switches between `", "` (when an optional follows) and `":\n"` (when nothing does) — matching pi05's all-dropped layout. Prefix-end (`";\n "`) is wrapped in `if has_any_optional:` so it's skipped entirely when there's no optional list to terminate.
- **`"Action: "` indicator att_masks (#226).** Changed from `[1] + [0]*(N-1)` (one bidirectional block) to `[1]*N` (one causal block per token). The embeddings already matched pi05 byte-for-byte, but `make_att_2d_masks` shifted every following discrete-action position's cumsum by `N-1`, silently moving the `prefix_out` boundary the discrete-action CE reads.
- **`forward()` cross-attention length (#205).** `num_cross_att_tokens` and `prefix_offsets` now subtract both the `"Action: "` indicator length and `discrete_action_max_length`, matching pi05's `discrete_action_indicator_max_length` logic so the action expert sees the same prefix length at train and at inference (where neither block is in the prefix).
- **`prepare_subgoal_images` NameError (#205).** Track `last_subgoal_img` / `last_mask` inside the present-keys loop and guard the `empty_cameras` fallback with `if last_subgoal_img is not None and last_mask is not None:`. Default `subgoal_is_pad` initializer flips from `torch.zeros` to `torch.ones` (a missing key means "padded", not "real").
- **`prepare_metadata` defaults.** Same `torch.zeros` → `torch.ones` flip for `speed_is_pad` / `quality_is_pad` / `mistake_is_pad`.

Also cherry-picked from main: [#209](https://github.com/TensorAuto/OpenTau/pull/209) (CPU CI `--ignore` for the GPU+slow integration test) and [#217](https://github.com/TensorAuto/OpenTau/pull/217) (GPU CI `--ignore` + comment dedupe). These workflow changes are already on `main` but had not propagated to `feat/pi07`, leaving CPU CI broken at test collection because of the unrelated pre-existing `VJEPA2VideoEncoder` import (TODO #210 / blocked on #192).

**Out of scope** (separate concerns on `fix/pi07_paligemma_fixes` worth handling in their own PRs against `feat/pi07`):
- V-JEPA2 → SpaceTime SigLIP video-encoder swap.
- The `";\n "` → `":\n"` prefix-end string cleanup (commit `6e3f82a`).
- Any change to the gemma3 pi07 (the *other* policy on this branch) or the high-level planner.

## How it was tested

- `pre-commit run --files src/.../modeling_pi07_low_level.py tests/.../test_pi07_paligemma_low_level_planner.py` — all hooks pass (ruff, ruff-format, pyupgrade, gitleaks, bandit, etc.).
- `PYTHONPATH=src pytest tests/policies/ -m "not gpu" --ignore=tests/policies/test_pi07_paligemma_low_level_planner.py -q` → **125 passed, 2 skipped, 19 deselected** in 152s. (PYTHONPATH override needed locally because the worktree's pi07/ module isn't installed in the parent venv; CI's `uv sync` step makes this unnecessary.)
- The integration test `tests/policies/test_pi07_paligemma_low_level_planner.py::test_complete_pi07_low_level_pipeline` is `@pytest.mark.gpu @pytest.mark.slow` and ignored in both CPU and GPU CI per #209/#217 (TODO #210 / blocked on #192). The change does **not** alter prefix shapes for the all-populated path that test exercises; it only changes the `"Action: "` indicator's `att_masks` pattern, which is captured and re-fed to `make_att_2d_masks`, so `_verify_vlm_attention_mask` continues to hold. `_verify_position_ids` and `_verify_action_expert_attention_mask` were updated to reflect the new `forward()` formulas.
- **Not run:** GPU integration test (no GPU on this dev machine and the test is CI-ignored anyway). A reviewer with GPU access running the test manually would also need to drop the `--ignore` for that one file.

```bash
gh pr checkout 228
pre-commit run --all-files
pytest -m "not gpu" -n auto
```

To exercise the integration test on a GPU box (oracle / mlbox):

```bash
pytest -m "gpu" -sx tests/policies/test_pi07_paligemma_low_level_planner.py::TestPI07LowLevelPlannerIntegration::test_complete_pi07_low_level_pipeline
```

To sanity-check the gating fires (Python REPL on a box where `pi05_mem.video_encoder.VJEPA2VideoEncoder` exists, or once #192 lands):

```python
# Build a minimal batch with no response/subgoal/metadata keys, call
# model.embed_prefix(...) directly, and confirm:
#   - prefix length shrinks by (response_max_length + subgoal_lead +
#     n_subgoal_cams * subgoal_tokens + comma + metadata_max_length +
#     prefix_end) tokens vs the all-populated case
#   - state-end is ":\n" instead of ", "
```

## Checklist

> GPU pytests checkbox left unchecked because the only relevant GPU test (`test_complete_pi07_low_level_pipeline`) is CI-ignored at import time on `feat/pi07` (pre-existing `VJEPA2VideoEncoder` issue, TODO #210). Happy to run it manually if a reviewer wants — please flag.

### DIFF
--- a/.github/workflows/cpu_test.yml
+++ b/.github/workflows/cpu_test.yml
@@ -102,7 +102,9 @@ jobs:
           python3 -c "import sys; print(sys.path)"
           python3 -c "import libero.libero" && echo "LIBERO config set successfully."
           echo "Running cpu based pytest and generating coverage report..."
-          pytest -m "not gpu" -n auto -v --cov=lerobot/ --cov-report=xml:cpu_test/cpu_test.xml --ignore=tests/planner/test_planner.py --ignore tests/utils/test_libero_utils.py --deselect=tests/envs/test_factory.py::TestMakeEnv::test_make_env_async_vector_env --deselect=tests/envs/test_factory.py::TestMakeEnv::test_make_env_sync_vector_env tests/
+          # TODO(#210): drop --ignore=tests/policies/test_pi07_paligemma_low_level_planner.py once pi07 migrates to SpaceTimeSiglipVideoEncoder (#192).
+          # TODO(#210): drop --ignore=tests/policies/test_pi07_paligemma_low_level_planner.py once pi07 migrates to SpaceTimeSiglipVideoEncoder (#192).
+          pytest -m "not gpu" -n auto -v --cov=lerobot/ --cov-report=xml:cpu_test/cpu_test.xml --ignore=tests/planner/test_planner.py --ignore tests/utils/test_libero_utils.py --ignore=tests/policies/test_pi07_paligemma_low_level_planner.py --deselect=tests/envs/test_factory.py::TestMakeEnv::test_make_env_async_vector_env --deselect=tests/envs/test_factory.py::TestMakeEnv::test_make_env_sync_vector_env tests/
           echo "Pytest execution and coverage report generation completed."
 
       - name: Upload coverage reports
--- a/.github/workflows/cpu_test.yml
+++ b/.github/workflows/cpu_test.yml
@@ -103,7 +103,6 @@ jobs:
           python3 -c "import libero.libero" && echo "LIBERO config set successfully."
           echo "Running cpu based pytest and generating coverage report..."
           # TODO(#210): drop --ignore=tests/policies/test_pi07_paligemma_low_level_planner.py once pi07 migrates to SpaceTimeSiglipVideoEncoder (#192).
-          # TODO(#210): drop --ignore=tests/policies/test_pi07_paligemma_low_level_planner.py once pi07 migrates to SpaceTimeSiglipVideoEncoder (#192).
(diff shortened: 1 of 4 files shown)


shuheng-liu:
Wrong PR. My intention was to fix the issue in `p07` policy but this touches `pi07_paligemma` instead. Will open a new one

PR #214, opened 2026-04-29
feat(pi0, pi06): add SDPA attention + gradient checkpointing options

shuheng-liu:
## What this does

Backports the pi05 PR #182 attention/checkpointing knobs to the two remaining
self-contained engines:

- `pi0/paligemma_with_expert.py` — adds `sdpa_attention_forward`,
  `gradient_checkpointing` config field, `_run_layer` extraction, and the
  `torch.utils.checkpoint.checkpoint(use_reentrant=False)` wrap. fa2 now warns
  and falls back to eager (was: dead-code validation in `__post_init__`).
- `pi06/gemma3_with_expert.py` — same shape, with two pi06-specific bits:
  - SDPA call passes `scale=` from Gemma 3's `query_pre_attn_scalar`.
  - fa2 was previously a hard `NotImplementedError` (cited the interleaved
    local/global mask pattern); π0.6 keeps the same block-causal mask at
    every layer (see existing `TestNoSlidingWindowEnforcement` test), so SDPA
    sees a regular bool mask and works without per-layer shape branching.
    Replaces the raise with a warn-and-fall-back to match pi05/pi0.

Each engine port is one commit so reviewers can isolate the architectures.

The `train.py:181` ZeRO-3/FSDP guard keys off `getattr(cfg.policy,
"gradient_checkpointing", False)` — policy-agnostic, so pi0/pi06 inherit it
automatically.

Defaults stay `attention_implementation="eager"` and `gradient_checkpointing=
False` — matches PR #182's choice of opt-in rollout.

## How it was tested

CPU-only equivalence + smoke tests added to `tests/policies/test_pi0.py` and
`tests/policies/test_pi06.py`:

- Dispatcher returns the right function for `eager` / `sdpa` / `fa2` (latter
  falls back to eager).
- `eager_attention_forward` vs `sdpa_attention_forward` outputs close to 1e-4
  in fp32 with realistic block-causal masks (pi0 + pi06).
- `gradient_checkpointing=True` forward output bit-identical (atol/rtol 1e-6)
  to `=False` for the pi06 engine — locks in that `_run_layer` extraction is
  bit-identical to the original inlined loop body.
- Updated existing `test_bad_attention_implementation_raises` in pi06 to use
  `"garbage"` since `"sdpa"` is now valid.

```bash
pytest -m "not gpu" tests/policies/test_pi0.py tests/policies/test_pi06.py -q
# 48 passed (pi0 + pi06 only)
```

Full CPU policies sweep (excluding pre-existing-broken
[...]

PR #235, opened 2026-05-02
fix(pi06): align with π0.6 model card paper

shuheng-liu:
## What this does

Audited the OpenTau pi06 implementation against the [π₀.₆ Model Card](https://arxiv.org/abs/2511.14759) (Physical Intelligence, Nov 17 2025) and fixed the inconsistencies the paper actually mandates. Out-of-scope items (metadata conditioning — pi07-only; reference pretraining mixture — separate issue) were intentionally skipped per discussion on the planning thread.

### Bugs fixed

1. **Language tokens were bidirectional in `embed_prefix`, paper requires causal.** Paper §2 says *"We keep bidirectional attention among all of the image tokens (as in π₀.₅) but use causal attention among the text tokens"* — an explicit divergence from π₀.₅, whose §III says *"image patch, textual prompt, and continuous action tokens use bidirectional attention."* The pi06 implementation appended language tokens to the same bidirectional block as the image prefix (`att_masks += [0] * num_lang_embs`), so every prompt token saw every other prompt token. Fixed by switching to `[1] * num_lang_embs`, opening one causal block per language token while preserving cross-attention back to the image prefix. Response and FAST blocks were already correct.

2. **`predict_response` docstring did not document its role for the paper's hierarchical design or its dual use as a VQA / grounding response target.** Paper §1 says pi06 *"preserves the hierarchical design of π₀.₅, providing high-level subtask prediction and low-level action generation"*; the path exists in code (`predict_response=True` → autoregressive subtask prediction at inference, then flow-matching conditioned on it) but was undocumented. Default stays `False` (most LeRobot datasets have no subtask annotation) and field name stays as-is (also doubles as the supervised target for VQA / grounding co-training, per π₀.₅ pretraining recipe).

3. **No regression test guarded MSE-loss masking on fully-padded action samples used for web/VQA co-training.** `VQADataset.__getitem__` correctly sets `actions_is_pad = ones(action_chunk)` (zero-padding action data for vision-text-only samples), and `PI06FlowMatching.forward` does `~actions_is_pad` to build `in_episode_bound` and AND it into `postfix_mask` — so the masking already worked. But there was no test locking this in: a future change to either side could silently start training the action expert to regress to zero on web samples. Added a unit test that replicates the masking arithmetic and asserts MSE is exactly 0 for an all-padded item, plus a partial-mask case.

4. **GPU smoke test (`test_complete_pi06_pipeline_integration_smoke`) crashed on a `(50, 32)` vs `(10, 32)` shape mismatch.** The shared `lerobot_dataset_metadata` fixture provides actions stats shaped `(50, 32)` (matching the default `PI06Config.chunk_size`), but the smoke test uses `chunk_size=10` to keep the model small — so the Normalize buffer ended up `(50, 32)` while the test batch's actions were `(B, 10, 32)`, raising `RuntimeError: The size of tensor a (10) must match the size of tensor b (50) at non-singleton dimension 1` inside `MIN_MAX` normalization. Pre-existing bug, never caught in CPU CI because the test is `@pytest.mark.gpu`-gated. Cherry-picked the fix from [#214](https://github.com/TensorAuto/OpenTau/pull/214) (commit [`6425cb4`](https://github.com/TensorAuto/OpenTau/pull/214/commits/6425cb4b490babd7848f2e22519bbe9b196f3301)) so the smoke test now passes on this branch standalone — it deep-copies the fixture stats and reshapes `actions["max" / "mean" / "min" / "std"]` from `(50, 32)` to `(chunk_size, 32)` before constructing the policy.

### Out of scope (filed elsewhere)

- Conditioning-metadata prompt slot — intentionally pi07-only, not a pi06 inconsistency.
- Paper-aligned pretraining-mixture reference config (FAST tokens + multi-modal web data co-training) — will file as a follow-up issue once this PR lands.
- Proprio-state tokenization — verified paper-faithful: π₀.₅ paper §III explicitly says *"The robot proprioceptive state is discretized and input to the model as text tokens"*, which matches the current decimal-text-in-prompt approach. No change needed.

### Relationship to PR #214

The smoke-test fixture fix above (bug 4) is the same commit that lives on [#214](https://github.com/TensorAuto/OpenTau/pull/214) (which also adds SDPA + gradient checkpointing — those parts stay on #214). Whichever PR merges first, the other will see the cherry-picked commit as already applied via patch-id matching, no manual conflict resolution needed.

## How it was tested

- `pre-commit run --files <changed>` — clean.
- `pytest -m "not gpu" -n 0 tests/policies/test_pi06.py` — **29 passed**.
- Full CPU suite: `pytest -m "not gpu" -n auto` — 765 passed; the 5 unrelated failures are pre-existing in `tests/envs/test_factory.py`, `tests/policies/test_pi07_paligemma_low_level_planner.py`, `tests/scripts/test_annotate_subtasks.py`, `tests/utils/test_libero_utils.py` (none touch pi06).
- Added `test_embed_prefix_layout_has_causal_language_block` in `TestMakeAtt2dMasks` — asserts the new prefix layout: image-bidirectional, language-causal (`triu(1).sum() == 0`), language can still attend back to the image prefix, images cannot peek forward at language.
- Added `TestMSEMaskOnFullyPaddedActions` with two cases: all-padded → loss == 0; mixed-padded → loss == MSE averaged over real timesteps only.
- GPU tests not run locally (no CUDA on this host); will run on the nightly `gpu_test.yml` job.

```bash
git fetch origin fix/pi06-paper-alignment && git checkout fix/pi06-paper-alignment
pre-commit run --files src/opentau/policies/pi06/modeling_pi06.py src/opentau/policies/pi06/configuration_pi06.py tests/policies/test_pi06.py
pytest -sx tests/policies/test_pi06.py
```

GPU smoke
```bash
pytest -m gpu -n 0 tests/policies/test_pi06.py::test_complete_pi06_pipeline_integration_smoke
```

## Checklist

### DIFF
--- a/src/opentau/policies/pi06/configuration_pi06.py
+++ b/src/opentau/policies/pi06/configuration_pi06.py
@@ -50,6 +50,15 @@ class PI06Config(PreTrainedConfig):
         max_state_dim: Maximum dimension for state vectors. Defaults to 32.
         max_action_dim: Maximum dimension for action vectors. Defaults to 32.
         predict_response: Whether to predict the response. Defaults to False.
+            Enabling this is required to reproduce the paper's hierarchical
+            design (π0.6 model card §1: "preserves the hierarchical design of
+            π0.5, providing high-level subtask prediction and low-level action
+            generation"). When False, π0.6 reduces to a flat low-level-only
+            model. The default is False because most LeRobot-style datasets do
+            not carry subtask annotations. The same field is also used to
+            supervise VQA / grounding-style textual targets during co-training
+            (this is why the field is named "response" rather than "subtask" —
+            it covers both uses, matching the π0.5 pretraining recipe).
         resize_imgs_with_padding: Target image size. Defaults to (448, 448).
         empty_cameras: Number of empty camera inputs to add. Defaults to 0.
             π0.6 pre-training uses up to 4 cameras (base + 2 wrist + optional
--- a/src/opentau/policies/pi06/modeling_pi06.py
+++ b/src/opentau/policies/pi06/modeling_pi06.py
@@ -793,8 +793,10 @@ class PI06FlowMatching(nn.Module):
         embs.append(lang_emb)
         pad_masks.append(lang_masks)
         num_lang_embs = lang_emb.shape[1]
-        # Language tokens continue the image block (bidirectional within prefix).
-        att_masks += [0] * num_lang_embs
+        # Language tokens use causal attention per the π0.6 model card §2:
+        # "use causal attention among the text tokens" — an explicit divergence
+        # from π0.5, whose language tokens shared the image block bidirectionally.
+        att_masks += [1] * num_lang_embs
 
         if response_tokens is not None:
             response_emb = self.gemma3_with_expert.embed_language_tokens(response_tokens)
--- a/src/opentau/policies/pi06/modeling_pi06.py
+++ b/src/opentau/policies/pi06/modeling_pi06.py
@@ -186,6 +186,42 @@ def resize_with_pad(img: Tensor, width: int, height: int, pad_value: int = -1) -
     return padded_img
 
 
+def flow_matching_masked_mse(
+    u_t: Tensor,
+    v_t: Tensor,
+    prefix_mask: Tensor,
+    actions_is_pad: Tensor | None,
+    max_action_dim: int,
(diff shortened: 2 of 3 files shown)


shuheng-liu:
@claude fix per review.

PR #258, opened 2026-05-04
feat(datasets): add tolerance_s and skip_timestamp_check config knobs

shuheng-liu:
## What this does

Exposes `LeRobotDataset`'s timestamp-sync tolerance through `DatasetConfig` / `DatasetMixtureConfig` and adds an explicit skip flag, so users can configure or bypass the load-time `check_timestamps_sync` call from a `TrainPipelineConfig` JSON. Previously the tolerance was hardcoded to `1e-4` with no escape hatch outside of `scripts/visualize_dataset.py`.

**New fields:**

| Field | Mixture-level default | Per-dataset override |
|---|---|---|
| `tolerance_s` | `DatasetMixtureConfig.tolerance_s: float = 1e-4` | `DatasetConfig.tolerance_s: float \| None = None` |
| `skip_timestamp_check` | `DatasetMixtureConfig.skip_timestamp_check: bool = False` | `DatasetConfig.skip_timestamp_check: bool \| None = None` |

Per-dataset values win over the mixture default when set; `None` means inherit. The skip only bypasses the load-time call in `LeRobotDataset.__init__`; the record-time call inside `add_episode` is intentionally **not** gated — bad timestamps in fresh recordings indicate a real capture bug, not tolerable drift.

Useful when one dataset in a mixture has slightly off-fps timestamps (override per-dataset to loosen for just that one) or as a debug knob to skip the check entirely while investigating timing data.

## How it was tested

- Added 4 config-level tests in `tests/configs/test_default.py`: defaults at both levels, validation rejects negative `tolerance_s` at both levels.
- Added 5 factory/dataset-level tests in `tests/datasets/test_datasets.py`: skip bypasses the load check, default runs it once, per-dataset overrides win over mixture defaults in both directions (true override, not set-when-True), mixture defaults flow through when per-dataset is `None`.
- `pre-commit run --all-files` (focused on modified files): all hooks pass.
- `pytest -m "not gpu" -n auto tests/configs/test_default.py tests/datasets/test_datasets.py`: 64 passed.
- Full CPU suite (`pytest -m "not gpu" -n auto`): 959 passed, 13 skipped, no regressions. (One pre-existing `test_libero_utils.py::test_libero2torch` failure due to missing local LIBERO benchmark data — unrelated to this PR.)

Run the focused tests:
```bash
pytest -sx tests/configs/test_default.py -k "tolerance"
pytest -sx tests/datasets/test_datasets.py -k "timestamp_check or tolerance"
```

Or exercise via a smoke-config CLI override:
```bash
opentau-train \
    --accelerate-config configs/examples/accelerate_ddp_config.yaml \
    --config_path=configs/examples/pi05_training_config.json \
    --dataset_mixture.skip_timestamp_check=true \
    --steps=2
```

Per-dataset override in JSON / CLI:
```bash
opentau-train ... \
    --dataset_mixture.datasets.0.tolerance_s=1e-3 \
    --dataset_mixture.datasets.0.skip_timestamp_check=true
```

## Checklist

### DIFF
--- a/src/opentau/configs/default.py
+++ b/src/opentau/configs/default.py
@@ -79,6 +79,20 @@ class DatasetConfig:
             metadata field. When provided (including the empty string), takes
             precedence over the value loaded from ``meta/info.json``. ``None``
             (default) leaves the loaded value untouched.
+        tolerance_s: Optional per-dataset override for the timestamp-sync
+            tolerance (in seconds) passed to ``LeRobotDataset``'s load-time
+            ``check_timestamps_sync`` call. ``None`` (default) inherits the
+            mixture-wide ``DatasetMixtureConfig.tolerance_s`` value. Set this
+            to a larger value (e.g. ``1e-3``) when a single dataset in the
+            mixture has slightly off-fps timestamps but you don't want to
+            loosen the check for the others. Must be ``>= 0`` when set.
+        skip_timestamp_check: Optional per-dataset override that bypasses the
+            load-time ``check_timestamps_sync`` call entirely. ``None``
+            (default) inherits the mixture-wide
+            ``DatasetMixtureConfig.skip_timestamp_check``. ``True`` skips the
+            check (a warning is logged); ``False`` forces the check on for this
+            dataset even if the mixture default is ``True``. Does not affect
+            the record-time check inside ``add_episode``.
 
     Raises:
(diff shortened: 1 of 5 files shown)


shuheng-liu:
@claude fix per suggestions and nits

Issue #239, opened 2026-05-04
Publish `TensorAuto/pi06-untrained` checkpoint (and pi0.7 follow-up)


shuheng-liu:
## Background

The `pi06` policy is implemented (`src/opentau/policies/pi06/`) and registered in `policies/factory.py`, and `configs/examples/pi06_training_config.json` exists, but there is no published initialization checkpoint. The README currently lists $\pi_{0.6}$ checkpoints as "Coming soon", which forces every new $\pi_{0.6}$ training run to assemble the backbone from scratch.

A separate $\pi_{0.7}$ checkpoint is tracked here as a follow-up — see the bottom of this issue.

## Ask

Publish a public, untrained $\pi_{0.6}$ initialization checkpoint to the Hub at:

```
TensorAuto/pi06-untrained
```

Composition:

- **VLM backbone**: load weights from [`google/gemma-3-4b-pt`](https://huggingface.co/google/gemma-3-4b-pt). The Gemma 3-4B PT checkpoint already bundles the SigLIP 400M vision tower, so we use it as-is to init the multimodal trunk inside `gemma3_with_expert.py`.
- **Action expert (~860M params)**: randomly initialized with the standard init used by the policy class — no pretrained weights.
- **Tokenizer / processor**: whatever ships with `google/gemma-3-4b-pt`.

Effectively this is a "from-scratch action expert on top of a frozen-init VLM" snapshot that any user can `from_pretrained("TensorAuto/pi06-untrained")` and start fine-tuning, instead of re-downloading + re-stitching the Gemma 3-4B + SigLIP 400M weights every time.

## Proposed implementation sketch (not part of this issue)

These are notes for whoever picks this up — happy to be overridden.

1. Add a small script under `src/opentau/scripts/` (e.g. `build_pi06_untrained.py`) that:
   - Instantiates `PI06Config` with the default 448×448, 5-step flow-matching settings.
   - Builds a `PI06Policy` via the factory, then loads `google/gemma-3-4b-pt` into the VLM submodule (vision tower + LM) — leaving the action expert at its default random init.
   - Calls `policy.save_pretrained(...)` and `tokenizer.save_pretrained(...)` to a local dir.
2. Push to `TensorAuto/pi06-untrained` (private → public after a sanity check).
3. Sanity-check by running `configs/examples/pi06_training_config.json` for a handful of steps starting from this checkpoint and confirming loss is finite and decreasing.
[...]

PR #263, opened 2026-05-05
test(pi07): verify pi07-{high,low}-untrained + drop expert in high

shuheng-liu:
## What this does

Closes the pi07 follow-up half of #239. Publishes two private init checkpoints to the Hub and adds the analogous regression tests:

- `TensorAuto/pi07-high-untrained` — `PI07HighLevelPlannerPolicy` initialised with `google/gemma-3-4b-pt`'s VLM weights (resampled SigLIP `position_embedding` 4096 → 1024 patches for the 448×448 input resolution). The Gemma-v1 action expert is **omitted** — see refactor below.
- `TensorAuto/pi07-low-untrained` — `PI07LowLevelPolicy` with the same recipe; the `SpaceTimeSiglipVideoEncoder` wraps six SigLIP encoder layers in place but does not change the state_dict key set, so the VLM weight transfer is straightforward.

The build was done out-of-band (no script committed, matching PR #245's pi06 approach). This PR carries the regression tests + two real OpenTau changes that surfaced while bootstrapping the checkpoints.

### Bug fix folded in: `transformers.PretrainedConfig` codec for draccus

pi07's policy configs declare `vlm_config: Gemma3WithExpertConfig` as a draccus dataclass field. `Gemma3WithExpertConfig` is a `transformers.PretrainedConfig`, not a draccus dataclass — so on `_save_pretrained`, draccus walks the field and raises `Exception("No parser for object ...")`. pi06 sidesteps this by constructing `Gemma3WithExpertConfig` programmatically inside the policy `__init__` (no dataclass field), but pi07 relies on the field for `--policy.vlm_config.*` CLI overrides. Fix: register `draccus.encode`/`draccus.decode` for `PretrainedConfig` with `include_subclasses=True` in `src/opentau/configs/policies.py`. Encoder is `obj.to_dict()`; decoder is `cls.from_dict(raw_value)` where draccus passes the declared subclass when `include_subclasses=True`. Without this, `PI07*Policy.from_pretrained` could not round-trip a freshly-built checkpoint through the standard API. Three CPU unit tests pin the round-trip in `tests/configs/test_default.py::TestPretrainedConfigCodec`.

### Refactor folded in: skip Gemma-v1 expert in high-level planner

The π0.7 high-level planner calls `Gemma3WithExpertModel.forward` with `inputs_embeds=[prefix_embs, None]` (modeling_pi07_high_level.py:1157) and discards the expert output `(prefix_out, _)`. The ~860M-parameter Gemma-v1 expert was being instantiated and serialised anyway — ~1.7 GB of dead weight on disk and in memory per high-level checkpoint. Add `disable_action_expert: bool = False` to `Gemma3WithExpertConfig`; when True, `__init__` leaves `self.gemma_expert = None` and the `forward`/`_run_layer` paths gate every `self.gemma_expert.{model.norm,model.layers}` lookup. A clear `ValueError` surfaces if a non-None expert stream is fed when the expert is disabled. `train_expert_only=True` is forbidden together with the new flag (no expert to train). `PI07HighLevelPlannerConfig.vlm_config` defaults the flag to `True`; `PI07LowLevelConfig` keeps it `False` (low-level actually consumes the expert via flow matching). Seven CPU unit tests in `TestDisableActionExpert` (`tests/policies/test_pi07_cpu.py`).

### New gpu+slow tests

Six `gpu+slow` tests in `tests/policies/test_pi07_untrained.py` (one file with shared fixtures so the ~10 GB `google/gemma-3-4b-pt` reference is downloaded only once across both checkpoints):
- `test_pi07_{high,low}_untrained_loads_with_no_missing_or_unexpected_keys` — checkpoint round-trips with no missing/unexpected keys (allowlists Gemma 3's tied embed_tokens/lm_head, which `save_model_as_safetensor` de-dupes on disk).
- `test_pi07_{high,low}_untrained_vlm_matches_gemma3_4b_pt` — Gemma 3 text tower + multimodal projector are bit-identical to `google/gemma-3-4b-pt`.
- `test_pi07_{high,low}_untrained_siglip_matches_gemma3_4b_pt` — SigLIP vision tower bit-identical, modulo the deterministic 4096 → 1024 bilinear resample of `position_embedding`. Same `vision_tower` filter works for both sub-policies because `SpaceTimeEncoderLayerWrapper` adopts the wrapped `SiglipEncoderLayer`'s submodules under their original attribute names (already pinned structurally by `tests/policies/test_pi07_video_encoder_cpu.py::test_state_dict_keys_match_vanilla_siglip`).

## How it was tested

- New CPU unit tests in `tests/configs/test_default.py::TestPretrainedConfigCodec` (3 tests) and `tests/policies/test_pi07_cpu.py::TestDisableActionExpert` (7 tests) — all passed locally.
- Existing `tests/policies/test_pi07_*` CPU tests (76 tests) pass unchanged after the refactor — every gating predicate is conditional on the new flag, so `disable_action_expert=False` (the default for low-level and every existing caller) is byte-identical to the prior behaviour.
- The six new `gpu+slow` tests in `tests/policies/test_pi07_untrained.py` were run on an internal GPU dev box against the just-published private repos — all six passed in ~28 minutes (10 GB download dominates the runtime; the comparisons themselves take seconds).
- Round-trip sanity: `PI07HighLevelPlannerPolicy(PI07HighLevelPlannerConfig())._save_pretrained(...)` followed by `PreTrainedConfig.from_pretrained(...)` returns a `PI07HighLevelPlannerConfig` instance with the `vlm_config` subtree fully reconstructed (was previously raising `Exception: No parser for object ... Gemma3WithExpertConfig`).

```bash
# CPU unit tests for the new codec + refactor (fast, no network)
pytest -sx tests/configs/test_default.py::TestPretrainedConfigCodec
pytest -sx tests/policies/test_pi07_cpu.py::TestDisableActionExpert
```

```bash
# Heavy tests: gpu+slow, downloads ~10 GB of weights (the private
# pi07-{high,low}-untrained repos and google/gemma-3-4b-pt). HF_TOKEN must
# be set and the user must have read access to the TensorAuto repos.
pytest -m "gpu and slow" -k pi07_untrained tests/policies/test_pi07_untrained.py -v
```

## Checklist

### DIFF
--- a/src/opentau/configs/policies.py
+++ b/src/opentau/configs/policies.py
@@ -33,6 +33,7 @@ import draccus
 from huggingface_hub import hf_hub_download
 from huggingface_hub.constants import CONFIG_NAME
 from huggingface_hub.errors import HfHubHTTPError
+from transformers import PretrainedConfig as _HFPretrainedConfig
 
 from opentau.configs.refs import resolve_refs
 from opentau.configs.types import FeatureType, NormalizationMode, PolicyFeature
@@ -42,6 +43,43 @@ from opentau.utils.hub import HubMixin
 
 # Generic variable that is either PreTrainedConfig or a subclass thereof
 T = TypeVar("T", bound="PreTrainedConfig")
+
+
+# --- transformers.PretrainedConfig <-> draccus codec ---
+# pi07's policy configs declare dataclass fields of type
+# `Gemma3WithExpertConfig` (a `transformers.PretrainedConfig` subclass) so the
+# VLM topology travels with the policy config and is serialised in
+# `config.json` via `_save_pretrained`. draccus's choice-type encode/decode
+# path traverses dataclass fields and looks up an encoder by field type —
+# without a registered handler for `PretrainedConfig`, encoding raises
+# `Exception("No parser for object ...")` because `PretrainedConfig` is not
+# itself a dataclass. Registered with `include_subclasses=True` so any
+# subclass dispatches here automatically; the decoder receives the declared
+# subclass as its first argument, which is what lets `cls.from_dict`
+# reconstruct the right concrete type. Lives in this module (rather than
+# `configs/default.py` next to the np.ndarray codec) so it's loaded as a
+# side-effect of any policy-config import path that hits `_save_pretrained`,
+# including standalone scripts that don't go through `configs/default.py`.
+def _decode_hf_pretrained_config(cls, raw_value, path=()):
+    """Decode a dict back into the declared ``PretrainedConfig`` subclass.
+
+    Args:
+        cls: The declared subclass of ``transformers.PretrainedConfig`` (e.g.
+            ``Gemma3WithExpertConfig``). Provided by draccus because this
+            decoder was registered with ``include_subclasses=True``.
+        raw_value: The serialised dict produced by the matching encoder
+            (``obj.to_dict()``).
+        path: Ignored; kept for compatibility with draccus's decoder
+            signature.
+
+    Returns:
+        An instance of ``cls`` reconstructed from ``raw_value``.
+    """
+    return cls.from_dict(raw_value)
+
+
+draccus.encode.register(_HFPretrainedConfig, lambda obj: obj.to_dict(), include_subclasses=True)
+draccus.decode.register(_HFPretrainedConfig, _decode_hf_pretrained_config, include_subclasses=True)
 
 _DEPRECATED_LATENCY_FIELDS = (
     "cloud_vlm_latency_mean",
--- a/src/opentau/policies/pi07/gemma3_with_expert.py
+++ b/src/opentau/policies/pi07/gemma3_with_expert.py
@@ -116,6 +116,7 @@ class Gemma3WithExpertConfig(PretrainedConfig):
         discrete_action_vocab_size: int | None = None,
         dropout: float = 0.1,
         gradient_checkpointing: bool = False,
+        disable_action_expert: bool = False,
         **kwargs,
     ):
         """Initializes the configuration.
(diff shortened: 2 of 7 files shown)


shuheng-liu:
@claude fix per review.

shuheng-liu:
**Note on file size**: it's worth flagging that the `pi07-low-untrained` checkpoint is *smaller* on disk than the older `TensorAuto/pi05_base`, even though pi07's low-level policy has more parameters than pi05's. This is purely a precision difference — the published `pi05_base` is fp32 throughout, while pi07 builds use the standard `Gemma3WithExpertModel.to_bfloat16_like_physical_intelligence()` plumbing.

Header-only metadata read (`safetensors` header parsed via HTTP range requests — no full download):

| Repo | Params | F32 | BF16 | Size |
|---|---|---|---|---|
| `TensorAuto/pi05_base` | 4.14 B | 100% | 0% | 16.57 GB |
| `TensorAuto/pi07-low-untrained` | 5.92 B | 6% | 94% | 12.55 GB |
| `TensorAuto/pi07-high-untrained` | 4.31 B | 0.2% | 99.8% | 8.64 GB |

Cast to a common dtype for an apples-to-apples comparison:
- `pi05_base` in bf16 → ~8.3 GB
- `pi07-low-untrained` in fp32 → ~23.7 GB

The 6% fp32 residual in `pi07-low-untrained` (~354M params, ~1.4 GB) is everything that `to_bfloat16_like_physical_intelligence` doesn't touch: `gemma_expert.model.norm`, `da_head`, `discrete_action_embedding`, the SpaceTime video-encoder temporal-attention additions, and the action/state projectors. Could be tightened to full-bf16 in a follow-up if desired, but not load-bearing for any test or training behavior in this PR.

Issue #264, opened 2026-05-05
pi07_low_level: max bs non-monotonic in num_cams under ZeRO-2


shuheng-liu:
### System Info

- 8×A100-80GB single node, NVIDIA driver / CUDA-12.x
- OpenTau branch: `main` @ commit 5ab5b56 (`feat(datasets): add tolerance_s and skip_timestamp_check config knobs (#258)`)
- Python 3.10, PyTorch from current `uv.lock`, `datasets==4.5.0`
- Backend: ZeRO-2 (DeepSpeed via `accelerate`), `mixed_precision: bf16`, `gradient_accumulation_steps: 1`
- Optimizer: fused AdamW (rebuilt with `FUSED_ADAMW=true` env var) + `MasterWeightOptimizer` (per-rank fp32 master + fp32 Adam state)
- Dataset: `physical-intelligence/libero` only, `n_obs_history=6`, `history_interval=1`, `skip_timestamp_check=true`
- Eager attention, no gradient checkpointing
- Policy: `pi07_low_level` (Gemma 3 4B + SpaceTime SigLIP, ~5.5B total params)

### Information

The problem arises when running `src/opentau/scripts/profile_step.py` to size max batch for ZeRO-2 across `num_cams ∈ {1, 2, 3, 4}`.

### Reproduction

I expected `max_bs` to be monotonically non-increasing in `num_cams` (more cameras → more SigLIP+expert activations → smaller fitting batch). Observed:

| num_cams | max bs (per rank) | global samples/s | step time |
|---|---|---|---|
| 1 | 6 | 12.2 | 3946 ms |
| 2 | 3 | 6.3 | ~3850 ms |
| **3** | **4** | **7.6** | ~4170 ms |
| 4 | 3 | 5.6 | ~4350 ms |

`num_cams=3 bs=4` reproducibly fits (7.6 / 7.7 samples/s across two independent runs); `num_cams=2 bs=4` reproducibly OOMs at the forward-pass attention matmul.

#### Probe command (per `(num_cams, bs)`)

```bash
FUSED_ADAMW=true PROFILE_STEPS=1 OMP_NUM_THREADS=4 \
  accelerate launch \
    --config_file=accelerate_zero2_8gpu_bf16.yaml \
    src/opentau/scripts/profile_step.py \
    --config_path=<libero_pi07_low_level.json> \
    --batch_size=$BS --dataloader_batch_size=$BS --num_cams=$NUM_CAMS
```

[...]

PR #265, opened 2026-05-05
feat(pi07): FSDP support via decoder-layer bundling refactor

shuheng-liu:
## What this does

Makes pi07 full-unfreeze training work under PyTorch FSDP-FULL_SHARD on 8×A100, with the same fp32-master / bf16-compute precision regime DeepSpeed uses — as a lower-overhead alternative to DeepSpeed ZeRO-2.

**200-step verified throughput (pi07 low-level, full unfreeze, sdpa+grad_ckpt, default stochastic drops):**

| Config | per-rank bs | step (ms) | samples/s | vs ZeRO-2 baseline |
| --- | ---: | ---: | ---: | ---: |
| ZeRO-2 baseline (no grad_ckpt, eager) | 6 | 3946 | 12.2 | 1.00× |
| **FSDP-FULL_SHARD, num_cams=1** | **12** | **6180** | **15.5** | **+1.27×** |
| **FSDP-FULL_SHARD, num_cams=4** (pre-training target) | **8** | **11038** | **5.8** | — (different cam count) |

**Precision regime — matches DeepSpeed ZeRO-1/2:**

| | DDP / single | DeepSpeed ZeRO-1/2 | **FSDP-FULL_SHARD (this PR)** |
| --- | --- | --- | --- |
| Live (compute) params | bf16 | bf16 | **bf16** (FSDP MixedPrecision downcast) |
| Master params | fp32 (`MasterWeightOptimizer`) | fp32 (`BF16_Optimizer`) | **fp32** (FSDP outer params) |
| Adam state | fp32 | fp32 | **fp32** (built over fp32 outer) |
| Gradient reduce-scatter | bf16 | bf16 | **bf16** (`reduce_dtype=bf16`) |

The fp32 master + fp32 Adam state is what protects late-training updates from bf16's 7-bit-mantissa underflow on small gradient magnitudes. Throughput cost vs the simpler bf16-everywhere regime is **<2%** (0.6% at cam=1, 1.7% at cam=4) — within step-time noise.

**The structural change:**

- **`InterleavedDecoderLayer(nn.Module)`** in `gemma3_with_expert.py` bundles one Gemma 3 backbone layer + one Gemma-v1 expert layer for one decoder step. `Gemma3WithExpertModel.__init__` reparents per-layer modules out of `gemma3.language_model.model.layers` / `gemma_expert.model.layers` into a flat `interleaved_layers` ModuleList, and empties the source ModuleLists so each Parameter is registered exactly once. The accelerate FSDP yaml wraps `InterleavedDecoderLayer` (instead of the inner `Gemma3DecoderLayer` / `GemmaDecoderLayer`) so the all-gather hook fires before the sub-component accesses inside `forward`. Without this refactor, both FSDP1 and ZeRO-3 either silently produce wrong gradients or hang at NCCL all-gather.
- **`Gemma3WithExpertConfig.disable_internal_bf16_cast`** (new) gates the in-`__init__` `to_bfloat16_like_physical_intelligence` call so the policy stays fp32 going into FSDP. train.py / profile_step.py set the flag whenever `accelerator.distributed_type == FSDP`. Other backends (DDP, ZeRO-1/2) keep the existing behavior. Without this gate, FSDP's `MixedPrecision` becomes a storage no-op (params already bf16) and AdamW silently allocates state in bf16.
- **`_zero3_disabled_init_context`** in `train.py` calls `unset_hf_deepspeed_config()` for the duration of `make_policy` so SigLIP's `lecun_normal_` (which needs the full tensor shape for fan-in/out) doesn't crash on a per-rank ZeRO-3 partition.
- **`_global_any` branch sync in `embed_prefix`** — every per-rank decision to enter an `if has_response / has_subgoal / has_metadata` branch is globally OR-reduced (1-element `MAX` all-reduce, no-op outside `torch.distributed`). Stochastic drops give different ranks different feature subsets; before the sync, ranks took different branches inside `embed_prefix`, which made FSDP issue a different number of all-gather collectives on different ranks and hung NCCL with mismatched SeqNum. The cleaner alternative — hoist the `embed_*` calls out of the `if`s entirely — is filed as a follow-up.
- **Step 0 wins** that always help, included here: `attention_implementation: "sdpa"`, `gradient_checkpointing: true`, `PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True` defaulted in the `opentau-train` launcher.
- **`dataset_mixture.py` `count`-key fix** (issue #264) — the empty-camera padding tolerated v2.0 stats that lack `count`. Unblocks `num_cams > actual_cams` benchmarks on libero.
- **CI fix** — the `InterleavedDecoderLayer` originally captured `model.get_attention_interface()` at init, which silently bypassed runtime patches to `model.eager_attention_forward` / `sdpa_attention_forward` (existing test pattern). Switched to a per-call lookup so monkey-patches still take effect.
- **`Normalize.forward` DTensor materialization** — first FSDP2 blocker. Tiny `_materialize(tensor)` helper calls `.full_tensor()` when the input is a DTensor, returns it as-is otherwise. No-op for FSDP1/DDP/single. Fast on these stats because their placement is "replicated" by default.

**Per-rank persistent memory under fp32-master (pi07, ~5.5B):**
- fp32 master params: 5.5B × 4 / 8 = ~2.75 GB
- fp32 Adam exp_avg + exp_avg_sq: ~5.5 GB
- bf16 grad buffer (post reduce-scatter): ~1.4 GB
- NCCL + cuDNN + allocator overhead: ~1–2 GB
- **~10–11 GB/rank persistent**, leaving ~70 GB on each 80 GB A100 for activations.

New files:
- `configs/examples/pi07_low_level_libero.json` — real-pi07 (Gemma 3) example (sdpa, grad_ckpt, default stochastic drops, 8 history frames).
- `configs/examples/accelerate_fsdp_config.yaml` — FSDP1 + FULL_SHARD (production).
- `configs/examples/accelerate_fsdp2_config.yaml` — FSDP2 (experimental — see commit `a20cd47` for status).
- `configs/examples/accelerate_deepspeed_zero3_config.yaml` — ZeRO-3 baseline.

## How it was tested

- **CPU pi07 suite** (`tests/policies/test_pi07_*.py`) all pass against the refactored model. The 9 `TestInterleavedDecoderLayer` tests pin: module count, source ModuleLists empty after init, every Parameter object registered exactly once, state-dict layer keys under the new `interleaved_layers.X.{backbone,expert}_layer.*` namespace, `train_expert_only=True` freeze paths, `train(mode=True)` eval propagation, attention dispatch resolved at call time (the CI regression), seeded-forward determinism, and that the cached attention-dispatch provider doesn't leak into state_dict. The 2 `TestDisableInternalBf16Cast` tests pin both directions of the new flag (default keeps the cast active; flag skips it and params stay fp32).
- **GPU pytest pi07** (`pytest -m gpu tests/policies/test_pi07_low_level.py tests/policies/test_pi07_high_level_planner.py`): 7 / 7 pass on 8×A100 in 2:11.
- **Byte-identical pre vs post-refactor**: built the pre-refactor model (main repo, OLD `_run_layer`) and the post-refactor model (this branch, `InterleavedDecoderLayer`) with the same seed and same input. Backbone and expert outputs are bit-equal (`max_abs_diff = 0`).
- **8×A100 profiling**: 200-step `profile_step.py` end-to-end at the two configurations above (cam=1 bs=12 and cam=4 bs=8). No NaNs / crashes / hangs at either.

Not yet run before merge:
- Optional: state_dict back-compat hook (existing pi07 checkpoints have `gemma3.language_model.model.layers.X.*` keys; reload onto the refactored model would miss layer params unless translated).
- pi07 high-level planner under FSDP — same `Gemma3WithExpertModel` so structurally compatible, but throughput benchmarking requires a libero subtask-annotated dataset.
- pi06 under FSDP — has its own copy of `gemma3_with_expert.py` without the refactor; FSDP fails at `accelerate.prepare` because the wrap target `InterleavedDecoderLayer` doesn't exist in pi06's namespace. Same architectural fix needs porting.
- FSDP2 end-to-end — `Normalize.forward` fix landed; `SiglipVisionEmbeddings` wrap helps; the third blocker (`video_encoder.py:346` direct sub-component access on the SpaceTime SigLIP) still requires the equivalent `nn.Module` bundling that `InterleavedDecoderLayer` does for the decoder. Documented inline in `accelerate_fsdp2_config.yaml`.

Pre-training target (cam=4 bs=8):

```bash
opentau-train \
    --accelerate-config configs/examples/accelerate_fsdp_config.yaml \
    --config_path configs/examples/pi07_low_level_libero.json \
    --batch_size 8 --dataloader_batch_size 8 --num_cams 4 --num_workers 8
```

Reproducing the cam=1 throughput headline:

```bash
PROFILE_STEPS=200 PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True \
    accelerate launch \
    --config_file configs/examples/accelerate_fsdp_config.yaml --num_processes=8 \
    --config_path configs/examples/pi07_low_level_libero.json \
    --batch_size=12 --dataloader_batch_size=12 --num_cams=1 \
    --policy.freeze_vision_encoder=false --policy.train_expert_only=false \
    --policy.gradient_checkpointing=true --policy.attention_implementation=sdpa \
    --wandb.enable=false
```

## Checklist

### DIFF
--- a/src/opentau/policies/pi07/low_level/modeling_pi07_low_level.py
+++ b/src/opentau/policies/pi07/low_level/modeling_pi07_low_level.py
@@ -1305,6 +1305,16 @@ class PI07LowLevelFlowMatching(nn.Module):
         # collectives. Ranks whose local data is all-dropped still embed real
         # tokens with all-False masks; downstream attention / loss respect the
         # mask, so accuracy is preserved.
+        #
+        # Scope of the sync: this only protects against per-rank divergence of
+        # the ``.any()`` outcome on a field that *is* present everywhere. The
+        # inner branch (``if has_response and response_tokens is not None
+        # and response_masks is not None``) still gates on per-rank Nones,
+        # which assumes field presence is a global property of the dataset
+        # config (true today — every rank instantiates the same ``embed_prefix``
+        # contract). If a future heterogeneous-mixture change makes some ranks
+        # carry e.g. ``response_tokens=None`` while others carry a Tensor, the
+        # branches would re-desync and need their own None-vs-Tensor sync.
         device = lang_tokens.device
         has_response_local = (
             response_tokens is not None and response_masks is not None and bool(response_masks.any())
--- a/src/opentau/policies/pi07/low_level/modeling_pi07_low_level.py
+++ b/src/opentau/policies/pi07/low_level/modeling_pi07_low_level.py
@@ -68,6 +68,35 @@ def _preferred_dtype():
     return torch.float32 if torch.onnx.is_in_onnx_export() else torch.bfloat16
 
 
+def _global_any(local: bool, device: torch.device) -> bool:
+    """Return ``local`` OR-reduced across all ranks of the current process group.
+
+    pi07's ``embed_prefix`` chooses Python-level branches based on whether the
+    local rank's micro-batch contains any subgoal images / response tokens /
+    metadata tokens. Each branch fires a different number of FSDP / ZeRO-3
+    all-gather collectives (one per ``embed_language_tokens`` call), so under
+    stochastic dataset drop probabilities — where one rank's batch may roll
+    "all dropped" while another rolls "some real" — the ranks would otherwise
+    take different branches and deadlock at NCCL with mismatched all-gather
+    sequence numbers.
+
+    This helper turns a per-rank ``bool`` into a globally-OR'd ``bool`` via a
+    single 1-element ``MAX`` all-reduce. We use it once per branch decision
+    in ``embed_prefix`` so every rank executes the same code path even when
+    its local data slice has different drop rolls.
+
+    The all-reduce cost is negligible (~tens of microseconds per call,
+    handful of calls per step). Outside ``torch.distributed`` (single-process
+    runs, CPU smoke tests) the helper is a fast no-op that just returns the
+    local value.
+    """
+    if not (torch.distributed.is_available() and torch.distributed.is_initialized()):
+        return bool(local)
+    flag = torch.tensor([1 if local else 0], device=device, dtype=torch.int32)
+    torch.distributed.all_reduce(flag, op=torch.distributed.ReduceOp.MAX)
+    return bool(flag.item())
+
+
 def create_sinusoidal_pos_embedding(
     time: Tensor, dimension: int, min_period: float, max_period: float, device: torch.device | str = "cpu"
 ) -> Tensor:
@@ -1264,15 +1293,31 @@ class PI07LowLevelFlowMatching(nn.Module):
         # state-end separator collapses to ":\n" and the trailing prefix-end is
         # omitted, eliminating spurious dangling tokens that would otherwise break
         # the cumsum at the indicator -> first-discrete boundary.
-        has_response = (
+        #
+        # Each ``has_*`` flag is OR-reduced across all ranks via ``_global_any``.
+        # Different ranks process different micro-batches, so under stochastic
+        # ``*_drop_prob`` settings their local ``.any()`` values can diverge.
+        # If we let each rank take its local branch, FSDP / ZeRO-3 would launch
+        # different numbers of param all-gathers per rank and deadlock at NCCL
+        # (observed empirically: rank 0,1,3,4,5,7 stuck at SeqNum=274 ALLREDUCE,
+        # rank 2,6 at SeqNum=279 ALLGATHER_BASE). Synchronizing the *branch
+        # decision* — not the data — keeps every rank running the same set of
+        # collectives. Ranks whose local data is all-dropped still embed real
+        # tokens with all-False masks; downstream attention / loss respect the
+        # mask, so accuracy is preserved.
+        device = lang_tokens.device
+        has_response_local = (
             response_tokens is not None and response_masks is not None and bool(response_masks.any())
         )
-        has_subgoal = (
+        has_subgoal_local = (
             bool(subgoal_images) and bool(subgoal_img_masks) and any(bool(m.any()) for m in subgoal_img_masks)
         )
-        has_metadata = (
+        has_metadata_local = (
             metadata_tokens is not None and metadata_masks is not None and bool(metadata_masks.any())
         )
+        has_response = _global_any(has_response_local, device)
+        has_subgoal = _global_any(has_subgoal_local, device)
+        has_metadata = _global_any(has_metadata_local, device)
         has_any_optional = bool(has_response or has_subgoal or has_metadata)
 
         for vid, vid_mask in zip(videos, vid_masks, strict=True):
@@ -1369,11 +1414,12 @@ class PI07LowLevelFlowMatching(nn.Module):
         # state-end separator (", " or ":\n") is text → causal per paper §VI.B.
         att_masks += [1] * num_state_end_embs
 
-        # Only embed response when at least one sample in the batch has real response tokens.
-        # With response_drop_prob=1.0 all masks are False; skipping avoids a spurious
-        # causal-block boundary (att_masks=[1,...]) that would corrupt cumsum for every
-        # subsequent token.
-        if response_tokens is not None and response_masks is not None and response_masks.any():
+        # Only embed response when at least one sample in *any rank's* batch has
+        # real response tokens. The decision is taken on the globally-synced
+        # ``has_response`` flag (computed above) — using a local ``.any()``
+        # check here would let different ranks diverge into different code
+        # paths and deadlock under FSDP / ZeRO-3.
+        if has_response and response_tokens is not None and response_masks is not None:
             response_emb = self.gemma3_with_expert.embed_language_tokens(response_tokens)
             embs.append(response_emb)
             pad_masks.append(response_masks)
@@ -1383,8 +1429,8 @@ class PI07LowLevelFlowMatching(nn.Module):
             num_response_embs = response_emb.shape[1]
             att_masks += [1] * num_response_embs
 
-        # Only embed metadata when at least one sample has real metadata tokens.
-        if metadata_tokens is not None and metadata_masks is not None and metadata_masks.any():
+        # Same global-synced gate for metadata.
+        if has_metadata and metadata_tokens is not None and metadata_masks is not None:
             metadata_emb = self.gemma3_with_expert.embed_language_tokens(metadata_tokens)
             embs.append(metadata_emb)
             pad_masks.append(metadata_masks)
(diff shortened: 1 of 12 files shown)


shuheng-liu:
Follow-up tracked in [#266](https://github.com/TensorAuto/OpenTau/issues/266) — the `_global_any` fix in this PR works but adds three small all-reduces per `embed_prefix` and makes ranks carry pad slots through every layer when only some ranks have content for an optional field. The cleaner pattern is to hoist the `embed_*` calls out of the `if has_*` branches (uniform call count → no sync needed) and leave only the bookkeeping inside. Filed as a follow-up so PR #265 stays focused on the FSDP-correctness fix.

shuheng-liu:
## Honest headline update — stochastic drops

The earlier headline ("19.0 samples/s") was measured with `drop_prob=1.0` on every `*_drop_prob` in the example config, which I had set as a workaround for a NCCL desync. That workaround skipped real `embed_language_tokens`/`embed_image` work for the response/metadata/subgoal blocks every step — so the number was inflated, not representative.

Commit `1f27b98` (`_global_any` global-OR sync of the data-dependent branch decisions in `embed_prefix`) is the actual fix for the desync. With that landed, stochastic drops work correctly under FSDP. The example config has been reverted to honest defaults (history=0.3, subgoal=0.75, response=0.3, etc.).

**Re-measured headline (200 steps, FSDP-FULL_SHARD, bs=12 num_cams=1, full unfreeze, sdpa+grad_ckpt, default stochastic drops):**

| backend (drops) | bs/rank | step (ms) | samples/s |
| --- | ---: | ---: | ---: |
| ZeRO-2 baseline (no grad_ckpt, eager, default drops) | 6 | 3946 | 12.2 |
| FSDP + sdpa + grad_ckpt, default drops | 12 | 6168 | **15.6** |
| FSDP + sdpa + grad_ckpt, all drops=1.0 (workaround, withdrawn) | 12 | 5057 | 19.0 |

**Apples-to-apples win: +1.28× throughput, +2× batch.** Smaller than the previously claimed 1.56×, but it's the honest number.

Filed [#266](https://github.com/TensorAuto/OpenTau/issues/266) as a follow-up: the `_global_any` fix could be replaced by hoisting the `embed_*` calls out of the `if has_*` branches entirely, dropping the all-reduces and letting ranks honestly carry per-rank prefix lengths.

shuheng-liu:
## Camera sweep — honest stochastic-drops re-measurement

Re-ran the camera sweep at the previously-found sweet spots under realistic stochastic drops (the same default `*_drop_prob` settings the user runs in production). Notable: cam=2's sweet spot **shifted from bs=12 down to bs=8** under stochastic drops because bs=12 hit allocator pressure (bwd ballooned ~3×). cam=3 and cam=4 throughputs barely changed — the per-cam vision compute already dominated.

| num_cams | best bs | step (ms) | samples/s | vs ZeRO-2 baseline (12.2 sps) |
| ---: | ---: | ---: | ---: | ---: |
| 1 | 12 | 6168 | **15.6** | +1.28× |
| 2 | 8  | 7437 | 8.6  | n/a (different bs) |
| 3 | 8  | 8408 | 7.6  | n/a |
| 4 | 8  | 10820 | 5.9  | n/a |

For comparison, deterministic-drop (drop=1.0 workaround, no longer recommended) numbers were 19.0 / 11.3 / 7.7 / 6.0 sps respectively — the 1-camera and 2-camera configurations are noticeably slower honestly because more of the response/metadata/subgoal work actually fires.

These are the numbers a reviewer should believe. The plan file appendix has the full breakdown.

src/opentau/scripts/train.py
```diff
@@ -326,18 +375,38 @@ def train(cfg: TrainPipelineConfig):
...
-    policy.to(torch.bfloat16)
+    # ZeRO does this through ``BF16_Optimizer``; under DDP/single we mirror
+    # that with ``MasterWeightOptimizer`` (see issue #181). Under FSDP,
+    # the policy must stay in fp32 here so FSDP's ``MixedPrecision`` (set
+    # by accelerate from ``mixed_precision: bf16`` in the FSDP yaml) can
+    # downcast on the fly while keeping the fp32 outer params for the
+    # optimizer to step — same fp32-master / bf16-compute split, but owned
+    # by FSDP rather than our wrapper.
+    if accelerator.distributed_type != accelerate.DistributedType.FSDP:
+        policy.to(torch.bfloat16)
```
shuheng-liu, line 419: Fixed in a4dea9b — agreed, the fp32-master claim was inaccurate. Updated the comments to honestly describe the FSDP regime as **pure bf16** (no fp32 master): `Gemma3WithExpertModel.__init__` calls `to_bfloat16_like_physical_intelligence` unconditionally, params are bf16 before FSDP wraps, MixedPrecision becomes a storage no-op, and the optimizer steps on bf16 sharded params. The same fix landed in `profile_step.py`. Documented that to actually get fp32-master under FSDP we'd need to gate the inner cast — out of scope here, would change the throughput numbers.

src/opentau/policies/pi07/low_level/modeling_pi07_low_level.py
```diff
@@ -1264,15 +1293,31 @@ def embed_prefix(
...
+        has_subgoal_local = (
             bool(subgoal_images) and bool(subgoal_img_masks) and any(bool(m.any()) for m in subgoal_img_masks)
         )
-        has_metadata = (
+        has_metadata_local = (
             metadata_tokens is not None and metadata_masks is not None and bool(metadata_masks.any())
         )
+        has_response = _global_any(has_response_local, device)
+        has_subgoal = _global_any(has_subgoal_local, device)
+        has_metadata = _global_any(has_metadata_local, device)
```
shuheng-liu, line 1330: Fixed in a4dea9b — agreed, added a paragraph clarifying that the sync only protects against per-rank `.any()` divergence on a field that is *uniformly present* across ranks (true today because field presence is a global property of the dataset config). Called out explicitly that a future heterogeneous-mixture change with per-rank None-vs-Tensor variation would re-desync and need its own None-vs-Tensor sync.

shuheng-liu:
Closing this. We surfaced throughput and memory observations that don't reconcile with the underlying memory / communication math, and we don't have enough confidence in the numbers to merge:

- At num_cams=4 bs=8, ZeRO-2 ran ~60% faster per step than FSDP-FULL_SHARD (9.3 vs 5.8 samples/s on a 10-step probe), with the gap concentrated in backward (ZeRO-2 4.3s vs FSDP 7.8s). Naive math predicts ~5–10% delta from FSDP's extra all-gather, not 60%.
- We never instrumented peak HBM use. ZeRO-2 has +11 GB persistent per rank (bf16 params replicated) vs FSDP-FULL_SHARD, so the math says ZeRO-2's batch ceiling should be lower — yet at cam=4 bs=8 both fit. No way to tell from the recorded data which one is wrong.
- The "FSDP +1.27× over ZeRO-2 baseline" headline number was unfair: the FSDP run used `sdpa` + `gradient_checkpointing`, while the ZeRO-2 baseline used `eager` + no grad_ckpt. A clean head-to-head requires the same optimizations on both.

Restarting with a structured 2×2×2 matrix (FSDP vs ZeRO-2 × `sdpa+grad_ckpt` toggle × num_cams 1/4) at maxed-out batch per cell, with peak HBM monitored throughout. The next investigation will also audit `profile_step.py` for timer / phase-accounting bugs first.

The architectural pieces on this branch — the `InterleavedDecoderLayer` refactor in `pi07/gemma3_with_expert.py`, the `_global_any` `embed_prefix` sync, the `disable_internal_bf16_cast` flag — are still required for FSDP to work at all on pi07, so the next attempt will likely build on those rather than redo them from scratch.

PR #278, opened 2026-05-06
chore(datasets): collapse skip-timestamp-check warning to once-per-process

shuheng-liu:
## What this does

The ``Skipping timestamp sync check for ...`` warning at
``src/opentau/datasets/lerobot_dataset.py:1348`` previously fired
``num_processes × num_datasets`` times — for a heterogeneous
pretraining mixture (~392 datasets) on 8 GPUs that's ~3,136 identical
lines between accelerate init and the first training step, drowning
the startup log.

``skip_timestamp_check`` is fundamentally a **mixture-wide decision**
(with optional per-dataset override), and the warning text is identical
for every dataset that has it set. So the right collapse is **once per
process** — combined with the rank-0 gate that's *one line per run*,
not per dataset.

Implementation mirrors the existing ``_CONTROL_MODE_WARNED`` dedup
pattern in the same file:

- New module-level ``_SKIP_TIMESTAMP_WARNED: bool`` flag.
- Inside ``LeRobotDataset.__init__``, the warning is gated by both
  rank (``acc.is_main_process``) and the flag — fire once, then flip.
- Falls through to ``logging.warning`` when no Accelerator is set so
  single-process dev / pytest runs are unchanged.
- Message now mentions the first dataset hit as an *example* rather
  than the only one, since the flag is mixture-wide.

## How it was tested

- ``pre-commit run --files src/opentau/datasets/lerobot_dataset.py`` — clean.
- ``pytest -m "not gpu" -n auto tests/datasets/`` — 375 passed, 7
  skipped, no regressions.

Behavior comparison:

| Setup | Lines emitted (before) | Lines emitted (after) |
| --- | ---: | ---: |
| 8 ranks × 392 datasets | ~3,136 | **1** |
| 1 rank (CPU dev) × 5 datasets | 5 | **1** |
| 1 rank × 1 dataset | 1 | 1 |

The opt-in semantic and safety implication of ``skip_timestamp_check``
itself are unchanged — only the per-rank × per-dataset duplication
is suppressed.

```bash
git fetch origin claude/skip-timestamp-warning-rank0
git checkout claude/skip-timestamp-warning-rank0
pytest -m "not gpu" -n auto tests/datasets/
```

## Checklist

### DIFF
--- a/src/opentau/datasets/lerobot_dataset.py
+++ b/src/opentau/datasets/lerobot_dataset.py
@@ -1346,12 +1346,20 @@ class LeRobotDataset(BaseDataset):
         # Check timestamps
         # If transform is set, with_transform will decode all columns of a row before returning the desired column(s).
         if self.skip_timestamp_check:
-            logging.warning(
-                "Skipping timestamp sync check for %s (skip_timestamp_check=True). "
-                "Frame-to-frame spacing is NOT verified against 1/fps; downstream "
-                "delta_timestamps lookups may sample unintended frames.",
-                self.repo_id,
-            )
+            # LeRobotDataset is constructed once per rank, so a naive
+            # ``logging.warning`` floods the run log with ``num_processes`` ×
+            # ``num_datasets`` copies of the same message (392 × 8 ≈ 3K lines
+            # for a wide pretraining mixture). Gate to rank 0 — fall through
+            # to logging when no Accelerator is set (single-process dev /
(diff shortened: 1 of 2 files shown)


shuheng-liu:
@claude fix the missing test(s).

shuheng-liu:
@claude fix the nit

WilliamYue37 (approved):

PR #288, opened 2026-05-08
fix(datasets): NaN-tolerant aggregate_feature_stats + load-time WARN

shuheng-liu:
## What this does

Fixes a NaN propagation bug in `aggregate_feature_stats` where a single dataset whose action `min`/`max` stats contained NaN at any dim poisoned the global mixture buffer. `np.min` / `np.max` / weighted-mean all propagate NaN, so the aggregated buffer's NaN dims turned every sample's normalized `discrete_actions` into NaN regardless of source repo, blowing up the FAST tokenizer at the first batch (`OverflowError: Python int too large to convert to C int` from `chr()` on garbage cast from NaN).

- `aggregate_feature_stats` now masks per-dim NaN contributors via `np.nanmin`/`np.nanmax` for min/max, plus a NaN-aware weighted mean/variance that zeroes a contributor's weight where its mean (or variance) is NaN. Clean dims aggregate exactly as before; a dim that is NaN across every contributor stays NaN so the downstream loader (or `Normalize` buffer) can still surface it.
- `DatasetMixtureMetadata._to_standard_data_format` now logs a `WARNING` with the offending `repo_id`, key, stat name, and bad flat-dim indices, so a contaminator can be identified from training logs without a separate diagnostic patch.

Label: bug.

## How it was tested

- Added two regressions in `tests/datasets/test_compute_stats.py`:
  - `test_aggregate_feature_stats_nan_tolerant_per_dim` — a contributor with NaN at one dim must not affect aggregation at clean dims for other contributors.
  - `test_aggregate_feature_stats_all_nan_dim_stays_nan` — if every contributor is NaN at a dim, the aggregated result stays NaN there (downstream is responsible for surfacing it).
- `pytest -m "not gpu" -n auto tests/datasets/test_compute_stats.py` → 17/17 pass.
- `pytest -m "not gpu" -n auto tests/datasets/` → 378 passed, 7 skipped (gpu-only), 0 failures.
- The original `test_aggregate_feature_stats` covers the no-NaN happy path and continues to pass — clean inputs aggregate bit-identically to the prior implementation.
- Root cause was independently reproduced and confirmed via a temporary `[NaN-trace]` diagnostic patch (already reverted) which showed 14 of 32 padded action dims had NaN in `normalize.actions.{min,max}` after `_to_standard_data_format` + `aggregate_stats`, while pre-normalize raw `batch['actions']` was finite — proving the contamination came from the aggregator buffer, not runtime parquet data.

End-to-end verification on real GPU infrastructure is left to a follow-up dispatch by the reviewer; the unit tests cover the aggregator path that was broken.

```bash
git checkout claude/sad-sanderson-957544
pytest -m "not gpu" -n auto tests/datasets/test_compute_stats.py
```

To exercise the load-time WARN, point any pi07/pi05/pi06 training config at a dataset whose action stats contain NaN — the warning lands at `DatasetMixtureMetadata` construction time, naming the offending `repo_id`, key, stat name, and bad flat-dim indices.

## Checklist

### DIFF
--- a/src/opentau/datasets/compute_stats.py
+++ b/src/opentau/datasets/compute_stats.py
@@ -71,6 +71,7 @@ Example:
         >>> aggregated = aggregate_stats(stats_list, weights=weights)
 """
 
+import warnings
 from typing import Optional
 
 import numpy as np
@@ -286,6 +287,12 @@ def aggregate_feature_stats(
     Computes weighted mean and variance using the parallel algorithm for variance
     computation. Min and max are taken as the global min/max across all stats.
 
+    NaN-tolerant per-dim: a contributor whose ``mean``/``std``/``min``/``max`` is
+    NaN at some dim is excluded from the aggregation at that dim only. Clean dims
+    aggregate normally. A dim that is NaN across every contributor stays NaN, so
+    the downstream loader (or ``Normalize`` buffer) can surface it -- but a single
+    bad dataset no longer poisons the global buffer for every other sample.
+
     Args:
         stats_ft_list: List of statistics dictionaries for the same feature.
         weights: Optional weights for each statistics entry. If None, uses
@@ -296,32 +303,44 @@ def aggregate_feature_stats(
     """
     means = np.stack([s["mean"] for s in stats_ft_list])
     variances = np.stack([s["std"] ** 2 for s in stats_ft_list])
+    mins = np.stack([s["min"] for s in stats_ft_list])
+    maxs = np.stack([s["max"] for s in stats_ft_list])
 
-    # if weights are provided, use them to compute the weighted mean and variance
-    # otherwise, use episode counts as weights
-    if weights is not None:
-        counts = np.stack(weights)
-        total_count = counts.sum(axis=0)
-    else:
-        counts = np.stack([s["count"] for s in stats_ft_list])
-        total_count = counts.sum(axis=0)
+    counts = np.stack(weights) if weights is not None else np.stack([s["count"] for s in stats_ft_list])
+    total_count = counts.sum(axis=0)
 
-    # Prepare weighted mean by matching number of dimensions
     while counts.ndim < means.ndim:
         counts = np.expand_dims(counts, axis=-1)
 
-    # Compute the weighted mean
-    weighted_means = means * counts
-    total_mean = weighted_means.sum(axis=0) / total_count
+    mean_weights = np.where(np.isnan(means), 0.0, counts).astype(np.float64)
+    mean_weight_sum = mean_weights.sum(axis=0)
+    safe_means = np.where(np.isnan(means), 0.0, means)
+    total_mean = np.divide(
+        (safe_means * mean_weights).sum(axis=0),
+        mean_weight_sum,
+        out=np.full(mean_weight_sum.shape, np.nan, dtype=np.float64),
+        where=mean_weight_sum > 0,
+    )
 
-    # Compute the variance using the parallel algorithm
-    delta_means = means - total_mean
-    weighted_variances = (variances + delta_means**2) * counts
-    total_variance = weighted_variances.sum(axis=0) / total_count
+    var_weights = np.where(np.isnan(variances) | np.isnan(means), 0.0, counts).astype(np.float64)
+    var_weight_sum = var_weights.sum(axis=0)
+    safe_variances = np.where(np.isnan(variances), 0.0, variances)
+    delta_means = np.where(np.isnan(means), 0.0, means - total_mean)
+    total_variance = np.divide(
+        ((safe_variances + delta_means**2) * var_weights).sum(axis=0),
+        var_weight_sum,
+        out=np.full(var_weight_sum.shape, np.nan, dtype=np.float64),
+        where=var_weight_sum > 0,
+    )
+
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore", category=RuntimeWarning)
+        agg_min = np.nanmin(mins, axis=0)
+        agg_max = np.nanmax(maxs, axis=0)
 
     return {
-        "min": np.min(np.stack([s["min"] for s in stats_ft_list]), axis=0),
-        "max": np.max(np.stack([s["max"] for s in stats_ft_list]), axis=0),
+        "min": agg_min,
+        "max": agg_max,
         "mean": total_mean,
         "std": np.sqrt(total_variance),
         "count": total_count,
--- a/src/opentau/datasets/compute_stats.py
+++ b/src/opentau/datasets/compute_stats.py
@@ -287,11 +287,19 @@ def aggregate_feature_stats(
     Computes weighted mean and variance using the parallel algorithm for variance
     computation. Min and max are taken as the global min/max across all stats.
 
-    NaN-tolerant per-dim: a contributor whose ``mean``/``std``/``min``/``max`` is
-    NaN at some dim is excluded from the aggregation at that dim only. Clean dims
-    aggregate normally. A dim that is NaN across every contributor stays NaN, so
-    the downstream loader (or ``Normalize`` buffer) can surface it -- but a single
-    bad dataset no longer poisons the global buffer for every other sample.
+    Non-finite-tolerant per-dim: a contributor whose ``mean``/``std``/``min``/``max``
+    is non-finite (NaN or +/-Inf) at some dim is excluded from the aggregation at
+    that dim only. Clean dims aggregate normally. A dim that is non-finite across
+    every contributor stays NaN, so the downstream loader (or ``Normalize`` buffer)
+    can surface it -- but a single bad dataset no longer poisons the global buffer
+    for every other sample.
+
+    The mean and variance use *different* per-dim masks: a contributor is dropped
+    from the mean at a dim iff its ``mean`` is non-finite there, but dropped from
+    the variance at a dim iff *either* its ``mean`` or its ``std`` is non-finite
+    there (you can't form a Chan-style variance contribution without a finite
+    ``std``). ``count`` is unconditional: the sum of contributors' counts, not a
+    per-dim effective contributor count.
 
     Args:
         stats_ft_list: List of statistics dictionaries for the same feature.
@@ -312,9 +320,12 @@ def aggregate_feature_stats(
     while counts.ndim < means.ndim:
         counts = np.expand_dims(counts, axis=-1)
 
-    mean_weights = np.where(np.isnan(means), 0.0, counts).astype(np.float64)
+    means_bad = ~np.isfinite(means)
+    variances_bad = ~np.isfinite(variances)
+
+    mean_weights = np.where(means_bad, 0.0, counts).astype(np.float64)
     mean_weight_sum = mean_weights.sum(axis=0)
-    safe_means = np.where(np.isnan(means), 0.0, means)
+    safe_means = np.where(means_bad, 0.0, means)
     total_mean = np.divide(
         (safe_means * mean_weights).sum(axis=0),
         mean_weight_sum,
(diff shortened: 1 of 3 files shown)


src/opentau/datasets/dataset_mixture.py
```diff
@@ -201,6 +201,28 @@ def _to_standard_data_format(
...
+        # Surface non-finite stats with the offending repo + dim, so that the
+        # NaN-tolerant aggregator in `aggregate_feature_stats` doesn't silently
+        # drop a dataset's contribution. Cameras get min=0/max=1/mean=0/std=0
+        # placeholders for missing slots, so we only check non-camera features.
+        for data in standard_stats:
+            if data.startswith("camera"):
+                continue
+            for stat_name in ("mean", "std", "min", "max"):
+                arr = np.asarray(standard_stats[data][stat_name])
+                bad = np.flatnonzero(~np.isfinite(arr.ravel()))
```
shuheng-liu, line 214: **Predicate mismatch with the aggregator.** This warning catches both NaN and ±Inf (`~np.isfinite`), and the message claims the dim is "excluded from aggregated stats" — but `aggregate_feature_stats` only masks NaN (via `np.isnan`). A contributor with `+Inf` in its stats would:

- Trigger this WARN with text saying it was excluded, but
- Still poison `max` (since `np.nanmax([3.0, +inf]) == +inf`), and
- Still poison `mean`/`std` (since `np.where(np.isnan(means), 0.0, means)` leaves `+inf` unchanged, so the weighted mean evaluates to `+inf`).

Two ways to make the WARN truthful:

1. **Tighten the warning to NaN-only** to match the aggregator: change to `np.isnan(arr.ravel())` and reword the message to "NaN values". Lower risk; ±Inf in stat files would be a separate upstream bug to chase if/when it shows up.
2. **Broaden the aggregator** to mask non-finite (NaN + Inf): switch the `np.isnan(means)` / `np.isnan(variances)` predicates to `~np.isfinite(...)`, and replace `np.nanmin`/`np.nanmax` with a finite-only filter. More defensive but a behavior change beyond what the PR title scopes.

Option 1 is the smaller diff and keeps this PR focused; option 2 is worth a follow-up if ±Inf actually shows up in the wild.

---

src/opentau/datasets/compute_stats.py
```diff
@@ -296,32 +303,44 @@ def aggregate_feature_stats(
...
+    mean_weight_sum = mean_weights.sum(axis=0)
+    safe_means = np.where(np.isnan(means), 0.0, means)
+    total_mean = np.divide(
+        (safe_means * mean_weights).sum(axis=0),
+        mean_weight_sum,
+        out=np.full(mean_weight_sum.shape, np.nan, dtype=np.float64),
+        where=mean_weight_sum > 0,
+    )
+
+    var_weights = np.where(np.isnan(variances) | np.isnan(means), 0.0, counts).astype(np.float64)
```
shuheng-liu: **Asymmetric NaN handling between mean and variance — intentional but worth a one-line note.** `mean_weights` masks only `np.isnan(means)`, while `var_weights` masks `np.isnan(variances) | np.isnan(means)`. So a contributor with a finite mean but NaN std contributes to the aggregated mean at that dim, but is excluded from the variance at that dim.

That's the right call — you can't compute a variance contribution without `std` — but a future reader skimming the two `np.where(...)` blocks will trip over the inconsistency. Suggest adding one line to the docstring (around the "NaN-tolerant per-dim" paragraph) noting that a NaN in any of `{mean, std}` at a dim drops that contributor from the variance at that dim, while only a NaN in `mean` drops it from the mean.

Also: `total_count` is summed over all contributors unconditionally (`counts.sum(axis=0)`), which means a contributor that is NaN at every dim still contributes to the output `count`. That matches "count is a scalar sample-count, not per-dim", which is reasonable — but worth being explicit about in the docstring so downstream consumers don't read `count` as an effective per-dim contributor count.

---

src/opentau/datasets/dataset_mixture.py
```diff
@@ -201,6 +201,28 @@ def _to_standard_data_format(
...
+                if bad.size > 0:
+                    logging.warning(
+                        "Dataset %r: non-finite values in %r.%r at flat indices %s "
+                        "(shape=%s); excluded from aggregated %r stats at those dims.",
+                        repo_id,
+                        data,
+                        stat_name,
+                        bad.tolist(),
+                        tuple(arr.shape),
+                        data,
```
shuheng-liu: Nit: `data` is interpolated twice (`%r.%r ... aggregated %r stats`), reading like "non-finite values in `'actions'.'mean'` ... excluded from aggregated `'actions'` stats at those dims." The trailing `'actions'` is redundant given it's already named at the start; could be tightened to e.g.:

```python
"Dataset %r: non-finite values in %r.%r at flat indices %s (shape=%s); "
"these dims are excluded from aggregation.",
repo_id, data, stat_name, bad.tolist(), tuple(arr.shape),
```

Saves a positional arg and removes the duplicated `data` interpolation. Not a blocker.

---

shuheng-liu (commented):
## Review summary

Solid, well-scoped fix. The aggregator change is mathematically correct, the per-dim masking preserves clean dims from other contributors so a single bad dataset can no longer poison the global mixture, and the regression tests cover both the partial-NaN and all-NaN-dim cases. The load-time WARN gives the on-call a concrete breadcrumb (offending `repo_id` + stat name + flat dim indices) instead of needing a one-off diagnostic patch — that's a real operational win.

CI on this PR: CPU Tests green, Pre-commit green, check-checklist green. (The `review` check failure is the Claude auto-review bot job, not a test failure.)

### Findings

1. **Mismatch between the WARN predicate and what the aggregator actually masks** — inline on `dataset_mixture.py`. The WARN uses `~np.isfinite` (NaN + Inf) but the aggregator only handles NaN. A dataset with `+Inf` in stats would get logged as "excluded" while still poisoning `max`/`mean`/`std` at that dim. Either tighten the WARN to `np.isnan` or extend the aggregator — option 1 keeps this PR scoped, option 2 is more defensive.

2. **Asymmetric NaN masking between mean and variance** — inline on `compute_stats.py`. Intentional (you can't compute a variance contribution without `std`), but the two `np.where(...)` blocks have different predicates and that asymmetry isn't called out anywhere. Suggest one extra line in the docstring. Same paragraph could clarify that `total_count` is summed unconditionally across contributors (i.e. it's a scalar sample count, not a per-dim effective contributor count).

3. **Defensive fix, not root-cause fix** — the PR description acknowledges this explicitly, and that's fine: the band-aid stops the bleeding (no more poisoned global buffer), and the WARN names the offender for the follow-up. Worth filing the upstream "why does dataset X have NaN in `actions.{min,max}`?" issue once the WARN identifies it on real infra, so this doesn't drift.

### Nits

- `dataset_mixture.py:215-224` — `data` is interpolated twice in the format string; the trailing `aggregated %r stats` repeats info already in the preamble. Minor.
- The original variance comment ("parallel algorithm for variance") was deleted in the rewrite of `aggregate_feature_stats`. Not load-bearing, but the algorithmic context — that the per-step parallel/Chan-style variance update is what's being preserved here — is no longer visible in the function body. A one-line `# Chan-style parallel weighted variance, NaN-masked per dim` would put it back cheaply.

### Tests

`test_aggregate_feature_stats_nan_tolerant_per_dim` and `test_aggregate_feature_stats_all_nan_dim_stays_nan` are the right two cases, the existing `test_aggregate_feature_stats` confirms the clean-input path is bit-identical, and `pytest -m "not gpu" -n auto tests/datasets/` passed at 378/378. No suggested additions.

Overall: ship this once the WARN/aggregator predicate mismatch (finding 1) is resolved one way or the other. Findings 2 and the nits are docstring / message polish and can land either in this PR or a follow-up.

---

src/opentau/datasets/dataset_mixture.py
shuheng-liu: Went with option 2 (broaden aggregator) in 689546f — the WARN predicate (`~np.isfinite`) was the right one for the operational signal (also catches +/-Inf, which is just as likely to come out of a buggy per-episode stat aggregation as NaN), so the aggregator now matches it instead of the WARN narrowing.

Concretely:
- `means_bad = ~np.isfinite(means)` and `variances_bad = ~np.isfinite(variances)` replace the two `np.isnan(...)` predicates.
- `mins`/`maxs` are masked to NaN where non-finite *before* `nanmin`/`nanmax`, so +Inf no longer leaks into `max` and -Inf no longer leaks into `min`.
- New regression `test_aggregate_feature_stats_inf_masked_per_dim` locks this in: a contributor with +/-Inf at every stat of one dim is fully excluded at that dim, clean dims unaffected.

WARN text stays "non-finite values in ..." since the aggregator now genuinely excludes the same set.

src/opentau/datasets/compute_stats.py
shuheng-liu: Added in 689546f. The docstring's "Non-finite-tolerant per-dim" paragraph now ends with:

> The mean and variance use *different* per-dim masks: a contributor is dropped from the mean at a dim iff its `mean` is non-finite there, but dropped from the variance at a dim iff *either* its `mean` or its `std` is non-finite there (you can't form a Chan-style variance contribution without a finite `std`). `count` is unconditional: the sum of contributors' counts, not a per-dim effective contributor count.

Also restored the algorithm note as an inline `# Chan-style parallel weighted variance, non-finite-masked per dim.` above the variance block (the original "parallel algorithm" comment had been dropped in the rewrite, as you noted under nits).

src/opentau/datasets/dataset_mixture.py
shuheng-liu: Tightened in 689546f — dropped the trailing `data` interpolation. Message is now:

```
"Dataset %r: non-finite values in %r.%r at flat indices %s (shape=%s); "
"these dims are excluded from aggregation."
```

Reads cleanly as e.g. `Dataset 'TensorAuto/foo-overlay': non-finite values in 'actions'.'min' at flat indices [0, 1, 2] (shape=(32,)); these dims are excluded from aggregation.`

shuheng-liu (commented):
## Re-review of 689546f

All three findings from the prior review are addressed in 689546f. Verification against the new diff:

### 1. Predicate mismatch (was: WARN catches non-finite, aggregator only masks NaN) — ✅ resolved

Author went with **option 2** (broaden the aggregator). `compute_stats.py:314-315` now defines `means_bad = ~np.isfinite(means)` / `variances_bad = ~np.isfinite(variances)` and uses those as the per-dim masks in `np.where`. Min/max are also protected: `compute_stats.py:340-341` masks ±Inf to NaN *before* `nanmin`/`nanmax` (`mins_safe = np.where(np.isfinite(mins), mins, np.nan)`), so `+Inf` no longer leaks into `max` and `-Inf` no longer leaks into `min`. WARN text and aggregator behavior are now consistent. New regression `test_aggregate_feature_stats_inf_masked_per_dim` locks in the +/-Inf path.

### 2. Asymmetric NaN handling between mean and variance — ✅ resolved

Two new docstring paragraphs in `aggregate_feature_stats` explicitly call out (a) which contributors get dropped from the mean vs. the variance and (b) that `count` is unconditional ("sum of contributors' counts, not a per-dim effective contributor count"). Plus the algorithmic note `# Chan-style parallel weighted variance, non-finite-masked per dim.` is back inline above the variance block — that addresses the dropped-comment nit from the last review too.

### 3. Duplicate `data` interpolation in WARN — ✅ resolved

`dataset_mixture.py:215-222` — trailing `data` interpolation removed. Message now reads `Dataset 'TensorAuto/foo-overlay': non-finite values in 'actions'.'min' at flat indices [0, 1, 2] (shape=(32,)); these dims are excluded from aggregation.` Clean.

### LGTM to ship

The fix is materially better than the v1: the aggregator now genuinely excludes every non-finite contributor (NaN and ±Inf), the docstring explains the per-dim semantics including the mean/variance asymmetry, and the test suite covers NaN per-dim, ±Inf per-dim, and all-non-finite-dim-stays-NaN. CPU tests, pre-commit, and check-checklist were all green on the prior commit and the new diff is additive (extra masks + new test + docstring), so I'd expect CI to stay green on 689546f.

**Note on thread resolution:** I tried to programmatically resolve the three review threads I opened, but the available MCP tooling doesn't expose `PullRequestReviewThread` node IDs in the `get_review_comments` response (only `is_resolved`/`is_outdated`/`is_collapsed`/`comments`), and `resolve_review_thread` requires the GraphQL node ID. Please resolve the three threads via the PR UI — all three are addressed.

---

WilliamYue37 (approved):

PR #295, opened 2026-05-12
feat(datasets): bucket speed by duration-seconds, not native frames

shuheng-liu:
## What this does

The `speed` optional key emitted by `LeRobotDataset.__getitem__` was a coarse "how long is this episode" bucket computed in **native frames**:

```python
item["speed_raw"] = int(round(self.episode_lengths[ep_idx] / 500) * 500)
```

That's FPS-blind: a 1500-frame episode buckets to `speed=1500` whether it's 50 s at 30 fps or 25 s at 60 fps. In a `WeightedDatasetMixture` that mixes datasets with different native FPS and resamples them all to `cfg.dataset_mixture.action_freq` (the mixture-level action-rate target plumbed via `delta_timestamps` in `datasets/factory.py`), the same bucket value means different physical durations across datasets. That's the wrong signal to feed the policy.

Switch to a physical-duration bucket — **seconds, rounded to multiples of 10**:

```python
duration_s = self.episode_lengths[ep_idx] / self.fps
item["speed_raw"] = int(round(duration_s / 10) * 10)
```

- Invariant to the dataset's native FPS (uses `self.fps` to normalize).
- Invariant to the mixture's `action_freq` (duration is a physical quantity; resampling doesn't change it).
- 10-second granularity is comparable to the prior 500-frame granularity at typical 30–50 fps datasets (≈10–17 s per bucket).
- Short episodes (<5 s) round down to 0, mirroring the prior "<250 frames → 0" behavior.

Also updates the optional-keys entry in `docs/source/concepts.rst` to describe the new semantics, the FPS-normalization rationale, and the invariance to mixture-level resampling.

## How it was tested

- `pre-commit run --files src/opentau/datasets/lerobot_dataset.py docs/source/concepts.rst` — all hooks pass (ruff lint+format, pyupgrade, bandit, typos, etc.).
- `pytest -m "not gpu" -n auto tests/datasets/ tests/policies/test_pi07_cpu.py` — 456 passed, 7 skipped (GPU fixtures), 0 failed.
- Verified bucketing math manually against `round()`'s banker's-rounding semantics (same as the original code's `round(frames/500)` did): `1500 frames @ 60 fps = 25 s → bucket 20`, `1500 frames @ 30 fps = 50 s → bucket 50`, etc.
- No validators or test assertions in `src/` or `tests/` enforce the old "multiple of 500" rule (verified via `grep -rn "% 500" src/opentau/ tests/` — empty). Policy modules consume `batch["speed"]` only by stringifying it into the metadata prompt, so they're unit-agnostic.

```bash
gh pr checkout 295
uv sync --extra dev
pytest -m "not gpu" -n auto tests/datasets/ tests/policies/test_pi07_cpu.py
```

Quick spot-check that the new bucket is duration-in-seconds:

```bash
python -c "
import torch
from opentau.datasets.factory import make_dataset
from opentau.configs.default import DatasetMixtureConfig, DatasetConfig
# Load any LeRobot dataset with known fps, e.g. lerobot/droid_100 (fps=15)
# and print item['speed'] for episode 0, frame 0. Expect:
#   round((episode_length_frames / 15) / 10) * 10
"
```

## Checklist

### DIFF
--- a/docs/source/concepts.rst
+++ b/docs/source/concepts.rst
@@ -150,12 +150,18 @@ see :ref:`Training-time dropout <standard-data-format-optional-keys-dropout>`.
                                    # segment, differs at segment boundaries). Clipped at episode
                                    # end. Empty string when unavailable.
 
-        "speed": torch.LongTensor,     # Scalar; episode length in frames rounded to the nearest multiple of
-                                       # 500 (so short <250-frame episodes bucket to 0). Populated
-                                       # unconditionally from ``info.json`` — available on every
-                                       # LeRobotDataset regardless of whether the dataset went through
+        "speed": torch.LongTensor,     # Scalar; episode **duration in seconds** rounded to the nearest
+                                       # multiple of 10 (so short <5 s episodes bucket to 0). Computed as
+                                       # ``round(episode_length_frames / fps / 10) * 10`` from
+                                       # ``meta/episodes.jsonl["length"]`` and ``meta/info.json["fps"]``.
+                                       # Using seconds (rather than raw frames) makes the bucket
+                                       # comparable across datasets in a mixture: it is invariant to both
+                                       # the dataset's native FPS and to ``cfg.dataset_mixture.action_freq``
+                                       # (which resamples each dataset to a common rate at sample time).
+                                       # Populated unconditionally — available on every LeRobotDataset
+                                       # regardless of whether the dataset went through
                                        # ``attach_metadata``. Name is historical; think
-                                       # "episode-length bucket".
+                                       # "episode-duration bucket".
         "speed_is_pad": torch.BoolTensor,  # True only when the dataset has no episode-length metadata
                                            # (pure VQA / legacy fake datasets) or when the metadata drop
                                            # rolls in _emit_optional_keys fire at training time.
--- a/src/opentau/datasets/lerobot_dataset.py
+++ b/src/opentau/datasets/lerobot_dataset.py
@@ -1980,7 +1980,8 @@ class LeRobotDataset(BaseDataset):
             quality = self.meta.episodes[ep_idx].get("quality")
             if quality is not None:
                 item["quality_raw"] = int(quality)
-            item["speed_raw"] = int(round(self.episode_lengths[ep_idx] / 500) * 500)
+            duration_s = self.episode_lengths[ep_idx] / self.fps
+            item["speed_raw"] = int(round(duration_s / 10) * 10)
             item.update(self._load_subgoal_frames(ep_idx, frame_in_ep))
 
             item = self._to_standard_data_format(item)
--- a/src/opentau/datasets/lerobot_dataset.py
+++ b/src/opentau/datasets/lerobot_dataset.py
@@ -237,6 +237,24 @@ def suppress_control_mode_warning(repo_id: str) -> None:
     _CONTROL_MODE_WARNED.add(repo_id)
 
 
+_SPEED_BUCKET_SECONDS = 10
+
+
+def speed_duration_bucket_s(num_frames: int, fps: float) -> int:
+    """Bucket an episode's physical duration to the nearest 10 seconds.
+
+    Used to compute the ``speed`` optional key emitted by
+    ``LeRobotDataset.__getitem__``. Working in seconds (rather than native
+    frames) makes the bucket invariant to the dataset's native FPS and to the
+    mixture's ``cfg.dataset_mixture.action_freq`` resampling — see
+    :ref:`standard-data-format-optional-keys` in ``docs/source/concepts.rst``.
+
+    Uses Python's built-in ``round()`` (banker's rounding), e.g. a 25 s
+    episode buckets to 20, not 30.
+    """
+    return int(round((num_frames / fps) / _SPEED_BUCKET_SECONDS) * _SPEED_BUCKET_SECONDS)
+
+
 class DatasetMetadata:
     """Base class for dataset metadata containing info and statistics.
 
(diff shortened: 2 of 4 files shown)


shuheng-liu (commented):
## Review

### Overview

Two-file, +13/-6 change that fixes a real bug: `speed_raw` was bucketed in native frames, so a 1500-frame episode at 30 fps (50 s) and one at 60 fps (25 s) both ended up in the same `1500` bucket. In a `WeightedDatasetMixture` mixing heterogeneous-FPS datasets — all resampled to `cfg.dataset_mixture.action_freq` at sample time — that bucket value was a noisy signal at best. Switching to seconds (`/ self.fps`) with a 10 s granularity makes the bucket physically meaningful and invariant to both native FPS and mixture-level resampling. Doc update is thorough and explains the rationale, not just the new behavior.

### Correctness — looks good

- `self.fps` is `self.meta.fps` (property at `lerobot_dataset.py:1562`) and is always populated for a valid LeRobotDataset, so no `AttributeError` risk.
- The change is inside `if self.standardize:` (`lerobot_dataset.py:1963`), so pure-VQA / fake datasets still fall through to `speed_is_pad=True` as before — no regression.
- Banker's rounding is preserved from the old expression (`round(2.5) → 2`, so 1500 frames @ 60 fps → bucket 20, not 30). Matches the prior code's semantics; worth knowing but not a defect.
- Verified no `% 500` or `/ 500` validators downstream — the consumer (`pi07*/modeling_*::embed_prefix`) only stringifies `speed` into a prompt segment, so it's unit-agnostic.

### Suggestions

1. **Add a small regression test for the bucketing math.** `tests/datasets/test_optional_keys.py` currently feeds `speed_raw=1500` as a fixture into `_emit_optional_keys`, which tests the plumbing but not the new `frames/fps/10` computation. A parametrized test that locks in a few cases would catch future drift:
   ```python
   @pytest.mark.parametrize("frames, fps, expected", [
       (1500, 30, 50),   # 50 s
       (1500, 60, 20),   # 25 s → banker's rounding to 20
       (100,  15,  10),  # 6.67 s
       (50,   30,  0),   # <5 s → 0
   ])
   def test_speed_raw_bucket(frames, fps, expected): ...
   ```

2. **Worth calling out the semantic shift for existing checkpoints in the PR body.** A model that has been trained on `"Speed: 1500"` will, after this lands, see `"Speed: 50"` for the same physical episode. Since the consumer just stringifies into a prompt it isn't a structural break, but anyone continuing training on a pre-PR checkpoint should expect a transient performance dip while the model re-anchors the speed token.

3. **(Nit) Consider hoisting the bucket size to a named constant.** `10` appears twice in the expression and again in the docstring. A module-level `_SPEED_BUCKET_SECONDS = 10` would make granularity tuning a one-liner — but for a 1-line code change this is purely optional.

4. **(Nit) Inline comment about banker's rounding.** The PR description correctly notes that `round()` uses banker's rounding, but anyone editing this code in the future will not see that. A one-liner like `# banker's rounding, e.g. round(2.5) == 2` next to the assignment would prevent surprise. Optional.

### Performance / security

- No allocation changes; one extra float division per `__getitem__`. Negligible.
- No new I/O, no new external inputs — nothing security-relevant.

### Verdict

Clean, surgical fix that matches the project's "minimum code that solves the problem" guideline. Approve once optional test coverage is added (or accept as-is and follow up with a test in a separate PR — either is fine).

---

shuheng-liu:
Thanks for the review! Addressed in [e2123ba](https://github.com/TensorAuto/OpenTau/commit/e2123ba):

**1. Regression test for the bucket math** ✅ — Extracted a module-level helper `speed_duration_bucket_s(num_frames, fps)` in `lerobot_dataset.py` so the formula is directly testable. Added `tests/datasets/test_optional_keys.py::TestSpeedDurationBucket` with a parametrized test covering: FPS invariance (`(1500, 50, 30)` and `(900, 30, 30)` pair to the same physical duration), banker's-rounding edge cases (`(1500, 60, 20)`, `(450, 30, 20)`), the <5 s → 0 floor (`100, 30, 0`), and the empty-episode degenerate case. Also pins the return type to `int` (`speed_raw` is consumed via `int(raw)` downstream which would silently coerce a float — better to fail loud).

**3. Hoist bucket size to a named constant** ✅ — Naturally addressed by the helper extraction. `_SPEED_BUCKET_SECONDS = 10` now lives at module scope; tuning granularity is a one-liner.

**4. Banker's-rounding comment** — Took a slightly different tack: instead of an inline comment, the banker's-rounding semantics are now documented (a) in the helper's docstring (`a 25 s episode buckets to 20, not 30`) and (b) by example in the parametrized test cases (`(1500, 60, 20),  # 25 s at 60 fps → banker's rounding to 20`). The test cases are harder to silently drift than a comment.

**2. Pretrained-checkpoint heads-up in PR body** — Skipped per offline clarification that no pretrained checkpoints exist yet, so there's nothing to invalidate.

shuheng-liu (commented):
## Follow-up review on `e2123ba`

The new commit cleanly addresses three of the four points from my prior review.

### What changed since last review

- Extracted `speed_duration_bucket_s(num_frames, fps) -> int` at module scope with a docstring that calls out the banker's-rounding behavior with a concrete example. (Addresses prior nits 3 + 4.)
- Hoisted `_SPEED_BUCKET_SECONDS = 10` so granularity is tunable in one place.
- Added `TestSpeedDurationBucket` in `tests/datasets/test_optional_keys.py` with 9 parametrized cases + a return-type assertion. (Addresses prior suggestion 1.)

### Tests look right

I re-derived all 9 parametrized cases — every row checks out, including the two banker's-rounding boundaries (`1500@60 → 25 s → 20`, `450@30 → 15 s → 20`) and the FPS-invariance pair (`1500@50` vs `900@30`, both 30 s, both bucket 30). The `(0, 30, 0)` degenerate-empty-episode case is a nice belt-and-braces — `__getitem__` shouldn't ever see a zero-length episode, but the helper now provably handles it.

### Small observations

- `fps: float` in the signature is the right call (older video data can be 29.97 fps), even though `LeRobotDataset.fps` returns `int` today.
- The `test_returns_python_int` guard uses `type(out) is int` (strict) rather than `isinstance`. That correctly rejects `np.int64` / `np.intp`, which would otherwise propagate into the `torch.LongTensor` conversion with a slightly different code path. Good call.
- The helper is exported without a leading underscore. Fine — it's a small pure function and being testable is worth the public-API surface.

### Remaining (non-blocking)

The only point from my prior review not addressed is the suggestion to call out in the PR body that existing trained checkpoints will see `"Speed: 50"` where they previously saw `"Speed: 1500"` for the same physical episode. Since the consumer just stringifies into a prompt token, it's a soft change, not a structural break — purely a heads-up for anyone resuming training. Optional.

### Verdict

LGTM. Ready to merge once CI is green.

---

PR #297, opened 2026-05-12
fix(metrics): only report L1/Accuracy when policy returns them

shuheng-liu:
## What this does

Stops reporting `l1_loss` and `accuracy` in the training and validation paths when the active policy doesn't emit them. Today every step calls `accelerator.gather_for_metrics(losses.get("L1", zero))` / `.get("Accuracy", zero)` and feeds those into the tracker + wandb, even though only `ValueFunction` populates those keys — every VLA policy (`pi0`, `pi05`, `pi05_mem`, `pi06`, `pi07/*`, `pi07_paligemma/*`) returns just `MSE` + `CE`. The result is two extra all-gathers per step and constant-zero `Training/L1 Loss` / `Training/Accuracy` / `Validation/*/L1 Loss` / `Validation/*/Accuracy` series cluttering wandb.

Changes in `src/opentau/scripts/train.py`:
- `update_policy`: gather L1/Accuracy only when the key is in `losses`; lazy-allocate the meter on first observation. Drop the `_first_loss_tensor` / `zero` plumbing — no longer needed once the `.get(..., zero)` fallback is gone.
- Validation loop: same pattern (gather, lazy meter, conditional logging) for the per-dataset trackers.
- Train + per-dataset + aggregate log blocks: gate the `accelerator.log` calls for L1/Accuracy on meter / key presence.
- `_make_val_tracker` and the training tracker construction: drop the always-on `l1_loss` / `accuracy` meters so the Python log line is clean for VLA policies.
- `_mixture_weighted_aggregate`: derive the metric key set from the trackers themselves instead of the hardcoded `_VAL_METRIC_KEYS` tuple (constant removed).

Rank-safety: the gating decision is determined by the policy class (identical on every rank and deterministic over batches), so the `gather_for_metrics` collective count stays aligned across ranks — matches the invariant in `CLAUDE.md` rule 5.

Label: 🐛 Bug

## How it was tested

- `pre-commit run --files src/opentau/scripts/train.py tests/scripts/test_train.py` — all hooks pass (ruff reformatted once, then clean).
- `pytest tests/scripts/test_train.py` — 17/17 pass, including:
  - `test_aggregates_only_keys_present_on_trackers` (new) — confirms the aggregate handles partial-meter trackers (the production VLA case).
  - `test_empty_trackers_returns_empty_dict` (updated) — empty input now returns `{}` since we can't derive metric keys without a tracker.
- Confirmed the unrelated CPU-suite failures (HF Hub network timeouts, missing `LIBERO_CONFIG_PATH`, weights downloads) reproduce on the base branch — they're sandbox/network issues, not caused by this PR.

```bash
git fetch origin claude/fix-metric-reporting-QkoKJ
git checkout claude/fix-metric-reporting-QkoKJ
pytest -sx tests/scripts/test_train.py
```

After running a smoke training with a VLA policy (e.g. `configs/examples/pi05_training_config.json`), the wandb dashboard should no longer show `Training/L1 Loss`, `Training/Accuracy`, or the corresponding validation series.

## Checklist

---
_Generated by [Claude Code](https://claude.ai/code/session_01TWGQmJHHYm7qaBBY5Pjkw9)_

### DIFF
--- a/src/opentau/scripts/train.py
+++ b/src/opentau/scripts/train.py
@@ -108,14 +108,23 @@ def update_policy(
     # This calls `torch.distributed.all_gather_into_tensor` under the hood, which is not so efficient.
     # We don't actually want to broadcast the gathered tensors to all processes, but only to the main process.
     # Nonetheless, we still do this for correctness, safety, and simplicity.
-    _first_loss_tensor = next(lt for lt in losses.values() if isinstance(lt, torch.Tensor))
-    zero = torch.tensor(0.0, device=_first_loss_tensor.device, dtype=_first_loss_tensor.dtype)
+    # ``L1`` and ``Accuracy`` are optional — only the value head currently returns
+    # them; the VLA policies (pi0/pi05/pi06/pi07) return only ``MSE`` and ``CE``.
+    # Gating the gather on ``key in losses`` keeps the collective count aligned
+    # across ranks because the keys returned by ``forward`` are determined by
+    # the policy class, which is identical on every rank.
     loss = accelerator.gather_for_metrics(loss).mean().item()
     mse_loss = accelerator.gather_for_metrics(losses["MSE"]).to(dtype=torch.float32).mean().item()
     ce_loss = accelerator.gather_for_metrics(losses["CE"]).to(dtype=torch.float32).mean().item()
-    l1_loss = accelerator.gather_for_metrics(losses.get("L1", zero)).to(dtype=torch.float32).mean().item()
+    l1_loss = (
+        accelerator.gather_for_metrics(losses["L1"]).to(dtype=torch.float32).mean().item()
+        if "L1" in losses
+        else None
+    )
     accuracy = (
-        accelerator.gather_for_metrics(losses.get("Accuracy", zero)).to(dtype=torch.float32).mean().item()
+        accelerator.gather_for_metrics(losses["Accuracy"]).to(dtype=torch.float32).mean().item()
+        if "Accuracy" in losses
+        else None
     )
     # This actually calls `.update` method of the `AverageMeter` class. This operation is not idempotent.
     # See MetricsTracker.__setattr__ for more details.
@@ -125,20 +134,22 @@ def update_policy(
         train_metrics.loss = loss
         train_metrics.mse_loss = mse_loss
         train_metrics.ce_loss = ce_loss
-        train_metrics.l1_loss = l1_loss
-        train_metrics.accuracy = accuracy
+        if l1_loss is not None:
+            if "l1_loss" not in train_metrics.metrics:
+                train_metrics.metrics["l1_loss"] = AverageMeter("l1_loss", ":.6f")
+            train_metrics.l1_loss = l1_loss
+        if accuracy is not None:
+            if "accuracy" not in train_metrics.metrics:
+                train_metrics.metrics["accuracy"] = AverageMeter("accuracy", ":.3f")
+            train_metrics.accuracy = accuracy
         train_metrics.lr = optimizer.param_groups[0]["lr"]
 
     return train_metrics
 
 
-_VAL_METRIC_KEYS: tuple[str, ...] = ("loss", "mse_loss", "ce_loss", "l1_loss", "accuracy")
-
-
 def _mixture_weighted_aggregate(
     per_dataset_trackers: dict[str, MetricsTracker],
     name_to_weight: dict[str, float],
-    metric_keys: tuple[str, ...] = _VAL_METRIC_KEYS,
 ) -> dict[str, float]:
     """Mixture-weighted average of per-dataset validation metrics.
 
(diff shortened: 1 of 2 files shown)


shuheng-liu:
## Review

### Overview
Targeted, well-scoped fix. Removes the two constant-zero all-gathers per train step (and per val batch) that fired for `L1`/`Accuracy` whenever the active policy didn't emit them, plus the corresponding `Training/L1 Loss`, `Training/Accuracy`, and `Validation/*/L1 Loss`, `Validation/*/Accuracy` series from wandb. Premise verified:
- `grep` over `src/opentau/policies/**/*.py` confirms only `policies/value/modeling_value.py:299-304` returns `{"MSE", "CE", "L1", "Accuracy"}`. Every other policy (`pi0`, `pi05`, `pi05_mem`, `pi06`, `pi07/{high_level_planner,low_level}`, `pi07_paligemma/*`) returns only `{"MSE", "CE"}` — deterministically, on every batch.
- Value-head training is preserved: the lazy meter is created on the first batch on every rank's main process, so all four metrics keep showing up end-to-end.

### Correctness
- **Rank-safety holds.** The gating decision (`"L1" in losses`, `"Accuracy" in losses`) is determined by the policy class, which is identical on every rank, so all-gather counts stay aligned across ranks — consistent with `CLAUDE.md` rule 5.
- **Meter mutation is main-process only.** Both lazy-allocation sites are inside `if accelerator.is_main_process:` blocks, so `tracker.metrics` only diverges across ranks in the same way it already did (training-side averaging was already main-process only).
- **`_mixture_weighted_aggregate` keys.** Deriving `metric_keys` from `next(iter(per_dataset_trackers.values())).metrics.keys()` works because every dataset's tracker has seen at least one batch by then (empty datasets are filtered out upstream by `WeightedDatasetMixture.get_per_dataset_dataloaders`, per the existing docstring) and every batch's `forward` returns the same key set. The new docstring documents this load-bearing assumption.
- No dangling references to the removed `_VAL_METRIC_KEYS` constant anywhere under `src/opentau/` or `tests/` — confirmed via grep.

### Tests
- 17/17 pass locally (`uv run pytest tests/scripts/test_train.py -v`).
- `test_aggregates_only_keys_present_on_trackers` directly mirrors the production VLA case (trackers with only `loss`/`mse_loss`/`ce_loss`) — good regression guard.
- `test_empty_trackers_returns_empty_dict` is correctly updated for the new "can't invent keys" behavior. The new docstring + test comment both note that production never hits this path, which is the right framing.

### CI
All checks green: `review`, `Pre-commit Hooks`, `CPU Tests`, `check-checklist`.

### Suggestions (non-blocking)

1. **DRY the lazy-meter pattern.** The `if "x" not in tracker.metrics: tracker.metrics["x"] = AverageMeter(...); tracker.x = value` block is repeated 4 times (train+val × l1+accuracy). A tiny helper would tighten it and make the intent explicit:
   ```python
   def _observe_optional(tracker, name, fmt, value):
       if value is None:
           return
       if name not in tracker.metrics:
           tracker.metrics[name] = AverageMeter(name, fmt)
       setattr(tracker, name, value)
   ```
   Then each site collapses to two calls. Mostly a readability win.

2. **Direct mutation of `tracker.metrics` is a minor leaky abstraction.** `MetricsTracker` was designed to be constructed with a fixed metric set and has no `register_metric()` method. The PR works around this by reaching into the dict, which is fine pragmatically but slightly muddies the class's contract. Adding a `register_metric(name, meter)` method on `MetricsTracker` would make the intent clearer and centralize the "what if already exists?" decision. Optional follow-up, not for this PR.

3. **Defensive assertion in `_mixture_weighted_aggregate`.** Since the new docstring makes "all trackers share the same meter set" a load-bearing invariant, a one-line guard would catch future regressions faster than a missing key in wandb:
   ```python
   metric_keys = tuple(next(iter(per_dataset_trackers.values())).metrics.keys())
   if __debug__ and len(per_dataset_trackers) > 1:
       for name, t in per_dataset_trackers.items():
           assert tuple(t.metrics.keys()) == metric_keys, f"tracker {name} diverges: {tuple(t.metrics.keys())} vs {metric_keys}"
   ```
   Up to you — also pure taste.

### Risks
None I can spot. The change is local to `scripts/train.py`, the rank-safety reasoning is sound, the value-head path is preserved by the lazy-meter creation, and the new test pins the behavior that matters in production. LGTM as-is; the suggestions above are stylistic.


---

shuheng-liu:
## Re-review (head `87d63d6`)

### What changed since last review
New commit `87d63d6` `refactor(train): collapse lazy-meter pattern into _observe_optional` addresses suggestion #1 from my previous review. Pure refactor, no behavior change.

### Review

- **Helper signature is clean.** `_observe_optional(tracker, key, display_name, fmt, value)` correctly separates the dict key (used both as `tracker.metrics[key]` and as the attribute name in `setattr`) from the `AverageMeter`'s display name — that's the right split given train uses `"l1_loss"` as both key and display name, while val uses `"l1_loss"` as the key and `"val_l1_loss"` as the display name. Existing convention preserved.
- **Docstring is excellent.** Explicitly documents the main-process precondition (`"Callers must already be on the main process..."`), which is the non-obvious safety property that matters for distributed correctness.
- **`None` short-circuit is right.** When the upstream gather returned `None` (key wasn't in `losses`), the helper no-ops, so the meter is never lazily allocated for a policy that never emits the key. Matches the previous inline behavior.
- **`setattr(tracker, key, value)` correctly routes through `MetricsTracker.__setattr__`**, which dispatches to `AverageMeter.update(value)` (per `utils/logging_utils.py:137-143`). Same semantics as the previous `tracker.l1_loss = value` form.
- **Module placement.** Helper is defined right after `update_policy`, before `_mixture_weighted_aggregate` — sensible location.
- **Test coverage unchanged and still relevant.** The helper isn't directly tested but its behavior is exercised end-to-end through `update_policy` and the val loop; `test_aggregates_only_keys_present_on_trackers` still pins the production VLA case.
- **Commit hygiene.** Subject line 65 chars, body explains the "why" (referencing this PR's review) and the net delta. 

### Verification
- `uv run pytest tests/scripts/test_train.py -v` → 17/17 pass on `87d63d6`.
- All CI green: `review`, `Pre-commit Hooks`, `CPU Tests`, `check-checklist`.

### Verdict
LGTM. The refactor improves readability without altering semantics; my remaining suggestions from the prior review (`register_metric()` on `MetricsTracker`, optional cross-tracker key-parity assert in `_mixture_weighted_aggregate`) are still optional and not for this PR. Ready to merge from my side.


---

PR #273, opened 2026-05-06
feat(pi07,fsdp): enable FSDP-FULL_SHARD for full unfreeze + profile_step audit + ZeRO-2 vs FSDP matrix

shuheng-liu:
## What this does

Enables FSDP-FULL_SHARD training for pi07_low_level (full unfreeze), audits
and fixes `profile_step.py` measurement bugs that caused unreconcilable
numbers in the prior closed PR, and runs a full FSDP vs ZeRO-2 throughput
+ peak HBM matrix on 8×A100 to ground the backend recommendation for the
upcoming cam=4 pretraining run.

The branch carries the FSDP-enabling code from earlier work (the closed
PR was on this same branch) plus two new commits with the
measurement-side fixes:

- `aaefa6e` — `InterleavedDecoderLayer` refactor in
  `policies/pi07/gemma3_with_expert.py`. Bundles each (Gemma3 backbone
  layer + Gemma-v1 expert layer) pair into one `nn.Module` so the FSDP
  all-gather hook fires before any sub-component access. Without this,
  FSDP1 hangs at NCCL with mismatched all-gather sizes.
- `1f27b98` — `_global_any` branch sync in
  `policies/pi07/low_level/modeling_pi07_low_level.py`. OR-reduces
  per-rank decisions to enter `if has_response / has_subgoal /
  has_metadata` blocks so all ranks issue the same number of FSDP
  collectives.
- `30ff84a` — `disable_internal_bf16_cast` flag on
  `Gemma3WithExpertConfig` plus FSDP-conditional setter in `train.py`
  and `profile_step.py`. Keeps the policy fp32 going into FSDP so
  MixedPrecision can manage the bf16-compute / fp32-master split itself
  (preserves fp32 Adam state).
- `f43fc91` — `unset HfDeepSpeedConfig` context manager so
  `make_policy` doesn't fault under ZeRO-3 init.
- `63df8c5` — relax the `gradient_checkpointing` guard to allow FSDP;
  ships `accelerate_fsdp_config.yaml` and
  `accelerate_deepspeed_zero3_config.yaml`.
- `3bd7817` — **NEW** four `profile_step.py` bug fixes (B1-B4) +
  DeepSpeed phase-blur banner. Detail below.
- `ffd6208` — **NEW** `pi07_low_level_libero.json` `n_obs_history` 8→6
  for the matrix.

### profile_step.py audit findings (fixed in 3bd7817)

The prior numbers couldn't reconcile because the script had four real
[...]

PR #298, opened 2026-05-12
feat(pi07): add ring-attention sequence parallelism

WilliamYue37:
## What this does

Adds a paper-style ring attention implementation
([Liu, Zaharia & Abbeel, 2023](https://arxiv.org/abs/2310.01889)) to the
pi07 prefix forward as a new `attention_implementation="ring"` choice
alongside `eager` / `sdpa`.

Each rank holds 1/W of the sequence along the seq axis; K/V rotate
around the ring via NCCL P2P (`isend`/`irecv`) while an online softmax
accumulates the per-rank output. Forward and backward are a custom
`torch.autograd.Function` so only Q, K, V, the final O, and the per-row
log-sum-exp survive the forward → backward boundary — full `(S, S)`
attention-score matrices are never stored on any rank. The per-block
score / probability tensors are also Q-tiled to keep peak memory bounded
by a small constant (`_RING_Q_TILE`, default 256) rather than `Sq_local
* Sk_block`.

Per the paper's Section 3.2.2, that blockwise rematerialisation
replaces the legacy per-layer `torch.utils.checkpoint` wrapping: under
ring, `gradient_checkpointing` is ignored (and reported in the
`attention_implementation` docstring as such).

Suffix (action-expert) forwards transparently route through SDPA via a
small `_ring_active` dispatch flag on the model — they are short and
cross-attend to the already-cached prefix K/V, so ring's fixed costs
outweigh the benefit. KV-cache writes `gather_seq` across ranks before
storing so the suffix forward sees the full prefix cache.

Scope is intentionally narrow: pi07 low-level training, prefix forward
only. The high-level planner is unchanged.

🗃️ Feature

### Hardware
2x NVIDIA RTX 3090 (24 GiB each) on a single node. The implementation
itself is hardware-agnostic — it uses standard NCCL collectives over
the default process group, so it transparently inherits the production
A100 + NVLink + InfiniBand path.

### Numerical correctness
Two scripts under `src/opentau/scripts/ringattn_experiments/`:

* `unit_test.py` — kernel-level: builds random Q/K/V, runs the eager
  reference and `_RingAttention.apply` on the same data, checks the
  per-rank slice of the output matches.

  ```
  [rank 0] ref local |max| = 1.2188, max |ref - ring| = 0.000e+00, mean = 0.000e+00
  [rank 1] ref local |max| = 0.9453, max |ref - ring| = 1.192e-07, mean = 2.183e-11
  PASS
  ```

* `correctness.py` — full-model: a tiny Gemma 3 with expert, prefix
  forward run under `eager`, `sdpa`, and `ring`. Ring should be no
  further from SDPA than SDPA is from eager (bf16 reassociation noise
  the model already absorbs today).

  ```
  ||eager|| = 512.0071   ||sdpa|| = 511.9952   ||ring|| = 511.9866
  max |eager - sdpa|  = 2.031e-01   mean = 2.706e-02
  max |eager - ring|  = 1.914e-01   mean = 2.609e-02
  max |sdpa  - ring|  = 2.734e-01   mean = 2.285e-02
  PASS
  ```

### Memory scaling (`memory_scaling.py`)
hidden=256, 1 backbone layer, ws=2, forward + backward:

| seq_len | sdpa (GiB) | ring (GiB) | reduction |
| ------: | ---------: | ---------: | --------: |
| 2048    | 0.077      | 0.075      | 2.6%      |
| 8192    | 0.423      | 0.285      | 32.8%     |
| 32768   | 3.879      | 3.036      | 21.7%     |
| 65536   | 13.681     | 12.052     | 11.9%     |

### Longest feasible context (`longest_context.py`)
Same harness, binary-search the largest seq_len that completes a
forward+backward without OOM:

* SDPA: largest OK ≈ **75264** tokens, first OOM at 75520.
* Ring: largest OK ≥ **81920** tokens (first OOM only seen at 98304).

So ring extends the feasible context by at least **~9%** on this
harness, with headroom for a more refined search.

### Existing tests
`pytest tests/policies/ -m "not gpu" -k "pi07 or pi06"` — 222 passed,
98 deselected (the GPU subset is gated for the nightly runner). No
existing tests regress.

```bash
git checkout feat/pi07-ring-attention
source .venv/bin/activate

# Kernel-level numerical sanity (any 2-GPU box):
torchrun --nproc_per_node=2 -m opentau.scripts.ringattn_experiments.unit_test

# Full-model output drift vs sdpa/eager:
torchrun --nproc_per_node=2 -m opentau.scripts.ringattn_experiments.correctness

# Memory scaling table:
torchrun --nproc_per_node=2 -m opentau.scripts.ringattn_experiments.memory_scaling

# Longest feasible context (slow — takes minutes per arm):
torchrun --nproc_per_node=2 -m opentau.scripts.ringattn_experiments.longest_context

# To enable ring on a real config, set:
#   --policy.gemma3_with_expert_config.attention_implementation=ring
# in any pi07 training command (only needs >=2 GPUs to do anything useful).
```

## Checklist

GPU pytest / nightly regression runs are pending — they need an A100
box; the 3090 harness here is for the experiment numbers above.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -183,6 +183,9 @@ select = ["E4", "E7", "E9", "F", "I", "N", "B", "C4", "SIM"]
 "src/opentau/scripts/grpc/server.py" = ["N802"]
 # Uppercase names for rotation matrices follow standard math convention
 "src/opentau/scripts/recordhuman_to_lerobot.py" = ["N803", "N806"]
+# Math-convention names (B, S, Hq, Dh, O, LSE, P, ...) for ring attention.
+"src/opentau/policies/pi07/ring_attention.py" = ["N803", "N806", "E741"]
+"src/opentau/scripts/ringattn_experiments/*.py" = ["N803", "N806", "E741"]
 
 [tool.bandit]
 exclude_dirs = [
--- a/src/opentau/configs/train.py
+++ b/src/opentau/configs/train.py
@@ -155,6 +155,14 @@ class TrainPipelineConfig(HubMixin):
     dataloader_batch_size: int | None = None
     # Prefetch factor for the dataloader.
     prefetch_factor: int | None = None
+    # Ring-attention sub-group size. When set (and the policy enables
+    # ``attention_implementation="ring"``), splits the WORLD process group
+    # into ``world_size / ring_group_size`` ring sub-groups of this size.
+    # Sequence parallelism happens within each ring; DP-replication happens
+    # across rings (ZeRO over WORLD scales the gradient correctly via a
+    # ``loss *= ring_group_size`` pre-backward, see train.py). None keeps the
+    # legacy single-ring behaviour (ring spans WORLD). Must divide world_size.
+    ring_group_size: int | None = None
     steps: int = 100_000
     log_freq: int = 200
     save_checkpoint: bool = True
@@ -285,6 +293,39 @@ class TrainPipelineConfig(HubMixin):
                         f"dataset_mixture.n_obs_history ({dm.n_obs_history}; "
                         "treated as 1 when unset). Set dataset_mixture.n_obs_history "
                         "to match policy.n_obs_steps."
+                    )
+
+        # Ring sub-group validation. ``validate()`` runs before the Accelerator
+        # initialises distributed, so probe ``WORLD_SIZE`` from the env (set
+        # by torchrun / accelerate) rather than ``dist.get_world_size()``.
+        if self.ring_group_size is not None:
+            import os as _os
+
+            if self.ring_group_size < 1:
+                raise ValueError(f"ring_group_size must be >= 1; got {self.ring_group_size}.")
+            world_size_env = _os.environ.get("WORLD_SIZE")
+            if world_size_env is not None:
+                world_size = int(world_size_env)
+                if world_size % self.ring_group_size != 0:
+                    raise ValueError(
+                        f"world_size ({world_size}) must be divisible by "
+                        f"ring_group_size ({self.ring_group_size})."
+                    )
+            # Ring sub-grouping is only meaningful when the policy actually
+            # consumes ring attention. pi07 exposes the flag directly on the
+            # policy config (it's plumbed into ``vlm_config`` in
+            # ``__post_init__``); other policies (pi05, etc.) don't have the
+            # field at all — for those we treat ring_group_size as a no-op
+            # rather than erroring, so a shared launcher script can leave
+            # the flag set without breaking non-pi07 runs.
+            if self.policy is not None:
+                impl = getattr(self.policy, "attention_implementation", None)
+                if impl is not None and impl != "ring":
+                    raise ValueError(
+                        "ring_group_size is set but "
+                        f"policy.attention_implementation = {impl!r}. "
+                        "Set attention_implementation='ring' or unset "
+                        "ring_group_size."
                     )
 
     @classmethod
--- a/src/opentau/datasets/dataset_mixture.py
+++ b/src/opentau/datasets/dataset_mixture.py
@@ -488,11 +488,22 @@ class WeightedDatasetMixture:
                 overrides[dataset_cfg.repo_id] = dataset_cfg.data_features_name_mapping
         return overrides
 
-    def get_dataloader(self) -> DataLoader:
+    def get_dataloader(self, sampler_seed: int | None = None) -> DataLoader:
         """Create and return a PyTorch DataLoader with weighted sampling.
 
         Uses HierarchicalSampler to first sample a dataset according to weights,
         then uniformly sample within that dataset.
+
+        Args:
+            sampler_seed: Optional seed for the HierarchicalSampler's RNG.
+                Identical seeds across ranks produce identical sample streams,
+                which is what 2D ring + DP parallelism needs: ranks in the same
+                ring sub-group should see the same batch (so sequence
+                parallelism is computing one consistent batch), while ranks in
+                different DP sub-groups should see different batches. Callers
+                that want this should pass ``base_seed + dp_rank * large_prime``.
+                None leaves the sampler unseeded (every rank's independent
+                ``torch.Generator()`` state), preserving the pre-ring behaviour.
 
         Returns:
             DataLoader configured for weighted hierarchical sampling.
@@ -535,6 +546,7 @@ class WeightedDatasetMixture:
             dataset_lengths=ds_lengths,
             dataset_probs=self.dataset_weights,
             num_samples=num_samples_per_epoch,
+            seed=sampler_seed,
         )
 
         dataloader = DataLoader(
--- a/src/opentau/policies/pi07/gemma3_with_expert.py
+++ b/src/opentau/policies/pi07/gemma3_with_expert.py
@@ -47,6 +47,12 @@ from transformers import (
 from transformers.models.auto import CONFIG_MAPPING
 from transformers.models.gemma import modeling_gemma
 
+from opentau.policies.pi07.ring_attention import (
+    gather_seq,
+    ring_attention_forward,
+    ring_world_size,
+)
+
 # Ensure the Gemma-v1 AdaRMS / gated-residual patches are live before we
 # construct an action expert. Import for side effects only.
 from opentau.utils import transformers_patch  # noqa: F401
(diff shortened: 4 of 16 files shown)


WilliamYue37:
GPU pytests on the 5090 box (worktree at `~/OpenTau-ringattn`, branch `feat/pi07-ring-attention` @ f35c1fa):

```
pytest -m "gpu" -n 0
...........ssssss....ss..ss.                                             [100%]
18 passed, 10 skipped, 1124 deselected, 5 warnings in 167.82s (0:02:47)
```

10 skips are pre-existing fixture-conditional cases (LIBERO env config etc.) — no regressions from this branch.

shuheng-liu:
A100 benchmark results — ring vs PR #273 SDPA baseline.

## Setup

- 8× A100 80GB, single node
- pi07 full unfreeze, `n_obs_history=6`, bf16, `expandable_segments`
- `gradient_accumulation_steps=1`, `PROFILE_STEPS=10` + 20 warmup (matches PR #273 methodology)
- All ring cells: `--policy.attention_implementation=ring`, `--policy.gradient_checkpointing=true` (ignored under ring, per this PR)
- Probe-walk batch size upward from PR #273's peak / OOM ceiling, per-cam OOM short-circuits

## Matrix

| cell | backend | cam | bs | alloc GiB | resv GiB | step ms | ring sps | PR #273 SDPA sps | Δ % | %resv |
| --- | --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| DS-1 | deepspeed | 4 | 20 | 57.96 | 59.41 |  8353 | 19.15 | 13.70 | +39.8% | 74% |
| DS-2 | deepspeed | 4 | 24 | 64.32 | 66.07 | 10285 | 18.67 | OOM in #273 | — | 83% |
| DS-3 | deepspeed | 4 | 28 | 70.68 | 72.71 | 11367 | 19.71 | not tried | — | 91% |
| DS-4 | deepspeed | 4 | 32 | 77.04 | 79.38 | 12434 | **20.59** | not tried | — | 99% |
| DS-5 | deepspeed | 1 | 40 | 66.03 | 67.25 |  7690 | **41.61** | 26.70 | +55.8% | 84% |
| DS-6 | deepspeed | 1 | 48 | 74.01 | 75.47 |  9415 | 40.79 | OOM in #273 | — | 94% |
| DS-7 | deepspeed | 1 | 56 | — | — | — | OOM | — | — | — |
| FS-1 | fsdp      | 4 |  8 | 54.99 | 71.12 |  7827 |  8.18 |  7.00 | +16.8% | 89% |

FSDP sweep stopped after FS-1 — at 8.2 sps it is 2.5× behind DS-ring's cam=4 peak (20.6 sps), so FSDP cannot beat DS at this scale and the per-task early-stop rule fired. FS-2..FS-6 were not run.

## Headline (peak per backend × cam)

| backend | cam | ring peak | PR #273 SDPA peak | ratio |
| --- | ---: | --- | --- | ---: |
| DeepSpeed ZeRO-2 | 1 | bs=40 → 41.6 sps | 26.7 sps | **1.56×** |
| DeepSpeed ZeRO-2 | 4 | bs=32 → 20.6 sps | 13.7 sps | **1.50×** |
| FSDP-FULL_SHARD  | 4 | bs=8  →  8.2 sps |  7.0 sps |  1.17×   |

## Verdict

**Ring attention is a clear win on the production backend (DeepSpeed ZeRO-2) at both camera counts.** Apples-to-apples at PR #273's batch sizes, ring buys 1.4-1.6× more samples/s. Past that, ring opens batch headroom: cam=4 fits **bs=32** under ring (PR #273 OOM'd at bs=24); cam=1 fits **bs=48** (PR #273 OOM at bs=44).

**Why ring helps more than the seq-length argument predicts.** The +56% throughput jump on cam=1 (whose prefix is relatively short) under-explains by the 2.6%-32.8% attention-memory reduction in this PR's 3090 scaling table alone. The dominant factor is almost certainly that ring subsumes per-layer `gradient_checkpointing` via blockwise rematerialisation in the kernel itself: PR #273's cam=1 ZeRO-2 cell was at 93% HBM where checkpointing was load-bearing both for memory and for throughput, and ring relaxes both at once.

**Production recommendation update vs PR #273.** PR #273 settled on bs=20 cam=4 → 13.7 sps. With ring the new sweet spot is **bs=32 cam=4 → 20.6 sps at 99% HBM** (1.50× PR #273), or **bs=28 cam=4 → 19.7 sps at 91%** if 99% feels too tight. For cam=1 (some pretrain phases): **bs=40 → 41.6 sps at 84% HBM**.

## Notes — two unrelated bugs surfaced en route

1. `configs/examples/pi07_low_level_libero.json` (added in PR #273) has `dataset_mixture.history_interval` and `policy.n_obs_history` fields that draccus rejects — they aren't declared on the corresponding dataclasses. That config currently cannot be loaded via the PR #273 PR-body repro recipe (any `profile_step.py` / `train.py` invocation against it crashes at config-decode). The matrix above used a sibling config with the two stray fields removed. Worth a small follow-up PR.
2. PR #273's PR-body suggests `--policy.gemma3_with_expert_config.disable_internal_bf16_cast=true` is a CLI flag — it isn't accepted by draccus. The flag is auto-applied inside `profile_step.py` / `train.py` when FSDP is detected (lines 217-218 / 392-393). FS-1 above ran with the default behaviour and still saw the ring gain.

## Repro for the new cam=4 production setting

```bash
PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True PROFILE_STEPS=10 \
  accelerate launch \
    --num_processes=8 \
    --config_path=configs/examples/pi07_low_level_libero.json \
    --batch_size=32 --dataloader_batch_size=32 \
    --gradient_accumulation_steps=1 --num_cams=4 \
    --policy.freeze_vision_encoder=false --policy.train_expert_only=false \
    --policy.gradient_checkpointing=true \
    --policy.attention_implementation=ring \
    --wandb.enable=false
```

(needs note 1 fixed first, or a hand-stripped copy of the JSON).

shuheng-liu:
See #300 for a discussion on NaNs caused by numerical instability with BF16 and FP32

WilliamYue37:
Pushed 2D parallelism wiring on top of the original PR:

- \`cfg.ring_group_size\` (TrainPipelineConfig) splits WORLD into ring sub-groups of size R along the intra-node axis. Validated in \`TrainPipelineConfig.validate()\` against \`WORLD_SIZE\` (from env) and the policy's \`attention_implementation\`.
- \`build_ring_and_dp_groups(R)\` creates ring sub-groups, pins via \`set_ring_group\`. DP topology is exposed *arithmetically* (\`world_rank // R\`) without a separate communicator — every extra NCCL sub-group competes with ZeRO's world-group collectives on the same CUDA streams.
- HierarchicalSampler accepts a seed; the trainer keys it on \`dp_rank\` so ranks inside the same ring sub-group draw identical sample streams.
- \`loss *= ring_group_size\` before backward rescales ZeRO's world-MEAN reduce into the correct \`(ring-SUM, DP-MEAN)\` gradient. ZeRO stays over WORLD — no DeepSpeed plugin surgery.

### Bugs found and fixed during testing on a100-training / gpu-387 (bot_387 reservation, ZeRO-2)
1. \`62c4a88\` — dropped a redundant per-step batch broadcast (the seeded sampler already gives identical batches within a ring sub-group; the broadcast was issuing an extra collective per step and could diverge if \`is_cuda\` differed across ranks).
2. \`6ce3cef\` — stopped creating an orthogonal DP communicator. It was unused (DP index is just \`world_rank // R\`), and creating the extra NCCL sub-group correlated with a first-step hang under ZeRO-2.
3. \`dc6b4df\` — \`P2POp\`'s \`peer\` argument is a *global* rank, not a within-group rank. When \`ring_group_size == world_size\` the bug was masked (the sub-group equals WORLD). With a proper sub-group (e.g. \`{4,5,6,7}\`), passing local rank 1 as the peer crashed at the first \`batch_isend_irecv\` with \"Global rank 1 is not part of group\". Now translates the local rotation step to a global peer via \`dist.get_global_rank\`.

### Verification
- Local 2× RTX 3090 \`unit_test.py\` (kernel-level): still bit-equivalent to eager (max abs diff 1.2e-7).
- a100-training, 8× A100, DeepSpeed ZeRO-2, pi07 low-level pretraining config:
  - \`ring_group_size=8\` (single ring) — runs cleanly (job 1310, 20 steps).
  - \`ring_group_size=4\` (2 rings × 4 GPUs) — runs after the three bugfixes above (job 1317 step 1: \`total_loss 64.94 scaled / 16.24 unscaled\`; matches the rs=8 baseline within bf16 noise).

### Known issue (separate from this PR's topology work)
Both \`rs=8\` and \`rs=4\` show \`total_loss = nan\` from step 2 onward in real pi07 training. The same NaN appears when \`ring_group_size\` is *unset* (legacy single-ring across WORLD from this PR's original commit, no loss scaling), so it is not introduced by the 2D wiring. It is a pre-existing bug in the ring attention kernel's real-data bf16 backward — likely in how the online-softmax accumulator interacts with ZeRO-2's bf16 master grads. Needs a separate investigation; tracking as a follow-up.

PR #313, opened 2026-05-20
fix(datasets): set batch_size in v2.0->v2.1 converter fake config

shuheng-liu:
## What this does

Fixes a startup crash in the LeRobot v2.0→v2.1 dataset converter (🐛 Bug).

`create_fake_train_config()` in `src/opentau/datasets/v21/convert_dataset_v20_to_v21.py` builds a `TrainPipelineConfig` without `batch_size` or `dataloader_batch_size`. `TrainPipelineConfig.__post_init__` now raises when both are unset:

```
ValueError: At least one of `batch_size` and `dataloader_batch_size` should be set.
```

So `convert_dataset(...)` — and the `convert_dataset_v20_to_v21.py` CLI — failed immediately, before loading any data. This sets `batch_size=1`; the conversion only reads data to recompute per-episode stats and never trains, so any valid batch size works.

## How it was tested

- Confirmed `create_fake_train_config()` now constructs successfully and returns a config with `batch_size=1`, `dataloader_batch_size=1`, `gradient_accumulation_steps=1` (the invariant `batch_size == dataloader_batch_size * gradient_accumulation_steps` holds). Before this change it raised `ValueError`.
- `pre-commit run --files src/opentau/datasets/v21/convert_dataset_v20_to_v21.py` passes (ruff, ruff-format, pyupgrade, gitleaks, bandit, etc.).

```bash
python -c "from opentau.datasets.v21.convert_dataset_v20_to_v21 import create_fake_train_config; print(create_fake_train_config().batch_size)"
```

Prints `1` instead of raising `ValueError`.

## Checklist

### DIFF
--- a/src/opentau/datasets/v21/convert_dataset_v20_to_v21.py
+++ b/src/opentau/datasets/v21/convert_dataset_v20_to_v21.py
@@ -95,7 +95,10 @@ def create_fake_train_config() -> TrainPipelineConfig:
     mixture_cfg = DatasetMixtureConfig(datasets=[dataset_cfg], weights=[1.0])
     policy_cfg = DummyPolicyConfig()
 
-    # Create the main config with minimal required parameters
+    # Create the main config with minimal required parameters.
+    # batch_size is required: TrainPipelineConfig.__post_init__ raises when neither
+    # batch_size nor dataloader_batch_size is set. Conversion never trains, so any
+    # valid value works.
     cfg = TrainPipelineConfig(
         dataset_mixture=mixture_cfg,
         policy=policy_cfg,
@@ -104,6 +107,7 @@ def create_fake_train_config() -> TrainPipelineConfig:
         max_state_dim=32,
         max_action_dim=32,
         action_chunk=50,
+        batch_size=1,
     )
 
     return cfg


shuheng-liu (commented):
**Reviewed — the fix is correct and minimal. LGTM, with two non-blocking follow-ups below.**

I traced the crash and the resolution to confirm the change does what the description says:

- **The crash is real.** `TrainPipelineConfig.__post_init__` raises when both fields are unset (`src/opentau/configs/train.py:184-185`), and `batch_size` / `dataloader_batch_size` both default to `None` (`train.py:153,155`). `create_fake_train_config()` set neither, so `convert_dataset()` died before reading any data.
- **`batch_size=1` resolves cleanly.** The guard passes, `dataloader_batch_size` back-computes to `1 // 1 = 1`, the `batch_size == dataloader_batch_size * gradient_accumulation_steps` invariant holds, and the `>= 1` asserts pass (`train.py:186-203`). Matches the description.
- **The value is irrelevant to correctness.** `convert_dataset()` only uses the config to build `LeRobotDataset` for the stats recompute (`convert_stats`); no DataLoader or training loop is created, so any valid batch size works. The choice is also consistent with the sibling builder `create_mock_train_config()` in `visualize_dataset.py:155` (`batch_size=8`).

Non-blocking follow-ups:

1. **No regression test.** Nothing in `tests/` exercises the v21 converter, so this can silently regress again. The full `convert_dataset()` needs hub access, but the builder doesn't — a one-liner asserting `create_fake_train_config()` constructs and has `batch_size == 1` would lock in the fix cheaply.
2. **Same latent crash elsewhere.** `tests/artifacts/image_transforms/save_image_transforms_to_safetensors.py:103` builds `TrainPipelineConfig(dataset_mixture=..., policy=...)` with no `batch_size`. It lives inside `main()` under a `__main__` guard (the test suite only imports `ARTIFACT_DIR`, so CI is unaffected), but anyone re-running that artifact generator hits the identical `ValueError`. Worth the same one-line fix if you're in here.

---

Issue #266, opened 2026-05-05
pi07 embed_prefix: hoist embed_* calls out of has_* branches to drop the global-sync all-reduces


shuheng-liu:
## Background

Pi07's `Gemma3WithExpertModel.embed_prefix` has data-dependent branches that decide whether to add response / metadata / subgoal / prefix-end blocks based on whether any sample in the *local* batch has real (non-padded) content for that field. Each branch's body calls `embed_language_tokens(...)` and/or `embed_image(...)`, which trigger forward calls into the FSDP-wrapped tree.

Under FSDP / ZeRO-3 with realistic stochastic `*_drop_prob` settings, different ranks roll different drop outcomes, take different branches, and end up issuing different counts of FSDP all-gather collectives → NCCL deadlock. PR #265 fixes this with a `_global_any(local, device)` helper that OR-reduces each branch decision across ranks via a 1-element `MAX` all-reduce, so every rank takes the same branch.

That fix works, but it has two warts:

1. **Three small all-reduces per `embed_prefix`** (~tens of µs total — negligible against step time, but still extra collectives).
2. **Ranks whose local micro-batch has no data for a field still embed pad tokens through the whole prefix** when *some other rank* has that field's content. Their `pad_masks` are all-False so attention/loss correctly ignore them, but every interleaved layer still spends compute carrying those pad slots.

## Proposed cleanup

Hoist every `embed_language_tokens(...)` and `embed_image(...)` call out of the `if has_*` branches. Make the **call** unconditional (so every rank issues the same FSDP all-gather count) but keep the **bookkeeping** (`embs.append`, `pad_masks.append`, `att_masks += [...]`) gated on the local condition.

Sketch:
\`\`\`python
# Always run — uniform FSDP all-gather count across ranks.
response_emb = (
    self.gemma3_with_expert.embed_language_tokens(response_tokens)
    if response_tokens is not None
    else None
)

# Bookkeeping stays conditional on the LOCAL data — no global sync needed.
if response_emb is not None and response_masks is not None and response_masks.any():
    embs.append(response_emb)
    pad_masks.append(response_masks)
    att_masks += [1] * response_emb.shape[1]
\`\`\`

Same shape for `metadata`, `prefix_end`, `subgoal_images` (the subgoal body has 2× `embed_language_tokens` + N× `embed_image`; all should be hoisted).

Different ranks would then have different prefix lengths — that's OK because:
[...]

PR #250, opened 2026-05-04
refactor(pi07): rename low_level_planner -> low_level (it's not a planner)

shuheng-liu:
## What this does

Closes #247.

The low-level component of π0.7 is a flow-matching action generator with FAST discrete-token prediction — not a planner. The high-level component IS a planner (predicts subtask strings autoregressively) and keeps its naming. This PR aligns the source-side naming with that distinction.

### Renames

- `src/opentau/policies/pi07/low_level_planner/` → `src/opentau/policies/pi07/low_level/`
- `PI07LowLevelPlannerConfig` → `PI07LowLevelConfig`
- `PI07LowLevelPlannerPolicy` → `PI07LowLevelPolicy`
- `PI07LowLevelPlannerFlowMatching` → `PI07LowLevelFlowMatching`
- `TestPI07LowLevelPlannerAttentionLayout` → `TestPI07LowLevelAttentionLayout` (in `tests/policies/test_pi07_attention_layout.py`)

Updated callers: `factory.py`, `policies/__init__.py`, `gemma3_with_expert.py`, plus the test suite (`test_pi07_cpu.py`, `test_pi07_attention_layout.py`, `test_pi07_low_level.py`, `test_pi07_video_encoder_cpu.py`, `test_available.py`). Docstrings that mentioned "low-level planner" were updated to "low-level component" for consistency.

### Backward compatibility (deprecation shim)

The old `low_level_planner/` path is preserved as a thin re-export shim that aliases the new classes under their old names and emits a `DeprecationWarning` at package-import time. This keeps existing user code, pickled checkpoints (config classes), and any draccus configs that pinned the old import path working through one release cycle.

The shim is exercised by `tests/policies/test_pi07_low_level_shim.py`:

- Verifies the warning fires on import.
- Verifies legacy `PI07LowLevelPlanner*` names are the **same class objects** as the new `PI07LowLevel*` names (so `isinstance` checks and pickled state survive).
- Verifies legacy `make_att_2d_masks`, `pad_discrete_tokens`, `resize_with_pad`, and the SpaceTime SigLIP encoder helpers still resolve.

Policy-type strings in `factory.py` (`pi07_low_level`, `pi07_high_level`) are unchanged, so draccus `policy.type=pi07_low_level` configs keep working without edits.

### Out of scope

- The legacy `pi07_paligemma/low_level_planner/` package and its `PI07lowlevelPlannerConfig` / `PI07LowLevelPlannerPolicy` classes are intentionally untouched — that's a separate retired package.
[...]

PR #315, opened 2026-05-20
fix(pi07_paligemma): align low-level policy with pi07 low-level

### LINKED ISSUE #314: pi07 vs pi07_paligemma low-level: reconcile non-backbone inconsistencies (prefix layout, subgoal path, metadata, init, defaults)
## Summary

`pi07/low_level` (Gemma 3) and `pi07_paligemma/low_level_planner` (PaliGemma) are nominally the same low-level policy differing only in the VLM backbone, but a careful audit shows they have diverged well beyond the backbone. This issue tracks the concrete divergences to reconcile.

**Excluded by design:** the prefix *attention-masking* scheme is intentional and is NOT part of this issue. pi07 uses the π0.7-paper design (bidirectional video/state/subgoal blocks + causal text); pi07_paligemma uses the π0.5 prefix-LM design (mostly-bidirectional prefix). That difference is paper-driven for every token type (language, state, response, subgoal, metadata, separators), not a bug.

`pi07/low_level` is the canonical π0.7 (per CLAUDE.md, fixes target `pi07/`); `pi07_paligemma` is the legacy variant kept for older checkpoints. Reconciling toward `pi07/low_level` is the likely direction unless noted, but the per-item call is left to the maintainer.

Files:
- pi07: `src/opentau/policies/pi07/low_level/{configuration_pi07_low_level.py, modeling_pi07_low_level.py}`
- paligemma: `src/opentau/policies/pi07_paligemma/low_level_planner/{configuration_pi07_low_level.py, modeling_pi07_low_level.py}`

(Line refs below are `modeling…` unless prefixed with `config`.)

## A. Behavioural / prefix-layout divergences

These change the token stream the VLM sees, independent of the (excluded) attention masks.

- [ ] **Subgoal vs metadata block order is swapped.**
  - pi07: `… response → metadata → ";\n " → Subgoal: + images → ", " → [Action: + discrete]` — subgoal at the tail, per π0.7 Fig. 19 ("image goals come after the text prompt"). Docstring `:1280-1284`; response `:1490`; metadata `:1501`; subgoal `:1542`.
  - paligemma: `… response → Subgoal: + images → metadata → ":\n" → [Action: + discrete]` — subgoal in the middle. `_build_prefix_items :694-710`; response `:776`; subgoal `:786-817`; metadata `:830`.

- [ ] **Optional blocks: drop-when-absent (variable length) vs always-emit (fixed length).**
  - pi07 drops response/metadata/subgoal entirely when no sample globally has them, adapts separators (`", "`+`";\n "` vs `":\n"`), and OR-reduces the branch decisions across ranks via `_global_or_branch_decisions` for FSDP/ZeRO-3 safety. Helper `:71-157`; usage `:1383-1389`, `:1468`, `:1511`.
  - paligemma always emits every block at fixed length and gates per-sample purely with pad masks (`sample_has_response :766`, `sample_has_subgoal :787`, `sample_has_metadata :820`); no cross-rank sync.
  - The two should agree on one strategy. Overlaps #266 (pi07-side cleanup of the global-sync all-reduces).

- [ ] **Subgoal image embedding path differs.**
  - pi07 routes subgoal images through `gemma3_with_expert.embed_image()` as 4-D `(B,C,H,W)`, manually rescaled to `[-1,1]` (`prepare_subgoal_images :1047`; embed `:1576`).
  - paligemma routes them through `embed_video()` (SpaceTime SigLIP at T=1) as 5-D `(B,1,C,H,W)` left in `[0,1]` (`prepare_subgoal_images :1207`; video `ContextItem :809-817`; `_embed_item :1661-1680`).
  - Same subgoal image can produce different token counts / preprocessing across the two.

- [ ] **Padded-history video frames: pixel-zero + mask vs mask only.**
  - pi07 zeroes padded frames at the pixel level in `prepare_videos` (`:901-903`) AND passes `obs_history_is_pad` to the encoder.
  - paligemma relies only on the encoder's temporal-attention mask, no pixel zeroing (`:1071-1109`, docstring `:1074-1076`).

- [ ] (minor) **Response separator placement.** pi07 bakes the trailing `", "` into the response string (`prepare_response :975`); paligemma inserts `", "` as a separate gated item *before* the response (`:767-775`), so the comma lands on the other side of the response tokens.

## B. Feature / conditioning divergences

- [ ] **Episode-metadata fields.** pi07 supports `speed`, `quality`, `mistake`, **`robot_type`, `control_mode`** (`prepare_metadata :1100-1134`; "Robot:"/"Control:" at `:1130-1134`). paligemma supports only `speed`, `quality`, `mistake` (`:1419-1424`). The PaliGemma variant has no robot-type / control-mode conditioning.

- [ ] **Weight-init strategy.** paligemma has `init_strategy` (`no_init` / `full_he_init` / `expert_only_he_init`) with He-init and special `lerobot/pi05` handling (`config :151`; `_init_weights :1542`; `_init_model :1556-1567`). pi07 has no init-strategy concept.

## C. Config / API divergences

- [ ] **Backbone config surface.** pi07 bundles a nested `vlm_config: Gemma3WithExpertConfig` and plumbs `attention_implementation` / `gradient_checkpointing` into it in `__post_init__` (`config :193-201`, `:245-248`), exposing `--policy.vlm_config.*`. paligemma builds `PaliGemmaWithExpertConfig` inline from flat fields (`:1509-1518`); no nested config.

- [ ] **Default-value mismatches** (not backbone-driven):
  - `n_obs_steps`: pi07 `6` (`config :132`) vs paligemma `1` (`config :81`)
  - `dropout`: pi07 `0.1` (`config :178`) vs paligemma `0.0` (`config :141`)
  - `num_steps`: pi07 `5` (`config :181`) vs paligemma `10` (`config :144`)
  - (Backbone-driven, excluded: `resize_imgs_with_padding` 448 vs 224, `proj_width` 1280 vs 1024 — track the vision-tower image size and action-expert hidden size.)

- [ ] **Vestigial config field.** paligemma defines `discrete_action_indicator_max_length` (`config :118`) but never uses it; pi07 computes the indicator length dynamically (`_action_indicator_len :1229`). Drop the dead field.

- [ ] **Config class-name casing.** paligemma's `PI07lowlevelPlannerConfig` (`config :38`) vs pi07's `PI07LowLevelConfig` (`config :40`); `factory.py` aliases over it. Already flagged as the "class-name typo" in #249.

## Excluded (intentional / inherent)

- Prefix attention-masking design (π0.7 causal-text + bidirectional blocks vs π0.5 prefix-LM) — intentional, paper-driven.
- Backbone-inherent: tokenizer (`google/gemma-3-4b-pt` vs `google/paligemma-3b-pt-224`), engine class (`Gemma3WithExpertModel` vs `PaliGemmaWithExpertModel`), and the Gemma 3 "no manual sqrt-scaling on embeddings" vs PaliGemma `* sqrt(hidden)` scaling.

## Related

- #249 — broad policy dedup audit; its table item #4 covers these two files but the "V-JEPA2" note is now stale (paligemma is on SpaceTime SigLIP). This issue is the behavioural-parity counterpart. The `ContextItem` layout refactor in paligemma (`:57-100`) vs pi07's monolithic `embed_prefix` (`:1258`) is the structural side of the same split.
- #266 — pi07 `embed_prefix` global-sync cleanup (overlaps A2).
- #226 / #230 — byte-identity tests; a parity test across these two policies would pin these divergences going forward.
- #210 / #211 — pi07_paligemma video-encoder migration (now on SpaceTime SigLIP).

shuheng-liu:
## What this does
Fixes #314. Reconciles the legacy `pi07_paligemma` low-level policy toward the canonical `pi07` low-level so the two differ only by VLM backbone and the intentional (paper-driven) prefix attention masks.

Kept `pi07_paligemma`'s layout-agnostic `ContextItem` always-emit prefix design (DDP-safe by construction); deliberately did **not** port pi07's cross-rank `_global_or_branch_decisions` all-reduces, which overlaps #266 (that issue removes them from pi07).

- **Prefix order (A1):** subgoal block moved to the prefix tail, after metadata (π0.7 Fig. 19 — image goals follow the text prompt). Training (`forward`) and inference (`sample_actions`) both build the prefix through the single `_build_prefix_items`, so the token order is identical except for the training-only `Action:`/discrete tail (pinned by a new parity test).
- **Subgoal embedding (A3):** subgoal images embed via the image tower as 4-D `(B, C, H, W)` rescaled to `[-1, 1]` instead of `embed_video` at T=1. Byte-identical on the shared SigLIP tower; the new `image` branch runs unconditionally, removing a latent per-sample-data-dependent skip that could diverge across ranks.
- **Padded history (A4):** padded history frames are zeroed at the pixel level in `prepare_videos` (in addition to the encoder's temporal-attention mask).
- **Metadata (B1):** added `robot_type` / `control_mode` conditioning and matched pi07's metadata string format.
- **Init (B2):** removed `init_strategy` (pi07 has none); `load_pretrained_paligemma` is fixed to `False`.
- **Defaults (C2):** `n_obs_steps` 1→6, `dropout` 0.0→0.1 (`num_steps` kept at 10 — this default intentionally not reconciled to pi07's 5).
- **Cleanup (C3/C4):** dropped the vestigial `discrete_action_indicator_max_length` field; renamed `PI07lowlevelPlannerConfig` → `PI07PaligemmaLowLevelPlannerConfig`.

Out of scope: nested `vlm_config` refactor (C1) and the pi07-side #266 all-reduce cleanup.

## How it was tested
- Updated the CPU tests for the new layout (subgoal-at-tail offsets, metadata-before-subgoal, `embed_image` subgoal path, new metadata fields) and added `test_train_inference_prefix_item_order_parity` pinning train/inference prefix-order parity.
- `pytest -m "not gpu" tests/policies/test_pi07_paligemma_low_level_planner.py tests/policies/test_pi07_cpu.py` — 36 + 77 pass.
- `pytest -m gpu -n 0 tests/policies/test_pi07_paligemma_low_level_planner.py` on a CUDA box — passes (real PaliGemma backbone regression; the memory-heavy integration test stays skipped). Nightly regression runs via `regression_test.yml`.
- `pre-commit run --all-files` — clean.
- Verified `make_policy_config("pi07_paligemma_low_level_planner")` resolves to the renamed class with the new defaults, and `configs/examples/pi07_libero.json` still loads (it pins `n_obs_steps=1`, `num_steps=10`).

```bash
pytest -sx tests/policies/test_pi07_paligemma_low_level_planner.py
```
```bash
pytest -m gpu -n 0 tests/policies/test_pi07_paligemma_low_level_planner.py
```

## Checklist


shuheng-liu (commented):
## Review: aligning `pi07_paligemma` low-level with `pi07`

Well-scoped change. I checked the headline parity claims against the canonical `pi07/low_level/modeling_pi07_low_level.py` and they all hold. One behavioral side-effect on `pi07_libero.json` is worth confirming, plus a few low-severity parity nits. No correctness blockers found.

### Verified against the `pi07` reference
- **Prefix order (A1)** — `pi07` emits `... response → metadata → subgoal → (Action: discrete)` (`embed_prefix`: metadata at L1501, subgoal block at L1542). The new `_build_prefix_items` ordering (metadata L788-806, subgoal L808-842) matches.
- **Subgoal via `embed_image` (A3)** — `pi07` embeds subgoals as 4-D `(B,C,H,W)` through `gemma3_with_expert.embed_image` (L1576) and normalizes `* 2.0 - 1.0` (L1047). The PR's `image` branch (`_embed_item` L1675-1689) and `prepare_subgoal_images` do the same via `pi05.paligemma_with_expert.embed_image` (confirmed defined, `pi05/paligemma_with_expert.py:323`). Not scaling the image branch (unlike the `text` branch's `* sqrt(d)`) is correct and matches `pi07`.
- **Metadata format (B1)** — byte-for-byte match: `f"Speed: {str(speed.item())}, "` per segment, `' '.join(...)`, plus `Robot:`/`Control:` from `robot_type`/`control_mode` string lists with empty-string as the pad signal (`pi07` L1120-1136 vs PR L1460-1469).
- **Defaults (C2)** — `n_obs_steps=6`, `num_steps=5`, `dropout=0.1`, and `metadata_max_length=response_max_length=52` all match `pi07`'s config.
- **Padded history zeroing (A4)** — matches `pi07`'s `prepare_videos` (L901-903).
- **Clean rename/removal** — no orphaned refs to `PI07lowlevelPlannerConfig`, `init_strategy`, or `discrete_action_indicator_max_length` anywhere in the tree; removed `logging`/`Literal` imports are no longer used; `configs/examples/pi07_libero.json` still resolves via the unchanged registration key.

### Strengths
- **Routing subgoals through an unconditional `embed_image` is a genuine DDP correctness fix, not just a refactor.** Previously subgoals went through the `video` branch, whose `if per_sample_mask.any()` skip (`_embed_item` L1703) is a per-sample, data-dependent branch. Since `subgoal_is_pad` varies per sample/rank, that skip could diverge across ranks and fire mismatched vision-tower all-gathers under FSDP/ZeRO-3 — exactly the failure mode the repo guards against. The new branch runs unconditionally, so all ranks stay in lockstep. Good catch to make this layout-agnostic + always-emit rather than porting `_global_or_branch_decisions`.
- `test_train_inference_prefix_item_order_parity` is a solid guard against a one-sided block append silently desyncing train vs. inference.

### Things to confirm / suggestions
1. **`dropout` default flip silently changes `pi07_libero.json` (please confirm intended).** That config pins `n_obs_steps=1` and `num_steps=10` but does **not** set `dropout`, so it now inherits the new `0.1` default (was `0.0`, with a removed comment that explicitly kept it deterministic). The "How it was tested" note confirms the config still *loads* but doesn't mention it will now train *with* regularization it didn't have before. Almost certainly the intent given C2 is explicit — just calling it out since it's an unstated behavioral change to a checked-in example.
2. **Missing-camera subgoal placeholder fill (low / parity nit).** `prepare_subgoal_images` fills missing cameras with `torch.zeros_like(last_img)` (L1245). Since subgoals are now in `[-1, 1]`, `0.0` is mid-gray; `pi07` deliberately uses `-1` (black) for these slots. Functionally inert (mask is `False`, tokens are masked out), but `-1` would be exact parity.
3. **Unconditional `embed_image` adds per-step compute when subgoals are globally absent (low / perf).** `pi07` OR-reduces `has_subgoal` across ranks and skips the entire subgoal block when no rank has subgoals; the always-emit design here instead runs a SigLIP image-tower forward per subgoal camera every step over zero-filled images (e.g. for a subgoal-free LIBERO run). Correct and DDP-safe — just a bounded overhead worth being aware of vs. the reference.
4. **`prepare_videos` dropped `pi07`'s 4-D-input guard (low / robustness).** `pi07` raises if `vid.ndim == 4` while `n_obs_steps > 1` (L893-898). The port omits it (L1111-1112), and the new pixel-zeroing multiply (L1118) makes the gap sharper: a 4-D video (unsqueezed to `T=1`) multiplied by a `(B, T>1)` `obs_history_is_pad` would *silently broadcast-replicate* the single frame to `T` frames instead of erroring. Not reachable on current call paths (4-D only occurs at `n_obs_steps=1`, where the pad mask is `(B,1)` or absent), but re-adding the guard would keep parity and fail loud if that invariant ever breaks.

### Note (out of scope, not introduced here)
Response / metadata / `Subgoal:` blocks use `attention="bidirectional"` in `pi07_paligemma` whereas `pi07` makes those text spans causal (paper §VI.B). This is pre-existing and explicitly out of scope per the PR description ("intentional paper-driven prefix attention masks") — flagging only so the "differ only by VLM backbone" framing is read with that caveat.

### Tests / CI
Test updates are internally consistent (subgoal-at-tail offsets, `embed_image` mock token count 6→4, metadata-before-subgoal slices). CI is still `pending` (no checks reported yet on the head SHA) — worth a second look once `cpu_test` / `pre-commit` report.

---

shuheng-liu:
Thanks for the thorough review. Addressed in `8e63288`:

**Fixed**
- **#4 (4-D-input guard):** re-added `prepare_videos`' guard that raises when `vid.ndim == 4 and n_obs_steps > 1`, matching pi07 — closes the silent broadcast-replicate gap against the new pixel-zeroing.
- **#2 (missing-camera placeholder):** missing-camera slots now fill `-1` (black at the `[-1, 1]` SigLIP range), matching pi07. I deliberately kept the *no-subgoal-keys* fabrication at `0` (not `-1`): that path is asserted byte-identical to the per-sample-padded path (`img * 0`) in `test_subgoal_tensors_all_is_pad_matches_missing_keys`, so `0` preserves train/inference byte-parity there. Both fills are mask-inert, and the present-but-padded vs missing-camera split now matches pi07 exactly (`0` via multiply / `-1` placeholder).

**Confirmed intended (no change)**
- **#1 (dropout 0.0→0.1 inherited by `pi07_libero.json`):** intended — the flip to match pi07 is explicit and the example doesn't pin `dropout`. Agreed it's worth calling out.
- **#3 (unconditional `embed_image` compute):** intended trade-off of keeping the always-emit design instead of porting `_global_or_branch_decisions`. Bounded overhead (one image-tower forward per subgoal camera over masked-out slots) in exchange for DDP-safety without cross-rank all-reduces (the #266 direction).

**Out-of-scope note (bidirectional vs causal text spans):** acknowledged — pre-existing and paper-driven, not touched here.

**Also in this push:** renamed `pi07_paligemma_low_level_planner` → `pi07_paligemma_low_level` (directory, registered type/name, and the `PI07PaligemmaLowLevel{Config,Policy,FlowMatching}` classes) — the low-level component is a controller, not a planner, mirroring pi07 (#250).

Re-verified: GPU pytest (`pytest -m gpu tests/policies/test_pi07_paligemma_low_level.py`) green on the renamed branch; CPU suite (113 tests) + `pre-commit` clean.

shuheng-liu:
Two follow-ups from maintainer feedback (`5f4d2df`):

- **`num_steps` kept at 10** (reverted the 10→5 default flip). This default is intentionally *not* reconciled to pi07's 5.
- **Reverted the missing-camera `-1` subgoal fill back to `0`** (re #2 above). On reflection it's purely cosmetic: those slots carry `pad_mask=False`, so they're excluded as attention keys (`make_att_2d_masks`) and the vision-tower output never reaches any real token — the fill value can't affect the model. Not worth the churn for byte-parity on a masked-out tensor. The 4-D-input guard (#4) stays.

shuheng-liu (commented):
## Re-review of the update (`8e63288` + `5f4d2df`)

Thanks for the detailed responses. I re-checked the new commits against the tree and the fixes all landed correctly:

- **#4 (4-D guard)** — confirmed present (`prepare_videos`, raises on `vid.ndim == 4 and n_obs_steps > 1` before the pixel-zeroing multiply). Closes the silent broadcast-replicate gap. ✅
- **#2 (subgoal fill)** — at HEAD both paths are `0` (no-subgoal-keys fabrication and missing-camera placeholder), and both are mask-inert. This matches your second comment (the brief `-1` from `8e63288` was reverted in `5f4d2df`) — final state is self-consistent and well-documented. Agree it's not worth churn on a masked-out tensor. ✅
- **`num_steps` → 10** — applied; ASCII diagram updated back to `x 10` too. ✅
- **Rename `low_level_planner` → `low_level`** — clean and complete: directory, registered type/name, and `PI07PaligemmaLowLevel{Config,Policy,FlowMatching}` classes. No dangling references to the old module path, old key, or any intermediate class name anywhere in the tree; `factory.py`, `pi07_libero.json`, and both test files (incl. the `test_pi07_cpu.py` parametrize path) are all updated consistently. ✅

### One thing to confirm — registration-key break
Renaming the draccus key `pi07_paligemma_low_level_planner` → `pi07_paligemma_low_level` is a **hard** backward-incompatibility: any saved `config.json` / checkpoint whose `policy.type` is the old key will now fail to *deserialize* (no registered subclass), not just behave differently. CLAUDE.md positions `pi07_paligemma` as the variant "kept for older checkpoints," so worth an explicit confirmation that no released or in-flight checkpoint depends on the old key.

That said — the prefix reorder / subgoal-embed / metadata changes already broke *semantic* compatibility with pre-PR checkpoints (they'd load by state-dict but run on a different prefix layout). Given that, a hard load-failure is arguably the safer fail-fast outcome (it stops someone silently running an old checkpoint through incompatible new code) rather than a regression. So this is really a "please confirm the clean break is intended, and ideally note it in the PR description / a changelog" item, not a request to add a back-compat alias.

### Minor (doc only)
With `num_steps` intentionally left at 10, the description's "the two differ only by VLM backbone and the intentional prefix attention masks" now also excludes the `num_steps` default — the C2 bullet already says so, but the one-line summary up top could add "and a couple of intentionally-unreconciled defaults" to match.

No blockers from my side. CI is still `pending` (no checks reported on the new head SHA yet) — worth confirming `cpu_test` / `pre-commit` go green before un-drafting.

---

shuheng-liu:
The renaming is intentional. The 5 vs 10 diffusion steps is assumed when using paligemma.

PR #316, opened 2026-05-21
feat(train): log LIBERO grid-summary eval videos to wandb

shuheng-liu:
## What this does
When LIBERO eval videos are generated during training, log the per-task `grid_summary.mp4` files to wandb via `accelerator.log(...)` when wandb is enabled. Individual `eval_episode_*.mp4` clips are intentionally skipped.

- New helper `collect_grid_summary_videos(videos_dir)` in `src/opentau/scripts/eval.py` globs the per-task `grid_summary.mp4` files, strips the trailing `_rank{N}` from the task subdir name (so wandb keys stay stable across GPU counts), and excludes individual episode clips.
- `src/opentau/scripts/train.py` logs the grid videos under `Eval Videos/{suite}_{task_id}` in the main-process eval block, guarded by `cfg.wandb.enable and cfg.eval.max_episodes_rendered > 0`.
- Bumps the `wandb` floor `0.16.3 → 0.27.0` (relock). **Intentional** — a deliberate modernization to standardize on a current wandb instead of the ~2-year-old floor; not strictly required by `wandb.Video` (which has accepted mp4 paths for many releases), bundled here as the change that touches wandb.

🗃️ Feature

## How it was tested
- Added `test_collect_grid_summary_videos_*` in `tests/scripts/test_eval.py` (CPU): verifies rank-stripping, sorted ordering, exclusion of episode clips, and the missing-dir case. `pytest -m "not gpu" tests/scripts/test_eval.py` → 2 passed.
- Verified the logging path itself: ran the exact `accelerator.log({f"Eval Videos/{name}": wandb.Video(path)}, step=step)` code against a generated `grid_summary.mp4` (plus a dummy episode clip) with wandb 0.27.0 + accelerate, in offline mode — both grid videos were staged under `media/videos/Eval Videos/...` and the clip was not logged.
- `pre-commit run --all-files` passes. `uv sync` verified on macOS and Linux with the bumped wandb.
- Not a policy change; the LIBERO sim / GPU eval path is unchanged, so GPU and regression tests were not run.

```bash
```
The logging fires automatically during a training run with eval + rendering enabled; grid videos then appear in wandb under `Eval Videos/*`:
```bash
opentau-train --config_path=configs/examples/pi05_libero_eval_config.json --wandb.enable=true
```

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -68,7 +68,7 @@ dependencies = [
     "torch>=2.7.1",
     "torchcodec>=0.4.0, <0.5.0; sys_platform != 'win32' and (sys_platform != 'linux' or (platform_machine != 'aarch64' and platform_machine != 'arm64' and platform_machine != 'armv7l')) and (sys_platform != 'darwin' or platform_machine != 'x86_64')",
     "torchvision>=0.22.1",
-    "wandb>=0.16.3",
+    "wandb>=0.27.0",
     "zarr>=2.17.0",
     "scikit-learn>=1.7.1",
     "onnx>=1.18.0",
--- a/src/opentau/scripts/eval.py
+++ b/src/opentau/scripts/eval.py
@@ -24,6 +24,7 @@ import concurrent.futures as cf
 import datetime as dt
 import json
 import logging
+import re
 import threading
 import time
 from collections import defaultdict
@@ -887,6 +888,23 @@ def consolidate_eval_info(eval_infos: list[dict]) -> dict:
         "per_group": per_groups,
         "overall": overall,
     }
+
+
+def collect_grid_summary_videos(videos_dir: Path) -> list[tuple[str, str]]:
+    """Find per-task grid_summary.mp4 files under ``videos_dir`` for wandb logging.
+
+    Returns sorted ``(task_name, path)`` pairs, where ``task_name`` is the
+    per-task subdir name with any trailing ``_rank{N}`` stripped (so wandb keys
+    stay stable across GPU-count changes). Individual ``eval_episode_*.mp4``
+    clips are intentionally excluded.
+    """
+    if not videos_dir.exists():
+        return []
+    results = [
+        (re.sub(r"_rank\d+$", "", grid_path.parent.name), str(grid_path))
+        for grid_path in videos_dir.rglob("grid_summary.mp4")
+    ]
+    return sorted(results)
 
 
 def main():
--- a/src/opentau/scripts/eval.py
+++ b/src/opentau/scripts/eval.py
@@ -894,9 +894,16 @@ def collect_grid_summary_videos(videos_dir: Path) -> list[tuple[str, str]]:
     """Find per-task grid_summary.mp4 files under ``videos_dir`` for wandb logging.
 
     Returns sorted ``(task_name, path)`` pairs, where ``task_name`` is the
-    per-task subdir name with any trailing ``_rank{N}`` stripped (so wandb keys
-    stay stable across GPU-count changes). Individual ``eval_episode_*.mp4``
-    clips are intentionally excluded.
+    per-task subdir name (``{task_group}_{task_id}_rank{N}``) with any trailing
+    ``_rank{N}`` stripped, so wandb keys stay stable across GPU-count changes.
+    Individual ``eval_episode_*.mp4`` clips are intentionally excluded.
+
+    Assumes each ``(task_group, task_id)`` is evaluated on exactly one rank, which
+    the disjoint round-robin task sharding guarantees (``idx % num_processes ==
+    process_index`` in ``envs/configs.py``). Under that assumption the stripped
+    key is unique; if the same task were ever evaluated on multiple ranks, the
+    keys would collide and the later video would overwrite the earlier one at the
+    same step.
     """
     if not videos_dir.exists():
         return []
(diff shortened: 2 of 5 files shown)


shuheng-liu (commented):
## Review: feat(train) — log LIBERO grid-summary eval videos to wandb

### Overview
Adds `collect_grid_summary_videos(videos_dir)` in `eval.py` (globs per-task `grid_summary.mp4`, strips trailing `_rank{N}` for stable keys, excludes episode clips) and logs each grid video under `Eval Videos/{task}` in the main-process eval block of `train.py`, guarded by `cfg.wandb.enable and cfg.eval.max_episodes_rendered > 0`. Also bumps the `wandb` floor `0.16.3 → 0.27.0`. Small, focused, and well-tested change.

### What's solid
- **No race condition.** The collection runs after `gather_object(...)` (train.py:783), which is a collective that synchronizes all ranks; every rank has finished `eval_policy_all` — and therefore written its `grid_summary.mp4` — before the main process globs the shared dir. Good.
- **Exclusion is by construction.** `rglob("grid_summary.mp4")` only matches grid files, so episode clips can never be picked up regardless of the test's `# must be excluded` clip. Robust.
- **Guard matches the data.** `max_episodes_rendered > 0` is exactly the condition under which any video exists, so no empty-glob churn.
- **Tests + docstrings present**, consistent with the checklist. Logging style (separate `accelerator.log(..., step=step)` per key) matches the existing eval-metric logging right above it.

### Points to consider
1. **wandb floor bump seems broader than the feature needs (question).** `wandb.Video` with an mp4 path has worked for many releases well before 0.27.0, and the lock already resolved 0.24.1. Raising a hard-dependency floor by 11 minor versions can force-upgrade downstream/user environments — is there a specific 0.27.0 behavior you're relying on? If not, a more conservative floor (or none) would reduce friction. If there *is* a reason (e.g. a video-logging fix), worth a one-line note in the PR body.

2. **Rank-stripping can collide if the same `(task_group, task_id)` is ever evaluated on more than one rank.** Subdirs are `{group}_{id}_rank{local_process_index}` (eval.py:704). Two such dirs that differ only in rank suffix both strip to the same `Eval Videos/{group}_{id}` key and are logged at the same `step`, so the second silently overwrites the first. With distinct-task sharding across ranks this won't trigger, and note the suffix uses `local_process_index` (not global) — so on multi-node, node0-local0 and node1-local0 share a suffix. Today's single-node default is fine; worth a comment documenting the "one rank per task" assumption, or de-duping keys, so a future multi-node/sharding change doesn't lose videos quietly.

3. **Shared-filesystem assumption.** Collection runs only on the main process and globs `videos_dir`, so it relies on all ranks writing to a filesystem the main process can read. True for the documented single-node 8-GPU setup; on multi-node without a shared FS, non-rank-0 videos won't be visible. Acceptable given current usage — flagging so it's a conscious assumption.

4. **Minor:** the per-video `accelerator.log` calls could be batched into one dict `{f"Eval Videos/{name}": wandb.Video(p) for ...}` for a single log at `step`. Not required — it matches the surrounding style — but slightly cleaner.

### Verdict
Correct and low-risk for the current single-node eval path. No blocking issues. The only thing I'd genuinely want answered before merge is **(1)** — justify or relax the wandb floor bump. **(2)** is a cheap future-proofing comment worth adding.

---

shuheng-liu:
Thanks for the review. Addressed in 2d3e42e:

1. **wandb floor bump** — intentional (deliberate modernization to standardize on a current wandb; not strictly required by `wandb.Video`). Added a note to the PR description.
2. **Rank-stripping collision** — documented the one-rank-per-task assumption in the `collect_grid_summary_videos` docstring: the disjoint round-robin task sharding (`idx % num_processes == process_index`) guarantees each `(task_group, task_id)` runs on exactly one rank, so the stripped key is unique; noted that putting the same task on multiple ranks would collide/overwrite at the same step.
3. **Shared-FS assumption** — documented at the log site (main-process-only glob of the shared `videos_dir`; non-rank-0 videos on a multi-node run without a shared FS won't be seen).
4. **Batching** — the per-task videos are now logged in a single `accelerator.log({...}, step=step)` call.

shuheng-liu (commented):
## Re-review of 2d3e42e — all prior points resolved

Re-reviewed after the follow-up commit. No blocking issues; this is good to go from my side.

- **(4) Batching** ✅ — per-task videos now go out in a single `accelerator.log(grid_videos, step=step)`, guarded by `if grid_videos`. Cleaner and avoids an empty log; behavior is unchanged.
- **(2) Rank-stripping collision** ✅ — the docstring now documents the one-rank-per-task assumption. I verified the cited guarantee: `envs/configs.py:268-269` shards via `idx % accelerator.num_processes == accelerator.process_index`, so each `(task_group, task_id)` lands on exactly one rank and the stripped key is unique. Worth noting the guarantee is even a bit stronger than the docstring implies — sharding is by *global* `process_index` while the dir suffix uses `local_process_index`, but since the `{task_group}_{task_id}` prefix is preserved and unique per task, keys can't collide even on multi-node. The documented caveat (same task on multiple ranks ⇒ overwrite) is the correct boundary.
- **(3) Shared-FS / main-process glob** ✅ — clearly noted at the log site.
- **(1) wandb floor bump** ✅ (resolved by explanation) — accepting the "deliberate modernization, not strictly required by `wandb.Video`" rationale; PR description updated. Your call as maintainer.

Tests are unaffected (`collect_grid_summary_videos` signature/return unchanged). Nothing further from me.

---

PR #317, opened 2026-05-22
fix(pi07_paligemma): decouple n_action_steps from chunk_size

shuheng-liu:
## What this does

`n_action_steps` is meant to be the inference-time **execution horizon** (how many actions from each predicted chunk are executed before the policy re-queries with fresh observations), while `chunk_size` is the **prediction horizon** the flow-matching model is trained to decode. The `pi07_paligemma_low_level` policy conflated the two: both the training loss and the inference denoise loop sliced the action-expert output to `n_action_steps`, so any `n_action_steps < chunk_size` would:

- crash inference — the Euler step `x_t += dt * v_t` broadcast `v_t` (length `n_action_steps`) against `x_t` (length `chunk_size`); and
- mis-shape the training MSE target (`v_t` vs `u_t = noise - actions`, which is length `chunk_size`).

That effectively forced `n_action_steps == chunk_size`, so the policy could only ever execute the *entire* predicted chunk open-loop before re-planning.

Changes (in `policies/pi07_paligemma/low_level/`):
- `modeling_pi07_low_level.py`: decode/supervise the full `chunk_size` chunk in the training forward and the denoise step (slice by `chunk_size`, not `n_action_steps`); in `select_action`, push only the first `n_action_steps` actions onto the queue so it drains and re-queries after `n_action_steps` steps (receding horizon).
- `configuration_pi07_low_level.py`: guard that a shortened execution horizon (`n_action_steps < chunk_size`) requires `max_delay == 0` (the real-time-delay prefix path is not yet compatible), and clarify the `chunk_size` / `n_action_steps` docstrings.

This is a **no-op for existing configs** where `n_action_steps == chunk_size` (the slices are identical and the queue fill is unchanged), so trained checkpoints and in-flight training runs are unaffected.

🐛 Bug

## How it was tested

- Added CPU regression tests (`tests/policies/test_pi07_paligemma_low_level.py::TestPI07PaligemmaLowLevelExecutionHorizon`) that run on the `cpu_test.yml` merge gate:
  - the new `__post_init__` guard rejects `n_action_steps < chunk_size` with `max_delay > 0`, and allows the valid combinations;
  - `select_action` with `n_action_steps < chunk_size` executes only the first `n_action_steps` of the decoded chunk and re-queries afterwards (the path that tripped the queue-length assert before the fix).
- `pytest -m "not gpu"` passes (1194 + the new tests; the only failures locally are pre-existing LIBERO-asset / stdin-capture env issues unrelated to this change).
- `pytest -m "gpu" tests/policies/test_pi07_paligemma_low_level.py` is green. Note: the heavy forward + `select_action` pipeline test is `@gpu`/`@slow` and currently skipped for memory, and it only ever ran at `n_action_steps == chunk_size` — so it does **not** by itself guard the shortened-horizon path; the CPU tests above do.
- No-op verification: when `n_action_steps == chunk_size` the edited slices are identical and the queue fill is unchanged, so existing checkpoints and in-flight training are unaffected.

```bash
pytest -sx tests/policies/test_pi07_paligemma_low_level.py
```

```bash
# Eval a trained pi07_paligemma_low_level checkpoint executing only the first 10 of the
# 30 predicted actions before re-planning (previously crashed; now works):
opentau-eval --config_path=<checkpoint>/train_config.json \
  --policy.n_action_steps=10 --eval.batch_size=4
```

## Checklist

### DIFF
--- a/src/opentau/policies/pi07_paligemma/low_level/configuration_pi07_low_level.py
+++ b/src/opentau/policies/pi07_paligemma/low_level/configuration_pi07_low_level.py
@@ -43,8 +43,11 @@ class PI07PaligemmaLowLevelConfig(PreTrainedConfig):
         n_obs_steps: Number of observation steps to use. Defaults to 6.
         history_interval: Temporal stride between the ``n_obs_steps`` stacked
             frames in the inference buffer, in environment steps. Defaults to 1.
-        chunk_size: Size of the action chunk. The upper bound for n_action_steps. Defaults to 50.
-        n_action_steps: Number of action steps to predict. Defaults to 50.
+        chunk_size: Trained action-chunk length, i.e. the prediction horizon the model
+            always decodes at inference. Upper bound for n_action_steps. Defaults to 50.
+        n_action_steps: Inference execution horizon -- how many actions from each
+            predicted chunk are executed before the policy re-queries with fresh
+            observations. Must be <= chunk_size. Defaults to 50.
         normalization_mapping: Mapping of feature names to normalization modes.
             Defaults to identity for visual features and mean-std for state and action.
         max_state_dim: Maximum dimension for state vectors. Shorter vectors are padded. Defaults to 32.
@@ -196,6 +199,15 @@ class PI07PaligemmaLowLevelConfig(PreTrainedConfig):
         if self.max_delay > self.chunk_size:
             raise ValueError(
                 f"The max delay must be less than or equal to the chunk size. Got {self.max_delay} for `max_delay` and {self.chunk_size} for `chunk_size`."
+            )
+
+        if self.n_action_steps < self.chunk_size and self.max_delay != 0:
+            raise ValueError(
+                "A shortened execution horizon (n_action_steps < chunk_size) is not yet "
+                "supported together with real-time inference delay (max_delay > 0); they "
+                "would entangle the action-queue prefix logic. Got "
+                f"n_action_steps={self.n_action_steps}, chunk_size={self.chunk_size}, "
+                f"max_delay={self.max_delay}."
             )
 
     def validate_features(self) -> None:
--- a/src/opentau/policies/pi07_paligemma/low_level/modeling_pi07_low_level.py
+++ b/src/opentau/policies/pi07_paligemma/low_level/modeling_pi07_low_level.py
@@ -1998,6 +1998,12 @@ class PI07PaligemmaLowLevelFlowMatching(nn.Module):
 
         Factored out so that :meth:`forward`, :meth:`sample_actions`, and
         :meth:`denoise_step` share one definition of what's in the suffix.
+
+        Invariant: the action block must stay the *trailing* ``chunk_size``
+        tokens of the suffix. :meth:`forward` and :meth:`denoise_step` recover
+        the action outputs with ``suffix_out[:, -chunk_size:]``; if a future
+        block is appended after the actions, that slice (and those call sites)
+        must be updated accordingly.
         """
         bsize = x_t.shape[0]
         action_mask = torch.ones(bsize, x_t.shape[1], dtype=torch.bool, device=x_t.device)
--- a/src/opentau/policies/pi07_paligemma/low_level/modeling_pi07_low_level.py
+++ b/src/opentau/policies/pi07_paligemma/low_level/modeling_pi07_low_level.py
@@ -660,7 +660,12 @@ class PI07PaligemmaLowLevelPolicy(PreTrainedPolicy):
             delay = torch.tensor(delay, dtype=torch.long, device=batch["state"].device)
             actions = self.sample_actions(batch, noise=noise, action_prefix=action_prefix, delay=delay)
             actions = rearrange(actions, "b c d -> c b d")
-            self._action_queue.extend(actions[delay:])
+            # Execute only the first n_action_steps of the predicted chunk, then
+            # re-query with fresh observations (receding horizon). The config guard
+            # (n_action_steps < chunk_size => max_delay == 0 => delay == 0) keeps this
+            # slice in range and exactly n_action_steps long; when n_action_steps ==
+            # chunk_size it clamps to actions[delay:] (unchanged behaviour).
+            self._action_queue.extend(actions[delay : delay + self.config.n_action_steps])
             assert len(self._action_queue) == self.config.n_action_steps, (
                 f"Action queue must have {self.config.n_action_steps} actions"
             )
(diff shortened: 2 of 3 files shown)


shuheng-liu:
## Review: decouple `n_action_steps` from `chunk_size`

Reviewed the diff plus surrounding context in `pi07_paligemma/low_level/`. The fix is correct and internally coherent — the three decode/supervise slices, the `select_action` queue slice, and the new config guard all line up, and I verified it's a genuine no-op when `n_action_steps == chunk_size`. Two things I'd want addressed before merge, plus a couple of nits.

### Correctness (verified)
- **Training forward** (`modeling_pi07_low_level.py:1943`): `u_t` (`:1907`) and `postfix_mask` (`:1897`) are both `chunk_size`-length, so slicing `v_t` to `chunk_size` aligns all three. Pre-fix, `n_action_steps < chunk_size` would broadcast-fail in `F.mse_loss`.
- **`denoise_step`** (`:2128`): `x_t` is `chunk_size` (`:2040`/`:2061`), so `v_t` at `chunk_size` matches the Euler update `x_t += dt * v_t` (`:2076`).
- **`select_action`** (`:668`): after rearrange `actions` has length `chunk_size`. `actions[delay : delay + n_action_steps]` clamps to `actions[delay:]` when `n_action_steps == chunk_size` (unchanged); and with the new guard (`n_action_steps < chunk_size` ⟹ `max_delay == 0` ⟹ regeneration only on an empty queue ⟹ `delay == 0`) the slice is exactly `n_action_steps` long, so the `maxlen=n_action_steps` deque assert holds. The guard is the keystone that keeps this in range — good that it's there.

### Should address before merge

**1. No regression test for the path being fixed.** The diff touches zero test files. The only `select_action` test (`test_pi07_paligemma_low_level.py::test_complete_pi07_low_level_pipeline`) runs with `n_action_steps=CHUNK_SIZE` (`:155`) — i.e. exactly the no-op case the fix leaves unchanged. So the actual bug (`n_action_steps < chunk_size`) and the new `__post_init__` guard have **no coverage**, and that test is `@gpu`-marked so it never runs in `cpu_test.yml` (the merge gate). Suggest a cheap CPU test that (a) asserts the guard raises for `n_action_steps < chunk_size and max_delay > 0`, and (b) exercises the training forward / `select_action` shapes with `n_action_steps < chunk_size`. Otherwise a future refactor can silently reintroduce the broadcast crash with CI staying green.

**2. The identical bug still lives in the primary `pi07/low_level` impl.** Per CLAUDE.md, `pi07/` (Gemma3) is the current π0.7 and `pi07_paligemma/` is the legacy variant. The same conflation is present verbatim in `src/opentau/policies/pi07/low_level/modeling_pi07_low_level.py` — `actions[delay:]` queue extend (`:638`), `suffix_out[:, -n_action_steps:]` in the training forward (`:1845`) and denoise step (`:2045`), the same `x_t += dt * v_t` (`:1993`) — and its config lacks the new guard. So `n_action_steps < chunk_size` is still broken on the more-used policy. Scoping this PR to paligemma is fine, but let's at least track a follow-up (or fold it in) so the primary path doesn't stay crashy.

### Nits
- **`suffix_out[:, -chunk_size:]` assumes the action block is the suffix tail.** True today (`_build_suffix_items` returns only the action item), but `embed_suffix`'s docstring notes the suffix "may add more in the future" — if anything is appended after the action block, `-chunk_size:` would grab the wrong tokens. Pre-existing assumption, not a regression, but since the PR edits these exact lines, a short comment/assert anchoring the action offset would be cheap insurance.
- **"How it was tested" slightly overstates coverage.** The GPU test covers the forward and `select_action` paths only at `n_action_steps == chunk_size` — the path the fix doesn't change — so it doesn't guard the regression. Worth correcting so a future reader doesn't assume otherwise.

Overall: the logic is sound and safe to land for existing configs (`n_action_steps == chunk_size`); the blockers are test coverage for the shortened-horizon path and the parallel bug in `pi07/`.

---

shuheng-liu:
Thanks for the review — addressed in d0a5947.

**1. Regression test (CPU, on the merge gate).** Added `TestPI07PaligemmaLowLevelExecutionHorizon` in `tests/policies/test_pi07_paligemma_low_level.py`:
- the `__post_init__` guard raises for `n_action_steps < chunk_size` with `max_delay > 0` (and allows the valid combinations);
- `select_action` with `n_action_steps < chunk_size` executes only the first `n_action_steps` of the decoded chunk, then re-queries — driven through a stubbed `sample_actions`, so it needs no GPU/model. It fails on the pre-fix code (the `len == n_action_steps` queue assert trips), so it actually guards the regression.

The forward/denoise broadcast paths aren't cheaply unit-testable on CPU (they need the full PaliGemma model — that's why the pipeline test is `@gpu` and skipped for memory). The guard makes the only reachable `n_action_steps < chunk_size` config `max_delay == 0`, and the three edited slices now uniformly use `chunk_size` (matching `u_t` / `x_t`).

**2. Parallel bug in `pi07/low_level`.** Good catch — filed #318 to track it across `pi0` / `pi05` / `pi05_mem` / `pi06` / `pi07`. Keeping this PR scoped to `pi07_paligemma`; the primary `pi07/` path will be handled in a follow-up per that issue.

**Nits.**
- Anchored the suffix-tail assumption in the `_build_suffix_items` docstring (the `-chunk_size:` slices rely on the action block staying the trailing suffix block).
- Corrected "How it was tested" — the GPU pipeline test only covers `n_action_steps == chunk_size` and is skipped for memory, so it does not guard the shortened-horizon path; the new CPU tests do.

shuheng-liu:
## Re-review of d0a5947

Pulled the update and went through the new commit; all three points from the last pass are addressed, and I ran the new tests locally to confirm.

**1. Regression test — verified.** `TestPI07PaligemmaLowLevelExecutionHorizon` is CPU-only (no `@gpu` marker), so it runs on the `cpu_test.yml` gate. I ran it: `4 passed`. I also reverted just the `select_action` slice back to `actions[delay:]` and re-ran — it fails, so the test genuinely guards the regression.

One correction to the reply: it fails via the **value/order assertion** (`[1007.0, 1008.0, 1009.0] != [1000.0, 1001.0, 1002.0]` at `:2250`), *not* the `len == n_action_steps` queue assert. The deque's `maxlen=n_action_steps` silently truncates the over-long `extend(actions[delay:])` down to the last 3 elements, so `len` stays 3 and that assert never trips. Same conclusion (the test does catch it), just a different mechanism than described. Worth knowing because it also means the production `assert len(...) == n_action_steps` at `:669` can only ever catch *under*-filling, never over-filling — `maxlen` defangs it. Pre-existing, not introduced here.

**2. Parallel `pi07/` bug — tracked.** #318 covers all five flow-matching policies and even calls out the `pi0` `safety_buffer` / `extend(actions)` nuance. Scoping this PR to `pi07_paligemma` is fine.

**3. Nits — done.** The suffix-tail invariant is now documented in `_build_suffix_items`, and "How it was tested" accurately reflects that the GPU pipeline test only exercises `n_action_steps == chunk_size`.

Minor: the diff touches `modeling_*.py`, which per CLAUDE.md §3 nominally wants a determinism check — but since the change is provably bit-identical for existing `n_action_steps == chunk_size` configs (established last pass), in-flight runs are unaffected by construction, so a fresh seeded-smoke run isn't really needed here.

Bottom line: correct and well-tested; good to merge once CI is green.

---

PR #320, opened 2026-05-22
fix(policies): decouple n_action_steps from chunk_size

### LINKED ISSUE #318: Decouple n_action_steps (execution horizon) from chunk_size in remaining flow-matching policies
The latent bug fixed for `pi07_paligemma_low_level` in #317 also exists in the other flow-matching policies: `n_action_steps` (intended as the inference **execution horizon**) is conflated with `chunk_size` (the trained **prediction horizon**), so `n_action_steps < chunk_size` is broken.

In each policy the action-expert output is sliced to `n_action_steps` in both the training forward and the inference denoise step, and `select_action` fills the queue with `actions[delay:]`. When `n_action_steps < chunk_size` this:

- crashes inference — the flow-matching Euler step `x_t += dt * v_t` mismatches `v_t` (length `n_action_steps`) against `x_t` (length `chunk_size`); and
- mis-shapes the training MSE target.

So today these policies can only execute the entire predicted chunk before re-planning.

### Affected files
Same `suffix_out[:, -self.config.n_action_steps :]` pattern (training forward + denoise step) plus `_action_queue.extend(actions[delay:])`:

- `src/opentau/policies/pi0/modeling_pi0.py`
- `src/opentau/policies/pi05/modeling_pi05.py`
- `src/opentau/policies/pi05_mem/modeling_pi05.py`
- `src/opentau/policies/pi06/modeling_pi06.py`
- `src/opentau/policies/pi07/low_level/modeling_pi07_low_level.py`

### Fix (mirror #317)
1. Slice the action-expert output by `chunk_size` (not `n_action_steps`) in the training forward and the denoise step.
2. In `select_action`, push only the first `n_action_steps` actions onto the queue (receding horizon).
3. Add the config guard that `n_action_steps < chunk_size` requires `max_delay == 0`.

Each change is a no-op when `n_action_steps == chunk_size` (the current configs), and each policy needs its GPU pytests + a determinism check.

Note: `pi0` uses `safety_buffer` and `extend(actions)` (no `delay`), so its `select_action` change differs slightly from the others.

shuheng-liu:
## What this does

Fixes #318. Mirrors #317 across the **remaining** flow-matching policies.

`n_action_steps` is the inference-time **execution horizon** (how many actions from each predicted chunk are executed before the policy re-queries with fresh observations); `chunk_size` is the **prediction horizon** the flow-matching model is trained to decode. #317 fixed `pi07_paligemma_low_level`, where the two were conflated. The same latent bug existed in the other flow-matching policies — `pi0`, `pi05`, `pi05_mem`, `pi06`, and `pi07/low_level` — where both the training loss and the inference denoise loop sliced the action-expert output to `n_action_steps`. Any `n_action_steps < chunk_size` would:

- crash inference — the Euler step `x_t += dt * v_t` broadcasts `v_t` (length `n_action_steps`) against `x_t` (length `chunk_size`); and
- mis-shape the training MSE target (`v_t` vs `u_t = noise - actions`, which is length `chunk_size`).

That forced `n_action_steps == chunk_size`, so these policies could only execute the *entire* predicted chunk open-loop before re-planning.

Changes in each of `pi0`, `pi05`, `pi05_mem`, `pi06`, `pi07/low_level`:
- `modeling_*.py`: decode/supervise the full `chunk_size` chunk in the training forward and the denoise step (slice by `chunk_size`, not `n_action_steps`); in `select_action`, push only the first `n_action_steps` actions onto the queue so it drains and re-queries after `n_action_steps` steps (receding horizon).
- `configuration_*.py`: guard that a shortened execution horizon (`n_action_steps < chunk_size`) requires `max_delay == 0` (the real-time-delay prefix path is not yet compatible), and clarify the `chunk_size` / `n_action_steps` docstrings.

`pi0` is the documented exception: it has no `max_delay`/delay-prefix path — it refills via `safety_buffer` and `extend(actions)`. So its `select_action` slices `actions[:n_action_steps]` (the action queue's `maxlen` is `n_action_steps`, so extending with the whole chunk would silently drop the leading actions and execute the wrong tail), and its guard is `n_action_steps < chunk_size` requires `safety_buffer == 0`.

This is a **no-op for existing configs** where `n_action_steps == chunk_size` — every config under `configs/` uses this, so the edited slices are byte-identical, the queue fill is unchanged, and no guard fires. Trained checkpoints and in-flight training runs are unaffected.

🐛 Bug

## How it was tested

- Added CPU regression tests — a `Test*ExecutionHorizon` class in each of `tests/policies/test_pi0.py`, `test_pi05.py`, `test_pi05_mem.py`, `test_pi06.py`, `test_pi07_low_level.py` (20 tests total). Each covers:
  - the new `__post_init__` guard rejects the shortened-horizon + delay combo (`safety_buffer` for `pi0`, `max_delay` for the rest) and allows the valid combinations;
  - `select_action` with `n_action_steps < chunk_size` decodes the full `chunk_size` chunk but executes only the first `n_action_steps` and re-queries afterwards (the path that crashed / tripped the queue-length assert before the fix).
- `pytest -m "not gpu"` on the five touched test files: **118 passed** (incl. the 20 new tests). Full-suite collection succeeds; the only collection errors are the pre-existing LIBERO / `robosuite` import errors, unrelated to this change.
- pre-commit (ruff lint + format, bandit, gitleaks, pyupgrade, …) clean on all touched files.
- No-op verification: every config under `configs/` has `n_action_steps == chunk_size`, so the edited slices are identical and the queue fill is unchanged.
- GPU pytests (`pytest -m "gpu"`) + regression tests are **pending** — being run separately before this leaves draft. The shortened-horizon path is guarded by the new CPU tests above (the heavy GPU pipeline tests only ever ran at `n_action_steps == chunk_size`), exactly as #317 documented.

```bash
pytest -sx tests/policies/test_pi0.py tests/policies/test_pi05.py tests/policies/test_pi05_mem.py tests/policies/test_pi06.py tests/policies/test_pi07_low_level.py
```

```bash
# Eval a trained checkpoint executing only the first 10 of the 50 predicted
# actions before re-planning (previously crashed; now works):
opentau-eval --config_path=<checkpoint>/train_config.json \
  --policy.n_action_steps=10 --eval.batch_size=4
```

## Checklist

> Note: this **is** a policy change. The policy / GPU-test boxes are intentionally left unchecked while in draft — the GPU pytests + regression tests haven't been run yet. Please check both once `pytest -m "gpu"` and the regression tests pass.

### DIFF
--- a/src/opentau/policies/pi0/configuration_pi0.py
+++ b/src/opentau/policies/pi0/configuration_pi0.py
@@ -43,8 +43,11 @@ class PI0Config(PreTrainedConfig):
 
     Args:
         n_obs_steps: Number of observation steps to use. Defaults to 1.
-        chunk_size: Size of the action chunk. The upper bound for n_action_steps. Defaults to 50.
-        n_action_steps: Number of action steps to predict. Defaults to 50.
+        chunk_size: Trained action-chunk length, i.e. the prediction horizon the model
+            always decodes at inference. Upper bound for n_action_steps. Defaults to 50.
+        n_action_steps: Inference execution horizon -- how many actions from each
+            predicted chunk are executed before the policy re-queries with fresh
+            observations. Must be <= chunk_size. Defaults to 50.
         safety_buffer: Safety buffer size. Defaults to 0.
         normalization_mapping: Mapping of feature names to normalization modes.
             Defaults to identity for visual features and mean-std for state and action.
@@ -175,6 +178,15 @@ class PI0Config(PreTrainedConfig):
         if self.n_obs_steps != 1:
             raise ValueError(
                 f"Multiple observation steps not handled yet. Got `nobs_steps={self.n_obs_steps}`"
+            )
+
+        if self.n_action_steps < self.chunk_size and self.safety_buffer != 0:
+            raise ValueError(
+                "A shortened execution horizon (n_action_steps < chunk_size) is not yet "
+                "supported together with a non-zero safety_buffer; they would entangle the "
+                "action-queue refill logic. Got "
+                f"n_action_steps={self.n_action_steps}, chunk_size={self.chunk_size}, "
+                f"safety_buffer={self.safety_buffer}."
             )
 
     def validate_features(self) -> None:
--- a/src/opentau/policies/pi0/modeling_pi0.py
+++ b/src/opentau/policies/pi0/modeling_pi0.py
@@ -771,8 +771,12 @@ class PI0FlowMatching(nn.Module):
         action_time_mask = torch.ones(bsize, action_time_dim, dtype=torch.bool, device=device)
         pad_masks.append(action_time_mask)
 
-        # Set attention masks so that image, language and state inputs do not attend to action tokens
-        att_masks += [1] + ([0] * (self.config.n_action_steps - 1))
+        # Set attention masks so that image, language and state inputs do not attend to action tokens.
+        # The action block spans the full chunk_size (= the noise/x_t length, both in the training
+        # forward and inference). n_action_steps is the execution horizon applied later in
+        # select_action, not the number of action tokens; using it here would mismatch the
+        # chunk_size-length pad mask and crash make_att_2d_masks when n_action_steps < chunk_size.
+        att_masks += [1] + ([0] * (self.config.chunk_size - 1))
 
         embs = torch.cat(embs, dim=1)
         pad_masks = torch.cat(pad_masks, dim=1)
@@ -873,7 +877,11 @@ class PI0FlowMatching(nn.Module):
         device = state.device
 
         if noise is None:
-            actions_shape = (bsize, self.config.n_action_steps, self.config.max_action_dim)
+            # Decode the full trained chunk (chunk_size); select_action executes only the
+            # first n_action_steps (receding horizon). This must be chunk_size so the action
+            # tokens, the embed_suffix attention mask, and the denoise_step output slice all
+            # align -- otherwise v_t (chunk_size) mismatches x_t (n_action_steps) in the Euler step.
+            actions_shape = (bsize, self.config.chunk_size, self.config.max_action_dim)
             noise = self.sample_noise(actions_shape, device)
 
         prefix_embs, prefix_pad_masks, prefix_att_masks = self.embed_prefix(
--- a/src/opentau/policies/pi0/modeling_pi0.py
+++ b/src/opentau/policies/pi0/modeling_pi0.py
@@ -368,7 +368,12 @@ class PI0Policy(PreTrainedPolicy):
         if len(self._action_queue) <= self.config.safety_buffer:
             actions = self.sample_actions(batch, noise=noise)
             actions = rearrange(actions, "b c d -> c b d")
-            self._action_queue.extend(actions)
+            # sample_actions decodes the full chunk_size chunk, but only the first
+            # n_action_steps form the execution horizon (receding horizon). The queue's
+            # maxlen is n_action_steps, so extending with the whole chunk would silently
+            # drop the leading actions and execute the wrong tail; slice explicitly. When
+            # n_action_steps == chunk_size this is the whole chunk (unchanged behaviour).
+            self._action_queue.extend(actions[: self.config.n_action_steps])
         return self._action_queue.popleft()
 
     @torch.no_grad()
@@ -831,7 +836,10 @@ class PI0FlowMatching(nn.Module):
             use_cache=False,
             fill_kv_cache=False,
         )
-        suffix_out = suffix_out[:, -self.config.n_action_steps :]
+        # Supervise the whole chunk the model was trained to predict. n_action_steps
+        # is the inference-time execution horizon only and must not truncate the
+        # training target (chunk_size is the prediction horizon).
+        suffix_out = suffix_out[:, -self.config.chunk_size :]
         # Original openpi code, upcast attention output
         v_t = self.action_out_proj(suffix_out)
         v_t = v_t.to(dtype=torch.float32)
@@ -947,7 +955,10 @@ class PI0FlowMatching(nn.Module):
             fill_kv_cache=False,
         )
         suffix_out = outputs_embeds[1]
-        suffix_out = suffix_out[:, -self.config.n_action_steps :]
+        # Denoise the full chunk_size chunk so v_t matches x_t in the Euler step.
+        # n_action_steps (execution horizon) is applied later in select_action, not
+        # at decode time.
+        suffix_out = suffix_out[:, -self.config.chunk_size :]
         v_t = self.action_out_proj(suffix_out)
         v_t = v_t.to(dtype=torch.float32)
         return v_t
--- a/src/opentau/policies/pi05/configuration_pi05.py
+++ b/src/opentau/policies/pi05/configuration_pi05.py
@@ -43,8 +43,11 @@ class PI05Config(PreTrainedConfig):
 
     Args:
         n_obs_steps: Number of observation steps to use. Defaults to 1.
-        chunk_size: Size of the action chunk. The upper bound for n_action_steps. Defaults to 50.
-        n_action_steps: Number of action steps to predict. Defaults to 50.
+        chunk_size: Trained action-chunk length, i.e. the prediction horizon the model
+            always decodes at inference. Upper bound for n_action_steps. Defaults to 50.
+        n_action_steps: Inference execution horizon -- how many actions from each
+            predicted chunk are executed before the policy re-queries with fresh
+            observations. Must be <= chunk_size. Defaults to 50.
         normalization_mapping: Mapping of feature names to normalization modes.
             Defaults to identity for visual features and mean-std for state and action.
         max_state_dim: Maximum dimension for state vectors. Shorter vectors are padded. Defaults to 32.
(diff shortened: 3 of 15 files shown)


shuheng-liu:
## Review: decouple `n_action_steps` from `chunk_size`

**TL;DR:** the change is a genuine no-op for existing configs (`n_action_steps == chunk_size`), so it's safe for current checkpoints and in-flight runs. But as written it does **not** actually enable `n_action_steps < chunk_size` for **pi0, pi05, or pi06** — a real (non-mocked) run still crashes — and the new tests don't catch this because they monkeypatch `sample_actions`. **pi05_mem** and **pi07/low_level** are correct. Recommend addressing the items below before removing draft.

### Critical

**1. pi05 & pi06: `embed_suffix` still hardcodes `n_action_steps` in the attention mask → real path crashes for `n_action_steps < chunk_size`.**

The PR moves the *output* slice in `forward`/`denoise_step` to `chunk_size`, but `embed_suffix` still builds the suffix attention mask as:

```python
att_masks += [1] + ([0] * (self.config.n_action_steps - 1))   # pi05 embed_suffix; identical in pi06
```

The action **pad** mask and embeddings, however, span the full `chunk_size` — the noise / `x_t` is `(B, chunk_size, …)` (see `sample_actions` and `forward`). So for `n_action_steps < chunk_size`, `embed_suffix` returns `suffix_att_masks` of length `n_action_steps` and `suffix_pad_masks` of length `chunk_size`, and `make_att_2d_masks` then does:

```
att_2d : (B, n_action_steps, n_action_steps)
pad_2d : (B, chunk_size,     chunk_size)
att_2d & pad_2d   ->  RuntimeError  (e.g. "size of tensor a (3) must match tensor b (10)")
```

This fires in **both** `forward` (training) and `denoise_step` (inference) — upstream of the Euler step, so it's actually the *first* crash, not the broadcast described in the PR body. The sibling policies already do this correctly:

- `pi05_mem` `embed_suffix`: `att_masks += [1] + ([0] * (self.config.chunk_size - 1))`
- `pi07/low_level` `embed_suffix`: same, `chunk_size`

**Fix:** in pi05 and pi06 `embed_suffix`, change `(self.config.n_action_steps - 1)` → `(self.config.chunk_size - 1)`.

**2. pi0: the denoise change is incoherent with pi0's inference noise shape; it breaks the very case it claims to enable.**

Unlike the others, pi0 builds its **inference** noise at `n_action_steps` (unchanged by this PR):

```python
actions_shape = (bsize, self.config.n_action_steps, self.config.max_action_dim)   # pi0 sample_actions
```

and its `embed_suffix` mask is also `n_action_steps` — so pi0 inference is internally consistent at `n_action_steps`. The PR changes `denoise_step` to `suffix_out[:, -self.config.chunk_size :]`, which slices the last `chunk_size` tokens of a suffix that only has `n_action_steps` action tokens, so `v_t` is now *longer* than `x_t` (length `n_action_steps`). The Euler step `x_t += dt * v_t` then raises for `n_action_steps < chunk_size`.

Consequently the `select_action` comment is inaccurate:

```python
# sample_actions decodes the full chunk_size chunk, but only the first n_action_steps ...
self._action_queue.extend(actions[: self.config.n_action_steps])
```

`sample_actions` returns `n_action_steps` actions (the noise is `n_action_steps`), so `actions[: n_action_steps]` is a no-op. To genuinely support a short horizon in pi0, the inference `actions_shape` and the `embed_suffix` mask also need to move to `chunk_size` (the training-`forward` slice change is fine on its own, since training `actions` are already `chunk_size`). As written, the pi0 portion doesn't deliver the feature and changes the denoise output shape in a way that breaks `sample_actions`.

**3. The new tests don't exercise the path that breaks.**

Each `test_select_action_executes_first_n_then_requeries` does `policy.sample_actions = fake_sample_actions`, so the real `embed_suffix → make_att_2d_masks → denoise_step → Euler` path is never run at `n_action_steps < chunk_size`. That's why all 118 pass despite (1)–(2). The planned GPU/regression tests won't catch it either — per the PR body they only ever run at `n_action_steps == chunk_size`. Suggest adding, for at least pi0/pi05/pi06, a test that calls the real `sample_actions` on a tiny CPU-constructable model (or simply asserts `embed_suffix`'s returned mask and pad lengths are equal) with `n_action_steps < chunk_size`.

### Looks correct

- **pi05_mem** and **pi07/low_level**: end-to-end coherent — noise is `chunk_size`, `embed_suffix` mask uses `chunk_size`, `forward`/`denoise_step` slice `chunk_size`, and `select_action` slices `[delay : delay + n_action_steps]`. The receding-horizon logic and the `delay == 0 when max_delay == 0` guard hold (verified in `select_action`).
- **No-op for existing configs**: at `n_action_steps == chunk_size` every changed slice is byte-identical and the queue fill is unchanged, so existing checkpoints / in-flight runs are unaffected, as claimed.
- Config guards (`max_delay` / `safety_buffer`) and the docstring clarifications are reasonable.

### Minor

- The PR body's failure diagnosis ("crash at the Euler step `x_t += dt * v_t`") is only accurate for pi0; for pi05/pi06 the first crash is upstream in `embed_suffix` / `make_att_2d_masks`.

---

shuheng-liu:
Thanks — this caught a real, important gap. You're right on all three critical points; the first commit only moved the **output** slices and missed that `embed_suffix` builds the action **attention mask** at `n_action_steps`, so the real (non-mocked) path still crashed for `n_action_steps < chunk_size`. Fixed in d6a260d:

**1. pi05 & pi06 `embed_suffix`** — the action att-mask block now spans `chunk_size`:
```python
att_masks += [1] + ([0] * (self.config.chunk_size - 1))
```
This is exactly what `pi05_mem` / `pi07/low_level` already did. The `chunk_size`-length pad mask and att mask now agree, so `make_att_2d_masks` no longer raises.

**2. pi0** — two changes, since (as you noted) pi0 was internally consistent at `n_action_steps`:
- `embed_suffix`: action att-mask block → `chunk_size`;
- `sample_actions`: inference `actions_shape` → `(bsize, chunk_size, max_action_dim)` (was `n_action_steps`).

Now pi0 decodes the full `chunk_size` chunk, the suffix mask matches it, `denoise_step`'s `[-chunk_size:]` slice lines up with `x_t` in the Euler step, and `select_action`'s `actions[:n_action_steps]` is a genuine receding-horizon slice rather than a no-op.

**3. Tests** — added `test_embed_suffix_attn_mask_spans_full_chunk` for pi0/pi05/pi06 that calls the real `embed_suffix` (stubbing only the projections / sinusoidal embedding, not `sample_actions`) at `n_action_steps=3, chunk_size=10` and asserts `pad_masks.shape[1] == att_masks.shape[1]`. Verified red→green: it fails `assert 10 == 3` against the old mask and passes after the fix. The earlier mocked `select_action` tests are kept (they still cover the queue/receding-horizon logic).

**Verified `pi05_mem` and `pi07/low_level` need no change** — both already build the suffix att mask at `chunk_size`, sample inference noise at `chunk_size`, and slice outputs at `chunk_size`, so they're end-to-end coherent (matching your "Looks correct" note).

**Minor** — the inline code comments now state the accurate failure mode (`embed_suffix` / `make_att_2d_masks` length mismatch upstream of the Euler step for pi05/pi06; the Euler-step mismatch was the pi0-specific symptom).

`pytest -m "not gpu"` on the five policy test files is green (121 passed). Still a no-op at `n_action_steps == chunk_size`. GPU pytests + regression tests still to be run before this leaves draft.

shuheng-liu:
## Re-review of `d6a260d` — all three findings resolved

Verified against the diff and by tracing tensor shapes through `embed_suffix` / `sample_actions` (not just the reply):

- **pi05 & pi06** `embed_suffix` now build the action block at `chunk_size` (`[1] + [0] * (chunk_size - 1)`), so `att_masks` and the `chunk_size`-length `pad_masks` agree and `make_att_2d_masks` no longer mismatches — in both `forward` and `denoise_step`. ✅
- **pi0** gets both required changes: `embed_suffix` mask → `chunk_size`, and inference `actions_shape` → `(bsize, chunk_size, max_action_dim)`. The chain is now coherent: noise (`chunk_size`) → suffix mask (`1 + chunk_size`) → `denoise_step` `[-chunk_size:]` (the action tokens, excluding the state token) → `x_t += dt * v_t` aligns → `select_action`'s `actions[: n_action_steps]` is a genuine receding-horizon slice. ✅
- **Tests**: the new `test_embed_suffix_attn_mask_spans_full_chunk` (pi0/pi05/pi06) calls the **real** `embed_suffix` (only the sinusoidal embedding + projections are stubbed) and asserts `pad_masks.shape[1] == att_masks.shape[1]` — this fails `10 == 3` against the old mask and passes now, so it's a real red→green guard, not another `sample_actions` mock. ✅

Also confirmed the change stays a **no-op at `n_action_steps == chunk_size`** (the `chunk_size` vs `n_action_steps` expressions are identical there), so existing checkpoints/runs are unaffected, and that **pi05_mem / pi07/low_level** correctly needed no change.

From a CPU-correctness standpoint this now does what the PR title says, for all five policies. (I verified statically + by shape-tracing; I didn't spin up a torch env to run the suite in this session — the reported 118 → 121 is consistent with the 3 added tests.)

### One non-blocking follow-up (out of this PR's scope)

The ONNX paths still size noise by `n_action_steps`, not `chunk_size`:
- `src/opentau/scripts/export_to_onnx.py:162` — `noise_shape = (1, policy.config.n_action_steps, max_action_dim)` (and `action_prefix` likewise)
- `src/opentau/scripts/onnx_inference.py` (`n_action_steps` arg → noise shape)

Since the config now *allows* `n_action_steps < chunk_size`, a shortened-horizon model would feed `n_action_steps`-length noise into a `sample_actions`/`embed_suffix` that now expects `chunk_size`, so export would hit the same length-mismatch. Never usable before either, so fine to leave as a separate follow-up — just flagging that the decoupling isn't complete on the export path (or the guard could be extended to reject `n_action_steps < chunk_size` for export until then).

---

PR #321, opened 2026-05-22
feat(pi05): add FlexAttention backend for block-causal attention

WilliamYue37:
## What this does

Adds `attention_implementation="flex"` to `PaliGemmaWithExpertModel` (the shared backbone reused by `pi05` and both submodules of `pi07_paligemma` — `high_level_planner/` and `low_level/`). The new path routes through `torch.nn.attention.flex_attention`, which JIT-compiles a fused Triton kernel specialized to the existing block-causal mask. From the caller's perspective the three backends (`eager` / `sdpa` / `flex`) are interchangeable.

Implementation notes:
- Mask is consumed as the same dense `(B, S_q, S_kv)` bool tensor `make_att_2d_masks` already produces — no API changes upstream. The tensor is wrapped in a `BlockMask` via a closed-over `mask_mod`, so block-sparse skipping kicks in on the prefix-bidirectional + causal-tail pattern.
- Ampere-friendly Triton tile sizes (`BLOCK_M=BLOCK_N=BLOCK_M{1,2}=BLOCK_N{1,2}=32`, `num_stages=1`) are pinned via `kernel_options` when `head_dim >= 192` so the kernel fits in 100 KB/SM shared memory on RTX 3090 / 4090. Defaults (Hopper-sized) OOM on Ampere at head_dim=256.
- Short / non-multiple-of-64 Q_LENs are right-padded (to the smallest length ≥ 128 that's a multiple of 64) — PyTorch 2.10's flex Inductor lowering asserts below that on Ampere. Padded Q rows attend to KV[0] only (keeps softmax finite) and are sliced off after the kernel. Production pi07_paligemma sequence lengths are O(1024), so the hot path is a no-op cast.

⚡️ Performance.

## How it was tested

Numerical parity (`tests/policies/test_pi05_paligemma_with_expert_flex.py`):
- Forward parity vs. SDPA in bf16 and fp32 across 7 shape variants (pi07-like GQA with KV prefix, equal Q/KV length, smaller head_dim path, non-multiple-of-64 Q_LEN, sub-128 Q_LEN that fully exercises the pad path) — max-abs-err < 5e-3 in bf16, < 1e-4 in fp32.
- Backward parity vs. SDPA in bf16: dQ within 1e-2, dK/dV within 5e-2 (one bf16 ULP through the 8x GQA reduction, matches flash-attn upstream's envelope).
- CPU dispatcher / config-roundtrip tests round out coverage.

Microbenchmark on 2x RTX 3090 (`src/opentau/scripts/bench_paligemma_attention.py`), fwd+bwd, bf16, head_dim=256, GQA 8/1:

| Shape | eager (ms) | sdpa (ms) | flex (ms) | flex vs eager | eager peak | flex peak | mem saved |
|---|---|---|---|---|---|---|---|
| B=2, S_q=1024, S_kv=1280 | 4.35 | 5.04 | 3.14 | **1.38x** | 448 MB | 160 MB | **64%** |
| B=4, S_q=1024, S_kv=1280 | 8.91 | 9.57 | 4.60 | **1.94x** | 877 MB | 303 MB | **65%** |
| B=2, S_q=2048, S_kv=2304 | 14.99 | 17.33 | 6.87 | **2.18x** | 1320 MB | 400 MB | **70%** |

SDPA is *slower* than eager at head_dim=256 on Ampere — PyTorch SDPA doesn't dispatch to FA2 at this head dim, so it falls back to the memory-efficient kernel which loses to eager's tensor-core matmul. Flex avoids this by compiling its own kernel.

End-to-end on 8x A100 (slurm `bot_387`, DeepSpeed ZeRO-2 bf16, `pi07_paligemma_low_level`, batch_size=32, `gradient_checkpointing=true`, 100 steady-state training steps, only `attention_implementation` differs):

| Backend | ms/step (steady) | Peak GiB / rank | vs. eager (time) | vs. eager (mem) |
|---|---|---|---|---|
| eager | 10650 | 75.15 | 1.00x | 1.00x |
| sdpa  | 9725  | 71.66 | **1.10x** | **0.95x** |
| flex  | 9737  | 71.66 | **1.09x** | **0.95x** |

On A100, SDPA *does* dispatch to FA2 at head_dim=256 (unlike Ampere consumer cards), so SDPA and flex run essentially the same fused kernel — they tie on both step time and peak memory. Both deliver the same ~9% wall-clock win and ~5% memory reduction vs. eager (smaller than the 3090 numbers above because `gradient_checkpointing=true` recomputes attention activations in the backward pass, hiding most of the per-layer activation savings). Loss values match across all three backends at every logged step (bf16 reassociation noise only).

DeepSpeed ZeRO-2 smoke (`src/opentau/scripts/smoke_paligemma_attention_deepspeed.py`) on 2x 3090 with `configs/examples/accelerate_deepspeed_config.yaml`:

```
eager: loss 5.0049 -> 4.6749   (5 steps)
sdpa : loss 5.0060 -> 4.6786   (5 steps)
flex : loss 5.0071 -> 4.6851   (5 steps)
```

All three backends produce within-noise loss curves, both ranks agree, no NCCL collective mismatches.

Pre-commit hooks pass; CPU tests for `test_pi0.py`, `test_pi05.py`, `test_pi07_paligemma_low_level.py`, `test_pi07_paligemma_high_level_planner.py`, and the new flex test all pass (21 + 40 + 8 items).

```bash
git fetch origin feat/flex-attention-pi07-paligemma
git checkout feat/flex-attention-pi07-paligemma

# Numerical parity (needs a CUDA box; CPU subset of the file still runs locally)
pytest -sx tests/policies/test_pi05_paligemma_with_expert_flex.py

# Microbenchmark vs. eager / sdpa
python -m opentau.scripts.bench_paligemma_attention \
    --backends eager sdpa flex \
    --batch-size 2 --s-q 1024 --s-kv 1280 --prefix-kv 256

# DeepSpeed ZeRO-2 smoke on 2 GPUs (set --num_processes to match)
accelerate launch --config_file configs/examples/accelerate_deepspeed_config.yaml \
    --num_processes 2 \
    src/opentau/scripts/smoke_paligemma_attention_deepspeed.py \
    --attention-implementation flex
```

To use in a real training config, set `policy.attention_implementation: "flex"` (for any policy that uses the shared pi05 backbone — pi05, pi07_paligemma).

## Checklist

GPU pytests were run on the RTX 5090 box. Results:

- **The 6 new GPU parity tests in `tests/policies/test_pi05_paligemma_with_expert_flex.py` pass cleanly** (1.97s).
- The full sequential `pytest -m "gpu" -n 0` run on a single 32 GB card hit 5 failures, but they are a **preexisting test-isolation issue, not introduced by this PR**: 4 of them (`test_pi05_loc_tokens_in_response_produce_finite_loss`, 3x `test_pi05_mem_gpu.py::*`) OOM / `CUDNN_STATUS_INTERNAL_ERROR` because the pi05_mem tests load PaliGemma-base weights and don't free them between sequential tests. I checked out `main` on the same box and confirmed those 7 tests **pass in isolation** (61 s, 7/7). The 5th failure (`test_flex_matches_sdpa_fp32[256]`) is also a `CUDA driver error: out of memory` — same fragmentation, not a numerical regression. The nightly CI runs on g6.12xlarge (4x L4) where pytest distribution sidesteps this.
- Regression tests (nightly cron) haven't been triggered against this branch yet; happy to run before merge if you'd like.

### DIFF
--- a/src/opentau/policies/pi05/paligemma_with_expert.py
+++ b/src/opentau/policies/pi05/paligemma_with_expert.py
@@ -26,6 +26,11 @@ import logging
 
 import torch
 from torch import nn
+from torch.nn.attention.flex_attention import (
+    BlockMask,
+    create_block_mask,
+    flex_attention,
+)
 from transformers import (
     AutoConfig,
     Cache,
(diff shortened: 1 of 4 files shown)


WilliamYue37:
Closing per the bench results on the team's actual training hardware.

Re-ran flex on 8x A100 (slurm `bot_387`, DeepSpeed ZeRO-2 bf16, `pi07_paligemma_low_level`, batch_size=32, `gradient_checkpointing=true`, 100 steady-state steps) after the padding fix (`d5c5ddf`):

| Backend | Commit | ms/step | Peak GiB/rank |
|---|---|---|---|
| sdpa | dd9c7bb | 9725 | **71.66** |
| flex | dd9c7bb (pre-pad) | 9738 | 71.66 |
| flex | d5c5ddf (post-pad) | 9412 | **72.40** |

On A100, flex ends up **0.74 GiB worse than SDPA on peak memory** (+1.0%) and the 3% step-time delta is well within run-to-run noise. The reason is that A100's SDPA dispatches to FA2 at head_dim=256, so SDPA and flex run essentially the same fused kernel — flex carries no advantage there.

The 1.4–2.1x speedup and 64–70% memory savings on RTX 3090 (microbenchmarks in the description) come specifically from Ampere consumer cards' SDPA *not* hitting FA2 at head_dim=256, which isn't representative of the team's training hardware.

Since the team trains on A100 and the criterion is a significant memory improvement vs the current default (SDPA), the value isn't there. Closing.

If we ever want to revisit this for inference on consumer GPUs, the implementation in this branch is ready to pick up.

PR #322, opened 2026-05-22
ci(gpu): point LIBERO at bundled package assets in nightly test

### LINKED ISSUE #319: Nightly GPU test failure: test_control_freq_reaches_real_robosuite_sim — missing LIBERO init_files in CI
## Summary

The nightly GPU pytest job failed on `main` in [run 26287432829](https://github.com/TensorAuto/OpenTau/actions/runs/26287432829/job/77378523795).

A single test fails:

```
FAILED tests/envs/test_libero_control_freq.py::test_control_freq_reaches_real_robosuite_sim
  - FileNotFoundError: [Errno 2] No such file or directory:
    '/tmp/libero-assets/libero/libero/init_files/libero_10/LIVING_ROOM_SCENE2_put_both_the_alphabet_soup_and_the_tomato_sauce_in_the_basket.pruned_init'

= 1 failed, 18 passed, 10 skipped, 1227 deselected, 5 warnings in 404.68s =
```

## Root cause

The test (added in #312) builds a real `OffScreenRenderEnv` via `create_libero_envs(task="libero_10", ...)` and `init_states=True` (the default), which calls `torch.load` on a `.pruned_init` file under `LIBERO_CONFIG_PATH/init_files/libero_10/`.

The `gpu_test.yml` workflow only creates the empty parent directory and never populates it:

```yaml
- name: Run Tests
  shell: bash
  run: |
    source .venv/bin/activate
    mkdir -p /tmp/libero-assets/libero/libero
    export LIBERO_CONFIG_PATH="$(pwd)/.github/assets/libero"
    pytest -m "gpu" -n 0 -v tests/
```

The captured stdout in the failing test confirms the assets directory is empty:

```
[Warning]: assets path /tmp/libero-assets/libero/libero/./assets does not exist!
[Warning]: bddl_files path /tmp/libero-assets/libero/libero/./bddl_files does not exist!
[Warning]: datasets path /tmp/libero-assets/libero/libero/../datasets does not exist!
[Warning]: init_states path /tmp/libero-assets/libero/libero/./init_files does not exist!
```

The mocked tests in the same file (`test_control_freq_passed_through`, `test_control_freq_defaults_to_20`) pass — they patch `OffScreenRenderEnv` and `get_libero_path` and don't touch real assets.

## Fix options

1. Populate `/tmp/libero-assets/libero/libero` with the real LIBERO assets in `gpu_test.yml` (or point `LIBERO_CONFIG_PATH` to a checkout of the forked LIBERO repo so `init_files/`, `bddl_files/`, `assets/` resolve).
2. Have the test set `init_states=False` (and any other flags needed) so the real-sim path only exercises what the test actually asserts (that `control_freq=13` reaches the underlying robosuite env).
3. Skip the test when the LIBERO assets aren't present, with a clear `pytest.skip(...)` message.

Option 2 seems cleanest — the test only reads back `vec.envs[0]._env.env.control_freq`, so init states aren't needed for the assertion.

## Pointers

- Test: `tests/envs/test_libero_control_freq.py:92`
- Code path that needs the file: `src/opentau/envs/libero.py:79` (`_get_task_init_states` → `torch.load`)
- Workflow: `.github/workflows/gpu_test.yml`
- Introduced in: #312 (`feat(envs): make LIBERO sim control frequency configurable`, commit 9aa5c10)

shuheng-liu:
## What this does

Fixes #319.

The nightly GPU test `test_control_freq_reaches_real_robosuite_sim` (added in #312) builds a **real** `OffScreenRenderEnv` to verify the configured `control_freq` actually reaches robosuite — the mocked sibling tests can't catch robosuite silently dropping the kwarg. Building a real env needs LIBERO's `bddl_files`/`init_files`/`assets`, but `gpu_test.yml` set `LIBERO_CONFIG_PATH` to an empty `/tmp` tree, so the test failed with:

```
FileNotFoundError: .../init_files/libero_10/LIVING_ROOM_SCENE2_..._in_the_basket.pruned_init
```

The empty-tree config is correct for the mocked CPU tests (they patch `get_libero_path`), but the real-sim GPU test needs actual assets. Note that just setting `init_states=False` in the test is **not** sufficient on its own: the env build in `src/opentau/envs/libero.py` reads the bddl file via `get_libero_path("bddl_files")`, so it would then fail on the missing bddl file instead. The workflow fix is the necessary part.

Changes:
- **`.github/workflows/gpu_test.yml`**: point `LIBERO_CONFIG_PATH` at the assets bundled inside the installed LIBERO package (the fork's `MANIFEST.in` does `graft libero`, shipping `bddl_files`/`init_files`/`assets`) via LIBERO's own `set_libero_default_path()`, instead of the empty `/tmp` tree. As a bonus this also drops a pre-existing `SC2155` shellcheck warning on the old `export VAR="$(pwd)/..."` line.
- **`tests/envs/test_libero_control_freq.py`**: set `init_states=False` to focus the test on its `control_freq` assertion (matching the two mocked sibling tests); the env build still exercises the real bddl/asset path.

## How it was tested

- `actionlint` on the edited workflow: the `Run Tests` step is shellcheck-clean, and the change removes a pre-existing `SC2155` warning.
- Verified the LIBERO fork bundles the exact failing `init_files/libero_10/*.pruned_init` (~23 KB real file, not an LFS pointer) and the `bddl_files/libero_10/*.bddl` for the task under test, and that `MANIFEST.in` grafts them into the installed package.
- Could **not** run the GPU test locally (no NVIDIA GPU; LIBERO not installed). Validation is the nightly job — `gpu_test.yml` was manually dispatched on this branch via `workflow_dispatch`; see the linked run in the checks.

```bash
# Dispatch the nightly GPU workflow on this branch:
gh workflow run gpu_test.yml --ref claude/magical-banach-eb58a8
```
```bash
# On a CUDA machine with the libero extra installed, run just the fixed test:
pytest -sx tests/envs/test_libero_control_freq.py::test_control_freq_reaches_real_robosuite_sim
```

## Checklist

### DIFF
--- a/.github/workflows/gpu_test.yml
+++ b/.github/workflows/gpu_test.yml
@@ -89,8 +89,16 @@ jobs:
         shell: bash
         run: |
           source .venv/bin/activate
-          mkdir -p /tmp/libero-assets/libero/libero
-          export LIBERO_CONFIG_PATH="$(pwd)/.github/assets/libero"
+          # Point LIBERO at the assets bundled inside the installed package (the
+          # fork's MANIFEST grafts bddl_files/init_files/assets) so the real-sim
+          # GPU test can build a genuine OffScreenRenderEnv. The .github/assets/libero
+          # config used by the mocked CPU tests points every path at an empty /tmp
+          # tree, which is fine for mocks but not for a real env build (it reads the
+          # bddl file via get_libero_path("bddl_files")).
+          export LIBERO_CONFIG_PATH=/tmp/libero-cfg
+          mkdir -p "$LIBERO_CONFIG_PATH"
+          touch "$LIBERO_CONFIG_PATH/config.yaml"  # pre-create so the LIBERO import skips its first-run input() prompt
+          python -c "from libero.libero import set_libero_default_path; set_libero_default_path()"
           pytest -m "gpu" -n 0 -v tests/
 
   stop-runner:
--- a/tests/envs/test_libero_control_freq.py
+++ b/tests/envs/test_libero_control_freq.py
@@ -105,6 +105,7 @@ def test_control_freq_reaches_real_robosuite_sim():
         n_envs=1,
         gym_kwargs={"task_ids": {"libero_10": [0]}, "control_freq": 13},
         env_cls=gym.vector.SyncVectorEnv,
+        init_states=False,
     )
     vec = envs["libero_10"][0]
     try:


shuheng-liu (commented):
**Review: the LIBERO-assets CI fix is correct and empirically validated.**

I traced the workflow incantation end-to-end against the pinned LIBERO fork (`45c5a01`) and the mechanics hold:

- `touch "$LIBERO_CONFIG_PATH/config.yaml"` — `libero/libero/__init__.py` only prompts on `if not os.path.exists(config_file)`. Pre-creating the file skips the interactive first-run `input()`, which would otherwise hit EOF and crash the import under CI. The empty file is harmless because nothing reads the config at import time (`get_libero_path` is lazy).
- `set_libero_default_path()` — reads `LIBERO_CONFIG_PATH` from the env and writes package-relative paths (`benchmark_root` = the installed `libero/libero/` dir) into that same `config.yaml`, overwriting the empty file. The subsequent `pytest` inherits the exported `LIBERO_CONFIG_PATH`, so `get_libero_path("bddl_files")` resolves into the installed package.
- The assets actually land in the install: `MANIFEST.in` does `graft libero`, and — the load-bearing part — `setup.py` sets `include_package_data=True`, so the grafted `bddl_files`/`init_files`/`assets` end up in site-packages, not just the sdist.

**Blast radius:** the env-var change applies to the whole `pytest -m "gpu"` run, but `test_control_freq_reaches_real_robosuite_sim` is the only GPU-marked test that touches LIBERO config paths — every other LIBERO test is CPU/mocked and patches `get_libero_path`. No collateral on the other GPU policy tests.

**`init_states=False`:** doesn't weaken the test — the `control_freq` read-back is independent of init states, it matches the two mocked siblings, and the real bddl path is still exercised by `_make_envs_task`. Belt-and-suspenders with the workflow fix, which is fine.

**SC2155:** accurate — switching `export VAR="$(pwd)/..."` to a literal removes the command-substitution-masks-return-value warning, and the new quoted `mkdir`/`touch` lines don't introduce any.

**Validation:** the manually-dispatched GPU run (26306227758) shows *Run Pytest on GPU* = success; CPU / pre-commit / check-checklist are all green.

Two optional, non-blocking notes:
1. The workflows now use two different LIBERO config strategies — gpu points at the installed package via `set_libero_default_path()`, while cpu/regression still point at the empty `.github/assets/libero` tree. Intentional and explained in the inline comment, but a future reader editing one might not notice the other differs; a one-line cross-pointer would help.
2. `set_libero_default_path()` trusts that the install contains the assets. If a future LIBERO bump dropped a file (or moved assets to LFS without materializing them), this would resurface as a `FileNotFoundError` at env-build time rather than anything clearer. Not worth guarding now — the nightly is the right place to catch it.

LGTM.

---

PR #324, opened 2026-05-23
fix(scripts): support action_freq=None in fit_fast_tokenizer manual sampler

shuheng-liu:
## What this does

Fixes a crash in `opentau.scripts.fit_fast_tokenizer` when the mixture has `action_freq: null`.

The fast manual sampling path (`_sample_via_manual`, the default) calls `float(mixture_cfg.action_freq)` in `main()` before dispatch — that raises `TypeError: float() argument must be ... not 'NoneType'` on mixed-frequency mixtures. With mixed-frequency training landed in [#323](https://github.com/TensorAuto/OpenTau/pull/323), `action_freq=None` is now a first-class setting that opts every dataset out of resampling and runs each at its native fps.

This PR aligns the manual sampler with [`factory.resolve_delta_timestamps`](https://github.com/TensorAuto/OpenTau/blob/main/src/opentau/datasets/factory.py): when `action_freq is None`, each worker substitutes that dataset's `meta.info["fps"]`. The chunk then spans `chunk_size` consecutive native frames and the `scipy.interpolate.interp1d` call lands exactly on native frame boundaries (so it's effectively a no-op rather than a misaligned resample). The legacy `action_freq: float` path is unchanged.

The mixture-dataloader path (`--use-mixture-dataloader`) already worked since it routes through `factory.resolve_delta_timestamps`; only the default fast path needed the fix.

🐛 Bug

## How it was tested

- Added `tests/scripts/test_fit_fast_tokenizer.py::TestSampleChunksManualActionFreq::test_none_action_freq_uses_native_fps`. Creates a 30 fps synthetic dataset via `lerobot_dataset_factory`, calls `_sample_chunks_for_dataset_manual` with `action_freq=None`, and confirms (a) no crash, (b) chunks come back with the expected `(chunk_size, native_action_dim)` shape, and (c) the same call with `action_freq=DEFAULT_FPS` succeeds with matching chunk counts.
- Full CPU suite: `uv run pytest -m "not gpu" -n auto tests/scripts tests/configs` → 259 passed.
- Pre-commit: `uv run pre-commit run --files src/opentau/scripts/fit_fast_tokenizer.py tests/scripts/test_fit_fast_tokenizer.py` → all hooks pass.

```bash
uv run pytest -sx tests/scripts/test_fit_fast_tokenizer.py
```

End-to-end fit against a mixed-frequency mixture (CPU-only):

```bash
uv run python -m opentau.scripts.fit_fast_tokenizer \
    --mixture-json /path/to/mixture_with_action_freq_null.json \
    --out-dir /tmp/fast-mix-freq \
    --chunk-size 30 --pilot
```

## Checklist

### DIFF
--- a/src/opentau/scripts/fit_fast_tokenizer.py
+++ b/src/opentau/scripts/fit_fast_tokenizer.py
@@ -405,7 +405,7 @@ def _compute_budgets_weighted(mixture_cfg: DatasetMixtureConfig, total_chunks: i
 
 
 def _sample_chunks_for_dataset_manual(
-    args: tuple[int, Any, int, int, float, int],
+    args: tuple[int, Any, int, int, float | None, int],
 ) -> tuple[int, list[np.ndarray], str | None]:
     """Worker: sample ``n_chunks`` action chunks of shape ``(chunk_size, native_dim)``
     resampled to ``action_freq`` via ``scipy.interpolate.interp1d``.
@@ -414,6 +414,11 @@ def _sample_chunks_for_dataset_manual(
     overhead. The interp matches LeRobot's vector_resample_strategy="nearest"
     closely enough for tokenizer fitting (``kind="linear"`` here is actually
     slightly more accurate; we don't need the strict step-resample behavior).
+
+    When ``action_freq is None`` (mixed-frequency mixture), each dataset is
+    sampled at its own native fps -- mirrors ``factory.resolve_delta_timestamps``
+    substituting ``ds_meta.fps`` so the chunk is ``chunk_size`` consecutive
+    native frames and the interp lands exactly on native frame boundaries.
     """
     idx, cfg, n_chunks, chunk_size, action_freq, seed = args
     if n_chunks <= 0:
@@ -426,6 +431,8 @@ def _sample_chunks_for_dataset_manual(
 
         meta = LeRobotDatasetMetadata(cfg.repo_id, root=cfg.root, revision=cfg.revision)
         native_fps = float(meta.info["fps"])
+        if action_freq is None:
+            action_freq = native_fps
         target_dt = 1.0 / action_freq
         chunk_duration = (chunk_size - 1) * target_dt  # seconds spanned by a chunk
         # Need at least ceil(chunk_duration * native_fps) + 1 native frames per chunk.
@@ -513,7 +520,7 @@ def _sample_via_manual(
     mixture_cfg: DatasetMixtureConfig,
     total_chunks: int,
     chunk_size: int,
-    action_freq: float,
+    action_freq: float | None,
     seed: int,
     num_workers: int,
 ) -> tuple[list[np.ndarray], list[int], dict[int, str]]:
@@ -1016,10 +1023,12 @@ def main() -> int:
     logger.info("Parsing mixture JSON: %s", args.mixture_json)
     mixture_cfg, mixture_hash = _parse_mixture(args.mixture_json)
     logger.info(
-        "Mixture has %d datasets (hash=%s); action_freq=%g",
+        "Mixture has %d datasets (hash=%s); action_freq=%s",
         len(mixture_cfg.datasets),
         mixture_hash[:12],
-        mixture_cfg.action_freq,
+        "None (mixed-frequency, per-dataset native fps)"
+        if mixture_cfg.action_freq is None
+        else f"{mixture_cfg.action_freq:g}",
     )
 
     # --- Phase 2: build mixture & gather chunks ---
@@ -1077,7 +1086,7 @@ def main() -> int:
             mixture_cfg,
             args.total_chunks,
             args.chunk_size,
-            float(mixture_cfg.action_freq),
+            mixture_cfg.action_freq,
             args.seed,
             args.num_workers,
         )
--- a/src/opentau/scripts/fit_fast_tokenizer.py
+++ b/src/opentau/scripts/fit_fast_tokenizer.py
@@ -38,8 +38,9 @@ Pipeline (all CPU):
        will see.
     3. Drain the weighted dataloader until ``--total-chunks`` action
        chunks have been collected; each batch yields chunks already
-       resampled to ``mixture.action_freq`` and right-padded to
-       ``max_action_dim``.
+       resampled to ``mixture.action_freq`` (or to each dataset's native
+       fps when ``action_freq is None`` -- mixed-frequency mixtures) and
+       right-padded to ``max_action_dim``.
     4. Min-max-normalize each chunk to ``[-1, 1]`` using
        ``mixture.meta.stats["actions"]`` -- the same min/max the policy
        will use at training via
@@ -191,8 +192,10 @@ def parse_args() -> argparse.Namespace:
             "standardization). Default: a hand-rolled equivalent that "
             "weight-allocates a budget per dataset, reads action+timestamp "
             "columns directly from parquet, and resamples to "
-            "mixture.action_freq via scipy.interp1d. Both paths respect "
-            "the mixture's weights and global FPS."
+            "mixture.action_freq via scipy.interp1d -- or to each dataset's "
+            "native fps when mixture.action_freq is None (mixed-frequency "
+            "mode). Both paths respect the mixture's weights and the same "
+            "fps convention."
         ),
     )
     p.add_argument(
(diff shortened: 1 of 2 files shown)


shuheng-liu (commented):
## Review

**Fix is correct and well-scoped.** The change mirrors `factory.resolve_delta_timestamps` (line 138-139) — substituting `meta.info["fps"]` per-dataset when `action_freq is None` — and the `float | None` type updates are consistent across `_sample_chunks_for_dataset_manual`, `_sample_via_manual`, and `main()`. The log-formatter swap from `%g` to `%s` is the right call.

### Findings

**1. Test gap (actionable).** The test's docstring and inline comment promise the two paths return matching chunks:

> They span the same wall-clock duration and target the same native timestamps, so the BPE-corpus content is equivalent under both call styles.

> Same seed + same effective action_freq must yield identical chunks.

…but no assertion compares `chunks_none[i]` against `chunks_native[i]` — only `err`, `idx`, `len`, and per-chunk shape are checked. A silent drift in the `None` branch (e.g. an off-by-one in `target_dt` indexing, or a future change to how `native_fps` is cast) would still pass this test today. Suggest:

```python
for c_none, c_native in zip(chunks_none, chunks_native, strict=True):
    np.testing.assert_allclose(c_none, c_native, rtol=1e-6, atol=1e-6)
```

(`assert_allclose` rather than `assert_array_equal` because the interp may land on float-rounded target timestamps even though the *intended* sampling points are identical — `t0 + k*(1/30)` is not bit-exact-equal to `native_timestamps[s+k]` for general `t0`.)

**2. Doc drift (nit, optional).** Two strings still describe the global-fps path only:
- `src/opentau/scripts/fit_fast_tokenizer.py:41` — *"resampled to `mixture.action_freq` and right-padded"*
- `src/opentau/scripts/fit_fast_tokenizer.py:194-195` — *"Both paths respect the mixture's weights and global FPS"*

A one-liner about the per-dataset-native-fps mode would help a cold reader. Out of scope here but cheap.

Nothing blocking — item 1 would make the test prove its own claim.

---

shuheng-liu:
Both findings addressed in b8f6b90:

**1. Test gap.** Added the `np.testing.assert_allclose(c_none, c_native, rtol=1e-6, atol=1e-6)` comparison the docstring promised. Also restructured to reuse one dataset (the worker is read-only on disk) so the comparison isn't pulling in any non-determinism in `lerobot_dataset_factory`'s synthetic data path — the *only* difference between the two calls is now the `action_freq` arg, which is what we actually want to assert about.

**2. Doc drift.** Updated both strings:
- Module docstring phase 3: now mentions "resampled to `mixture.action_freq` (or to each dataset's native fps when `action_freq is None` — mixed-frequency mixtures)".
- `--use-mixture-dataloader` CLI help: same treatment.

Local: test passes, pre-commit clean.

shuheng-liu (commented):
## Second pass — extending finding (1)

I ran the test locally and dug deeper into the "chunks should match" claim. The result changes my recommendation in a meaningful way.

**Empirical result.** Running the worker twice against the test fixture:

| setup | max abs diff between `chunks_none[i]` and `chunks_native[i]` |
|---|---|
| Two separate roots (as in the PR) | **~9.68e-01** (chunks unrelated) |
| Single shared root | **0.00e+00** (chunks bit-identical) |

So the chunks would be bit-identical *if* both calls were made against the same dataset — but the test as written can't observe this because each `lerobot_dataset_factory(root=...)` call materializes a *fresh* synthetic dataset with different random action values. The two roots aren't just isolated file paths; they're two independent datasets. So `chunks_none` is sampled from dataset A and `chunks_native` from dataset B, and even though both invocations pick the same episode indices and start positions (same `np.random.default_rng(seed=0)`), the underlying action data differs.

**The "caching side-effects" comment is the bug source.** The justification given for two roots:

> Use a separate root to keep the deterministic per-dataset rng split from being contaminated by any caching side-effects on the first call.

…doesn't hold up. The worker constructs a fresh `LeRobotDatasetMetadata` from disk and a fresh `np.random.default_rng(seed)` each call — there is no shared state between invocations. The precaution is unfounded and actively prevents the equivalence check the docstring promises.

**Suggested patch.** One root, then `assert_array_equal`:

```python
def test_none_action_freq_uses_native_fps(self, tmp_path, lerobot_dataset_factory):
    root = tmp_path / "ds"
    lerobot_dataset_factory(root=root, total_episodes=3, total_frames=150)

    _, chunks_none, err_none = self._run(root, action_freq=None)
    _, chunks_native, err_native = self._run(root, action_freq=float(DEFAULT_FPS))
    assert err_none is None and err_native is None
    assert len(chunks_none) == len(chunks_native) == 8
    for c_none, c_native in zip(chunks_none, chunks_native, strict=True):
        assert c_none.shape == (10, 6)
        np.testing.assert_array_equal(c_none, c_native)
```

(Corrects my prior `assert_allclose` suggestion — `assert_array_equal` is right. When the interp lands on existing native frame boundaries, the output is bit-equal, verified locally.)

Item (2) from the first pass (doc drift at lines 41 and 194-195) still stands as a nit.

---

shuheng-liu:
Adopted in 0f5614d. Empirically the two paths produce **bit-identical** chunks on the test fixture (max abs diff = 0.0 across all 8 chunks), so `assert_array_equal` is the right call — any future drift in the substitution will surface immediately instead of getting absorbed by an rtol.

One note for context: my earlier `assert_allclose` was defending against a worry that turned out to be unfounded. `target_ts = t0 + k*(1/fps)` is not bit-equal to `(s+k)/fps` in float64 *in general*, but the interp's clip + linear-interp path lands exactly on the native sample whenever the offset is within float epsilon, which is what we see here. Worth being explicit about — the comment now spells out why bit-equality holds.

Note that the test file in the PR already uses one root (since b8f6b90, before this comment) — your "Two separate roots (as in the PR)" table was reading the original 89adc51 commit, not the current head. The suggested patch matches what's already there modulo `assert_array_equal` vs `assert_allclose`, which is now reconciled.

shuheng-liu (commented):
## Third pass — type-signature observation

Having walked the call graph in both directions, the only thing left I'd note is a type-annotation gap that the PR widens but doesn't cause:

**`WeightedDatasetMixture.__init__` still declares `action_freq: float`** at `src/opentau/datasets/dataset_mixture.py:338`, but `fit_fast_tokenizer._build_mixture_parallel:711` now passes `train_cfg.dataset_mixture.action_freq` straight through — and after this PR, mixed-frequency mixtures (`action_freq=None`) can legitimately reach that path via `--use-mixture-dataloader`. The attribute is stored as informational state and never used arithmetically inside `WeightedDatasetMixture`, so there's no runtime bug — but the signature now lies. The project doesn't run mypy/pyright in pre-commit, so this lands silently.

Tightening it to `action_freq: float | None` (and updating the docstring at line 59) is a one-line follow-up that probably belongs in a separate type-sweep PR, not here, since the fix in this PR is correctly scoped to just the crash in `fit_fast_tokenizer`. Flagging it so it's not forgotten.

### Summary of review across all three passes

| # | Finding | Status |
|---|---|---|
| 1 | Test asserts only count/shape; docstring promises chunk equivalence | **Actionable** — fix in this PR |
| 2 | Two-root design defeats the chunk-equivalence check (chunks differ ~0.97 with two roots, bit-equal with one). Use a single root + `np.testing.assert_array_equal` | **Actionable** — fix in this PR (subsumes #1) |
| 3 | Module docstring at L41 + argparse help at L194-195 still describe global-fps path only | Optional nit |
| 4 | `WeightedDatasetMixture.__init__` `action_freq: float` mismatches the now-reachable `None` value | Out of scope; follow-up |

Fix itself is correct and minimal. LGTM modulo finding #2.

---

shuheng-liu:
Third-pass summary:

- **Findings #1, #2, #3 are already in the PR.** Single-root structure + `assert_array_equal` landed in b8f6b90 and 0f5614d (commits before this comment). Module docstring (L41-ish) + `--use-mixture-dataloader` CLI help (L194-195) updated in b8f6b90 to mention the per-dataset-native-fps mode. Your summary table looks like it was generated against an earlier head — current head 0f5614d already reflects all three.

- **Finding #4 deferred.** Agree it's out of scope for this PR — the fix here is correctly narrowed to the crash in `fit_fast_tokenizer`. Filed a follow-up to tighten `WeightedDatasetMixture.__init__` from `action_freq: float` to `float | None`, update the docstring example at line 59, and add a unit test. Will land separately so this PR stays focused.

Nothing left on my end here — happy to mark ready-for-review whenever you'd like to flip it out of draft.

PR #326, opened 2026-05-23
fix(datasets): resolve per-episode task_index from parquet for speed bucketing

shuheng-liu:
## What this does

Two-commit fix for the noisy `Episode task label '...' is not present in tasks.jsonl` warning and the resulting silent degradation of `speed_raw` to the median bucket for most episodes when `tasks.jsonl` has been paraphrased relative to `episodes.jsonl::tasks[0]`.

**Root cause.** The previous `episode_to_task_index_from_episodes` resolved each episode's task by string-matching `episodes.jsonl::tasks[0]` against `tasks.jsonl`. Any paraphrase between the two files broke the match, dropping every affected episode to `SPARSE_TASK_BUCKET = 50` in the downstream speed-bucket lookup. The string match is unnecessary: every row of the per-episode parquet already carries an authoritative integer `task_index` column (the same one `__getitem__` reads).

**Commit 1 — switch the lookup to the parquet.** Replaces `episode_to_task_index_from_episodes` with `episode_to_task_index_from_hf_dataset`, which reads `task_index` directly from the per-frame parquet via `episode_data_index`. Paraphrase is now inert. The new warning fires only when the parquet itself points at a `task_index` that `tasks.jsonl` doesn't define (genuine metadata corruption). Adds a one-shot migration in `load_or_compute_speed_percentiles` that detects + recomputes existing `meta/speed_percentiles.jsonl` files written under the buggy code path (those files are missing rows for every drifted task; the new loader checks the expected vs on-disk task-index sets and rewrites once).

**Commit 2 — directional staleness check.** Commit 1 also changed `episode_to_task_index` to cover *selected* episodes only (the parquet read can only see selected ones). For a heterogeneous-mixture run this makes the on-disk per-task `n_episodes` total naturally larger than the current load's count, firing the existing `Stale ...` warning on every dataset in the mixture even though nothing is stale. The fix restricts the warning to the genuine direction (`on_disk_total < current_total`, i.e. new episodes appended). Subset / mixture loads (`on_disk_total > current_total`) are silent — the on-disk percentiles are a more robust sample than a recompute on the subset would produce.

## How it was tested

- `tests/datasets/test_speed_percentiles.py`: replaced the four string-lookup tests with `TestEpisodeToTaskIndexFromHfDataset` (5 tests covering happy path, unknown-index warning, dedup, empty-episodes short-circuit, paraphrase-doesn't-warn). Added `test_recompute_when_persisted_file_missing_tasks` for the migration, `test_no_stale_warning_for_subset_load` for directional staleness, and `test_clean_regeneration_after_file_deletion` for the rm-and-rebuild workflow.
- All 43 cpu tests in `tests/datasets/test_speed_percentiles.py` pass; 441 tests in `tests/datasets/` pass.
- Verified on a live `pi07_paligemma_pretrain_10percent_mix_freq` run: the `Episode task label '...' is not present in tasks.jsonl` warnings are gone, and after the second commit the `Stale ...` warnings on subset loads are also gone. Existing pre-fix `meta/speed_percentiles.jsonl` files are migrated automatically on first load with the new code.

```bash
pytest -sx tests/datasets/test_speed_percentiles.py
```

To see the end-to-end effect, load any dataset whose `tasks.jsonl` is paraphrased relative to `episodes.jsonl::tasks[0]`:

```python
from collections import Counter
from opentau.datasets.lerobot_dataset import LeRobotDataset
ds = LeRobotDataset("<repo_id>")
print(Counter(ds.speed_raw_by_episode.values()))
```

Before this PR, the distribution is dominated by `50` (sparse fallback). After this PR — once `meta/speed_percentiles.jsonl` is migrated on first load — buckets span the full `{0, 10, 20, ..., 100}` range for tasks with ≥10 distinct episode lengths.

## Checklist

### DIFF
--- a/src/opentau/datasets/speed_percentiles.py
+++ b/src/opentau/datasets/speed_percentiles.py
@@ -239,10 +239,13 @@ def load_or_compute_speed_percentiles(
 
     If ``root / meta/speed_percentiles.jsonl`` already exists, load it
     verbatim — staleness is accepted by design (a WARNING is logged when
-    the on-disk per-task ``n_episodes`` total disagrees with the current
-    load, but the file is still trusted; delete it to force a recompute).
-    Otherwise compute the percentiles from ``episode_lengths`` and
-    persist the file.
+    the on-disk file was computed from *fewer* episodes than the current
+    load is using, i.e. episodes were appended since; subset loads, where
+    the current load uses fewer episodes than on-disk, are silent because
+    the on-disk percentiles are a more robust sample). The file is
+    trusted in both cases — delete it to force a recompute. Otherwise
+    compute the percentiles from ``episode_lengths`` and persist the
+    file.
 
     One-shot migration: when the on-disk file is missing rows for tasks
     that *do* have resolved episodes in this load, the file is recomputed
@@ -311,15 +314,24 @@ def load_or_compute_speed_percentiles(
             missing = expected_indices - on_disk_indices
             if not missing:
                 if is_main_or_solo:
-                    # `episode_to_task_index` already drops episodes that couldn't
-                    # be resolved, so its length is the per-task episode count we
-                    # want to compare against the on-disk sum.
+                    # ``episode_to_task_index`` covers only episodes that
+                    # were both (a) resolved by the per-frame ``task_index``
+                    # lookup and (b) selected for this load. The directional
+                    # check below warns only when the current load has *more*
+                    # episodes than the on-disk file was computed from
+                    # (i.e. episodes were appended since the file was
+                    # written). Subset loads (current uses fewer episodes
+                    # than on-disk) are silent — the on-disk percentiles
+                    # were computed from a larger, more robust sample, so
+                    # using them is if anything an improvement.
                     on_disk_total = sum(int(row.get("n_episodes", 0)) for row in rows)
                     current_total = len(episode_to_task_index)
-                    if on_disk_total != current_total:
+                    if on_disk_total < current_total:
                         logging.warning(
-                            "Stale %s: on-disk per-task n_episodes sums to %d, current load has %d. "
-                            "Using on-disk percentiles as-is; delete the file to force recompute.",
+                            "Stale %s: on-disk percentiles were computed from %d episodes, "
+                            "but current load has %d. New episodes may have been appended "
+                            "since the file was written. Using on-disk percentiles as-is; "
+                            "delete the file to force recompute.",
                             path,
                             on_disk_total,
                             current_total,
--- a/src/opentau/datasets/speed_percentiles.py
+++ b/src/opentau/datasets/speed_percentiles.py
@@ -41,11 +41,16 @@ import os
 import uuid
 from collections import defaultdict
 from pathlib import Path
+from typing import TYPE_CHECKING
 
 import numpy as np
 
 from opentau.datasets.utils import load_jsonlines, write_jsonlines
 from opentau.utils.accelerate_utils import get_proc_accelerator
+
+if TYPE_CHECKING:
+    import datasets
+    import torch
 
 # Boundary percentiles, ascending. Length 10 by design: 11 buckets.
 SPEED_PERCENTILES: tuple[int, ...] = (5, 15, 25, 35, 45, 55, 65, 75, 85, 95)
@@ -140,9 +145,9 @@ def _group_lengths_by_task(
 
 
 def episode_to_task_index_from_hf_dataset(
-    hf_dataset,
+    hf_dataset: datasets.Dataset,
     episodes: list[int],
-    episode_data_index: dict[str, object],
+    episode_data_index: dict[str, torch.Tensor],
     epi2idx: dict[int, int],
     valid_task_indices: set[int],
 ) -> dict[int, int]:
@@ -247,15 +252,18 @@ def load_or_compute_speed_percentiles(
     compute the percentiles from ``episode_lengths`` and persist the
     file.
 
-    One-shot migration: when the on-disk file is missing rows for tasks
-    that *do* have resolved episodes in this load, the file is recomputed
-    (logged at INFO) and rewritten with the full task set. This recovers
-    datasets whose existing percentile file was written before
-    :func:`episode_to_task_index_from_hf_dataset` — those files only have
-    rows for the subset of tasks whose ``episodes.jsonl::tasks[0]`` string
-    happened to match ``tasks.jsonl``, so almost every episode bucketed to
-    ``SPARSE_TASK_BUCKET``. The recompute is a no-op once the file is
-    rewritten.
+    Append-only migration: when the on-disk file is missing rows for
+    tasks that *do* have resolved episodes in this load, percentiles are
+    computed for the missing tasks only and appended to the file (logged
+    at WARNING so the rewrite is visible). Existing rows are preserved
+    verbatim. This recovers datasets whose existing percentile file was
+    written before :func:`episode_to_task_index_from_hf_dataset` — those
+    files were missing rows for tasks whose ``episodes.jsonl::tasks[0]``
+    string didn't match ``tasks.jsonl``, so every episode of those tasks
+    bucketed to ``SPARSE_TASK_BUCKET``. The append-only design avoids
+    clobbering existing percentiles with subset-derived samples when the
+    current load is a mixture / subset; force a full recompute by
+    deleting the file.
 
     Distributed-safe: every rank computes the percentiles in-memory
     (the result is deterministic from the inputs), but only the main
@@ -307,9 +315,10 @@ def load_or_compute_speed_percentiles(
     # for every subsequent collective in the mixture-load loop — manifesting
     # as a NCCL hang at a much later (and entirely unrelated) sync point.
     try:
+        existing_rows: list[dict] = []
         if path.is_file():
-            rows = load_jsonlines(path)
-            on_disk_indices = {int(row["task_index"]) for row in rows}
+            existing_rows = load_jsonlines(path)
+            on_disk_indices = {int(row["task_index"]) for row in existing_rows}
             expected_indices = set(episode_to_task_index.values())
             missing = expected_indices - on_disk_indices
             if not missing:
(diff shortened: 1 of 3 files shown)


src/opentau/datasets/speed_percentiles.py
```diff
@@ -305,10 +308,45 @@ def load_or_compute_speed_percentiles(
...
+
+            if is_main_or_solo:
+                logging.info(
+                    "Recomputing %s: on-disk file is missing %d task(s) that have resolved "
+                    "episodes in this load — likely written before the per-frame task_index "
+                    "fix. Rewriting with the full task set.",
+                    path,
+                    len(missing),
+                )
+            # Fall through to the compute + write branch below.
```
shuheng-liu: **Subset-load migration may degrade on-disk data quality.**

The fall-through recomputes `by_task` from the *current* load's `episode_lengths` / `episode_to_task_index`. If this load is a subset (e.g. a 10% mixture), the rewrite will use only the subset's data for *every* task — including tasks whose pre-fix on-disk percentiles happened to be string-resolved correctly from the full dataset. Those entries get clobbered with subset-derived (less robust) percentiles.

Scenario:
1. Pre-fix file has correct percentiles for tasks `{0, 2, 5}` (string-match worked) computed from the full ~450-episode dataset.
2. New code loads a 10% subset that now resolves tasks `{0, 1, 2, 5}` via the parquet.
3. `missing = {1}` → migration fires → file rewritten with subset (~45 episode) percentiles for all four tasks, losing the higher-fidelity numbers for `{0, 2, 5}`.
4. The next full load sees no missing tasks and triggers the directional `on_disk_total < current_total` warning — but only after the subset training has already happened.

Also relevant: subsequent subset loads with a different task-resolution outcome would each trigger another full rewrite (a migration storm across heterogeneous subsets).

Mitigation options to consider:
- Gate the migration to full loads only (e.g. additionally check `len(episode_to_task_index) >= on_disk_total` before rewriting), letting subset loads use the incomplete file plus the existing SPARSE_TASK_BUCKET fallback for the missing tasks.
- Bump the migration log from INFO to WARNING so users notice the one-shot rewrite, since it materially changes downstream behavior.

Given this is a one-time legacy migration, accepting the trade-off and documenting it is also defensible — but the current docstring frames the recompute as harmless ("The recompute is a no-op once the file is rewritten") which understates the subset case.

---

src/opentau/datasets/speed_percentiles.py
```diff
@@ -138,44 +139,64 @@ def _group_lengths_by_task(
...
 
-def episode_to_task_index_from_episodes(
-    episodes: dict[int, dict],
-    task_to_task_index: dict[str, int],
+def episode_to_task_index_from_hf_dataset(
+    hf_dataset,
+    episodes: list[int],
+    episode_data_index: dict[str, object],
+    epi2idx: dict[int, int],
+    valid_task_indices: set[int],
```
shuheng-liu, line 152: Minor: `hf_dataset` has no type annotation while the rest of the params are typed; `episode_data_index: dict[str, object]` is weaker than it needs to be — `get_episode_data_index` (`utils.py:845`) returns `dict[str, torch.Tensor]`. A forward-ref string annotation would keep this consistent without forcing an `import datasets` at module top:

```python
hf_dataset: "datasets.Dataset",
episodes: list[int],
episode_data_index: dict[str, "torch.Tensor"],
```

---

src/opentau/datasets/speed_percentiles.py
```diff
@@ -138,44 +139,64 @@ def _group_lengths_by_task(
...
+            ``episode_data_index["from"]``.
+        valid_task_indices: All ``task_index`` values defined in
+            ``tasks.jsonl`` (i.e. ``set(meta.task_to_task_index.values())``).
+            Used solely to detect parquet rows that point at a task the
+            metadata doesn't know about.
     """
+    if not episodes:
+        return {}
+    start_indices = [int(episode_data_index["from"][epi2idx[ep]].item()) for ep in episodes]
+    task_indices = hf_dataset.with_format("arrow").select(start_indices)["task_index"].to_pylist()
```
shuheng-liu, line 190: Worth noting in passing: this assumes `self.hf_dataset` is laid out in `self.episodes` order so that `episode_data_index["from"][epi2idx[ep]]` is the correct row offset. That assumption is guaranteed by `load_hf_dataset` (`lerobot_dataset.py:1737`), which sorts files in `self.episodes` order — so this is sound today. Just flagging the cross-file invariant since the call site is a few hundred lines away from where it's enforced. Not a change request.

---

tests/datasets/test_speed_percentiles.py
```diff
@@ -356,6 +539,61 @@ def test_handles_missing_task_in_lookup_defensively(self, tmp_path):
...
+        # The on-disk file was rewritten with both tasks.
+        rows = load_jsonlines(path)
+        assert {int(r["task_index"]) for r in rows} == {0, 1}
+
+        # Second call takes the fast read path — no Recompute, no Stale.
+        caplog.clear()
+        with caplog.at_level("INFO"):
+            load_or_compute_speed_percentiles(tmp_path, el, e2t, t2i)
+        assert not any("Recomputing" in r.message for r in caplog.records)
+        assert not any("Stale" in r.message for r in caplog.records)
```
shuheng-liu: Missing coverage: this test exercises the migration with a "full" load (all episodes in `_fake_meta` are passed in). The interesting case for the inline comment above on `speed_percentiles.py:341-349` is the migration firing on a *subset* load — pre-fix file has rows for some tasks, the current (subset) load resolves an additional task, and the rewrite clobbers the pre-existing rows with subset-sized samples. Worth pinning whatever behavior is intended there, even if it's "we accept the trade-off."

---

shuheng-liu (commented):
## Review summary

Solid two-commit fix. Root cause analysis is correct, the parquet `task_index` is genuinely authoritative, and the directional staleness check is the right call for heterogeneous-mixture training. Test coverage is strong — happy path, paraphrase, unknown-index dedup, empty-episodes short-circuit, regen-after-deletion, and the migration are all pinned.

**Correctness spot-checks (all pass):**
- `episode_data_index["from"][epi2idx[ep]]` is the right offset because `load_hf_dataset` (`lerobot_dataset.py:1737`) builds the parquet files in `self.episodes` order.
- `hf_dataset.with_format("arrow").select(start_indices)["task_index"].to_pylist()` preserves index order, so the `zip(episodes, task_indices, strict=True)` alignment holds.
- Scoping `speed_raw_by_episode` from `self.meta.episodes` to `self.episodes` (`lerobot_dataset.py:1511`) is safe — the only downstream read at line 2186 always uses an `ep_idx` drawn from `self.hf_dataset`, i.e. always a selected episode.

**Main concern (inline on `speed_percentiles.py:341-349`):** the legacy-migration fall-through recomputes from the *current* load, so a subset load (the dominant case for mixture training) will clobber any well-resolved tasks on disk with subset-sized samples. Worth either gating to full loads only or bumping the migration log to WARNING so the rewrite is visible.

**Minor nits:**
- `hf_dataset` parameter has no type annotation, and `episode_data_index: dict[str, object]` is weaker than the actual `dict[str, torch.Tensor]`.
- No test for the migration-on-subset case — would be useful to pin whatever behavior you settle on.

Nothing here is blocking; the substantive item is a trade-off to discuss rather than a strict bug.

---

src/opentau/datasets/speed_percentiles.py
```diff
@@ -304,28 +315,79 @@ def load_or_compute_speed_percentiles(
...
+            task_indices_to_compute = missing
+        else:
+            # File doesn't exist — fresh compute over every resolved task.
+            task_indices_to_compute = set(episode_to_task_index.values())
+
+        # Compute percentiles for the target task set only.
+        by_task: dict[int, list[int]] = defaultdict(list)
+        for ep_idx, task_idx in episode_to_task_index.items():
+            if task_idx in task_indices_to_compute:
+                by_task[task_idx].append(episode_lengths[ep_idx])
```
shuheng-liu, line 374: Nit: the new inline grouping at lines 371-374 replaces the only call site of `_group_lengths_by_task` (defined at line 136), leaving it as dead code. Either remove the function or keep it and call it here with a `task_indices_to_compute` filter parameter — the inlined version doesn't share enough with the helper to justify keeping both.

---

src/opentau/datasets/speed_percentiles.py
```diff
@@ -304,28 +315,79 @@ def load_or_compute_speed_percentiles(
...
                 "task_index": task_idx,
                 "task": index_to_task.get(task_idx, ""),
                 "n_episodes": len(by_task.get(task_idx, [])),
-                "percentiles": percentiles[task_idx],
+                "percentiles": new_percentiles[task_idx],
             }
-            for task_idx in sorted(percentiles)
+            for task_idx in sorted(new_percentiles)
         ]
+        rows_to_write = existing_rows + new_rows
```
shuheng-liu, line 386: After multiple migrations the file ordering becomes `[existing_rows in original order] + [new_rows sorted by task_index]` rather than globally sorted. Cosmetic only — affects human inspection of the JSONL, not correctness of any reader. If you care, `sorted(rows_to_write, key=lambda r: int(r["task_index"]))` keeps the file tidy.

---

shuheng-liu (commented):
## Second-pass review (979752a)

All three items from the prior review are addressed cleanly:

1. **Migration is now append-only** — `task_indices_to_compute = missing`, existing rows preserved verbatim, merged dict returned. The trade-off ("first migrator wins per task; users force a full recompute by deleting") is documented in the docstring and the warning text. ✓
2. **Migration log bumped to WARNING** with the missing-task list inlined for diagnosability. ✓
3. **Type annotations** via `TYPE_CHECKING` for `datasets.Dataset` / `torch.Tensor` — clean and avoids the eager import. ✓

**New test `test_migration_on_subset_load_does_not_clobber_existing`** pins the exact regression I flagged: full-set taskA percentiles are stored first, a subset load then resolves an additional taskB, and the assertion `out[0] == full_task_a` confirms the high-fidelity taskA row survives. The use of `subset_lengths = full_lengths[:10]` is sharp because the subset *would* have produced different (non-None) percentiles, so the test distinguishes "preserved" from "happened to be same". ✓

**Correctness re-checks on the new code:**
- `task_indices_to_compute = missing = expected - on_disk_indices` is disjoint from `existing_rows`'s task indices, so the `merged.update(new_percentiles)` at line 405 can't silently overwrite a preserved row.
- All ranks compute the same `new_percentiles` (deterministic from shared inputs), only `is_main_or_solo` writes, the `wait_for_everyone()` barrier in `finally` runs unconditionally. Distributed safety preserved.
- The append-only path correctly handles a row in `existing_rows` whose `task_index` is *not* in `expected_indices` (e.g. tasks no longer in the current selection) — it's still preserved, which is the right behavior.

Two minor nits inline (dead `_group_lengths_by_task`, file ordering after migration). Both cosmetic, neither blocking.

CI is green. Ready to merge once the nits are decided on.

---

PR #327, opened 2026-05-23
fix(scripts): pass format="mp4" to wandb.Video for grid_summary logging

shuheng-liu:
## What this does

Fixes the `wandb: WARNING format argument was not provided, defaulting to gif` warning that fires on every eval step that logs a `grid_summary.mp4`. The single `wandb.Video(grid_path)` call at `src/opentau/scripts/train.py:826` now passes `format="mp4"` explicitly. `grid_path` is always a path to an mp4 file (produced by `eval.py::create_grid_summary_video` and globbed by `eval.py::collect_grid_summary_videos`), so `"mp4"` is the only correct value.

The warning also reflected a real latent bug: with no `format=` kwarg, wandb defaulted to `gif`, silently mis-classifying the artifact in the wandb UI as a gif rather than the mp4 we wrote to disk.

(🐛 Bug)

## How it was tested

- `pre-commit run --files src/opentau/scripts/train.py` — all hooks pass.
- `LIBERO_CONFIG_PATH=.github/assets/libero pytest -m "not gpu" -n auto` — 1291 passed, 14 skipped. One pre-existing unrelated failure: `tests/utils/test_libero_utils.py::test_libero2torch` (`FileNotFoundError` on LIBERO benchmark assets not present in the dev env).
- This call site is not exercised by the CPU suite — it requires wandb-enabled eval with rendered episodes. The change is a strict addition of a `format="mp4"` kwarg to a one-line constructor, with no behavior change for any other path.

```bash
gh pr checkout <PR#>
grep -n "wandb.Video" src/opentau/scripts/train.py
```

The line should read:

```python
f"Eval Videos/{task_name}": wandb.Video(grid_path, format="mp4")
```

Next eval step that logs a grid summary should produce no `wandb: WARNING format argument was not provided` line, and the artifact in the wandb UI should appear as an mp4 (not a gif).

## Checklist

### DIFF
--- a/src/opentau/scripts/train.py
+++ b/src/opentau/scripts/train.py
@@ -823,7 +823,7 @@ def train(cfg: TrainPipelineConfig):
                 # videos on a multi-node run without a shared FS won't be seen).
                 if cfg.wandb.enable and cfg.eval.max_episodes_rendered > 0:
                     grid_videos = {
-                        f"Eval Videos/{task_name}": wandb.Video(grid_path)
+                        f"Eval Videos/{task_name}": wandb.Video(grid_path, format="mp4")
                         for task_name, grid_path in collect_grid_summary_videos(videos_dir)
                     }
                     if grid_videos:


shuheng-liu:
_Posted on behalf of @shuheng-liu by Claude Code._

LGTM. Quick verification of the four things worth checking on a one-liner like this:

1. **Fix is correct.** `wandb.Video` emits this warning unconditionally when `format=None`, regardless of whether the input is a numpy array or a file path. For path inputs the kwarg has no effect on the uploaded artifact (container is determined by the file extension via wandb's `Video.EXTS` validation), so `format="mp4"` just silences the warning without changing what gets uploaded. Also future-proofs against wandb 0.20 where the kwarg becomes required.

2. **Path is guaranteed to be `.mp4`.** Verified upstream:
   - `src/opentau/scripts/eval.py:402` — `grid_summary_path = videos_dir / "grid_summary.mp4"` (hard-coded extension).
   - `src/opentau/scripts/eval.py:912` — `videos_dir.rglob("grid_summary.mp4")` in `collect_grid_summary_videos`, so only `.mp4` files reach the call site.

3. **No other call sites missed.** `wandb.Video(` appears exactly once under `src/` (the line being patched), so this single change covers every place we log videos to wandb.

4. **Risk surface is null.** Touches `scripts/train.py` but only the wandb-logging path, not the forward/backward/optimizer loop, so the determinism invariant in `CLAUDE.md` is unaffected. The call site is gated on `cfg.wandb.enable and cfg.eval.max_episodes_rendered > 0`, which isn't exercised by `cpu_test.yml`, so no new test is needed.

Ship it once CI is green.

---

PR #328, opened 2026-05-23
fix(envs): zero-fill missing camera slots at eval; libero num_cams

shuheng-liu:
## What this does

Fixes a train↔eval mismatch in LIBERO eval rollouts when the policy was trained with `cfg.num_cams > 2` against a heterogeneous (multi-domain) mixture but the env exposes fewer cameras.

At training time, `BaseDataset._standardize_images` (`src/opentau/datasets/lerobot_dataset.py`) always emits a `cfg.num_cams`-wide image stack — when a dataset has fewer cameras than `cfg.num_cams`, the missing `cameraN` slots are zero-filled and `img_is_pad[i]=True` is set for those slots. The eval-side `preprocess_observation` did *neither*: missing keys stayed absent in the batch and `img_is_pad` was created as all-`False` regardless of how many cameras the env actually produced. With e.g. a `num_cams=4` policy evaluated on 2-camera LIBERO, the downstream effect is a strictly shorter VLM visual prefix at eval than at any training step, plus a `Warning: cameraN was missing from the batch during normalization` line from `Normalize` for every absent slot, every forward, every env, every step.

**Fix:** `preprocess_observation` now mirrors `_standardize_images` (`src/opentau/envs/utils.py`). Missing `cameraN` slots are zero-filled to `(B, 3, *cfg.resolution)` and the matching `img_is_pad[:, i]=True` is set. Since the keys are now present, `Normalize.warn_missing_keys` stops firing for them. To use a single-camera LIBERO rollout, set `env.camera_name="agentview_image"` directly — `camera_name.split(",")` already encodes both count and ordering of LIBERO render cameras, so no separate `num_cams` knob is needed on `LiberoEnv` (an earlier revision of this PR added one and was reverted after review).

## How it was tested

- `TestPreprocessObservationCameraPadding` (4 cases, `tests/envs/test_utils.py`): pins shape/pad-pattern parity for the zero-fill path — `num_cams=4` with 2 present cameras zero-fills `camera2`/`camera3` and sets `img_is_pad=[F,F,T,T]`; all-present case sets no pads; shape is always `(B, num_cams)`; zero-filled cameras use `cfg.resolution`, not the env's raw HxW.
- `pre-commit run` on all modified files: clean.
- `pytest tests/envs/ tests/scripts/test_eval.py -m "not gpu" -n auto`: all passing.
- Full CPU subset (`pytest -m "not gpu" -n auto`): 1299 passed, 14 skipped. The 2 local failures (`test_deferred_video_multi_episode`, `test_libero2torch`) are pre-existing flakes / missing-asset issues unrelated to this change.

```bash
LIBERO_CONFIG_PATH="$(pwd)/.github/assets/libero" \
    pytest -sx tests/envs/test_utils.py::TestPreprocessObservationCameraPadding
```

For an end-to-end LIBERO eval rollout with a policy whose `cfg.num_cams > 2`: no config change required — the next LIBERO eval should stop emitting `cameraN was missing` warnings and the policy receives a fully-padded camera stack matching the training-time `_standardize_images` shape contract. To run an agentview-only LIBERO rollout (single render camera, remaining slots zero-padded by `preprocess_observation`), set `env.camera_name="agentview_image"`.

## Checklist

### DIFF
--- a/src/opentau/envs/configs.py
+++ b/src/opentau/envs/configs.py
@@ -197,16 +197,16 @@ class LiberoEnv(EnvConfig):
         episode_length: Maximum length of each episode in steps.
         obs_type: Type of observations to use (e.g., ``"pixels_agent_pos"``).
         render_mode: Rendering mode for the environment (e.g., ``"rgb_array"``).
-        num_cams: Number of LIBERO cameras to expose at eval. Must be ``1``
-            (agentview only) or ``2`` (agentview + wrist eye-in-hand);
-            defaults to ``2``. When the underlying policy was trained with a
-            larger ``cfg.num_cams`` (e.g. a multi-domain mixture with 4 camera
-            slots), ``preprocess_observation`` zero-fills the remaining
-            ``cameraN`` slots so the train↔eval input structure stays aligned.
-        camera_name: Comma-separated names of LIBERO raw cameras to render.
-            Auto-derived from ``num_cams`` when unspecified (``num_cams=1`` →
-            ``"agentview_image"``, ``num_cams=2`` → ``"agentview_image,robot0_eye_in_hand_image"``).
-            If set explicitly it must contain exactly ``num_cams`` entries.
+        camera_name: Comma-separated LIBERO raw camera names to render — both
+            count and ordering of LIBERO cameras at eval are driven by this
+            string. Defaults to ``"agentview_image,robot0_eye_in_hand_image"``
+            (agentview + wrist eye-in-hand). Set to ``"agentview_image"``
+            (single camera) for agentview-only rollouts. When the underlying
+            policy was trained with a larger ``cfg.num_cams`` (e.g. a
+            multi-domain mixture with 4 camera slots),
+            ``preprocess_observation`` zero-fills the remaining ``cameraN``
+            slots so the train↔eval input structure stays aligned —
+            independent of how many real LIBERO cameras this field renders.
         init_states: Whether to initialize states randomly.
         camera_name_mapping: Optional mapping from camera names to standardized keys.
         features: Mapping from logical feature names to :class:`~opentau.configs.types.PolicyFeature` definitions.
@@ -219,7 +219,6 @@ class LiberoEnv(EnvConfig):
     episode_length: int = 520
     obs_type: str = "pixels_agent_pos"
     render_mode: str = "rgb_array"
-    num_cams: int = 2
     camera_name: str = "agentview_image,robot0_eye_in_hand_image"
     init_states: bool = True
     camera_name_mapping: dict[str, str] | None = None
@@ -238,23 +237,6 @@ class LiberoEnv(EnvConfig):
     )
 
     def __post_init__(self):
-        if self.num_cams not in (1, 2):
-            raise ValueError(
-                f"LIBERO env.num_cams must be 1 (agentview only) or 2 (agentview + wrist), "
-                f"got {self.num_cams}. LIBERO exposes at most two cameras."
-            )
-        # Auto-derive ``camera_name`` from ``num_cams`` when the user only set
-        # ``num_cams=1`` and left ``camera_name`` at the 2-cam default. Set both
-        # explicitly to use a non-default LIBERO camera pair.
-        if self.num_cams == 1 and self.camera_name == "agentview_image,robot0_eye_in_hand_image":
-            self.camera_name = "agentview_image"
-        n_in_name = len([c for c in self.camera_name.split(",") if c.strip()])
-        if n_in_name != self.num_cams:
-            raise ValueError(
-                f"LIBERO env: num_cams={self.num_cams} but camera_name has {n_in_name} "
-                f"entries ({self.camera_name!r}). Set them consistently."
-            )
-
         if self.fps <= 0:
             raise ValueError(
                 f"LIBERO env.fps (robosuite control frequency in Hz) must be positive, got {self.fps}"
--- a/src/opentau/envs/configs.py
+++ b/src/opentau/envs/configs.py
@@ -197,7 +197,16 @@ class LiberoEnv(EnvConfig):
         episode_length: Maximum length of each episode in steps.
         obs_type: Type of observations to use (e.g., ``"pixels_agent_pos"``).
         render_mode: Rendering mode for the environment (e.g., ``"rgb_array"``).
-        camera_name: Comma-separated names of cameras to use for rendering.
+        num_cams: Number of LIBERO cameras to expose at eval. Must be ``1``
+            (agentview only) or ``2`` (agentview + wrist eye-in-hand);
+            defaults to ``2``. When the underlying policy was trained with a
+            larger ``cfg.num_cams`` (e.g. a multi-domain mixture with 4 camera
+            slots), ``preprocess_observation`` zero-fills the remaining
+            ``cameraN`` slots so the train↔eval input structure stays aligned.
+        camera_name: Comma-separated names of LIBERO raw cameras to render.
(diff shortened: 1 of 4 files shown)


shuheng-liu:
## Code Review

### Overview

Fixes a real train↔eval input-shape mismatch: when a policy is trained with `cfg.num_cams > 2` on a multi-domain mixture and then evaluated on LIBERO (which exposes ≤2 cameras), the eval `preprocess_observation` was leaving the missing `cameraN` keys absent and `img_is_pad` all-False — so the VLM saw a strictly shorter visual prefix at eval than at any training step, and `Normalize.warn_missing_keys` fired every forward.

The fix is two-part: (1) `preprocess_observation` zero-fills missing `cameraN` slots and flips the matching `img_is_pad[:, i]=True`, mirroring `BaseDataset._standardize_images`; (2) a new `LiberoEnv.num_cams ∈ {1, 2}` field with auto-derivation of `camera_name` and consistency validation in `__post_init__`.

### Correctness

- **Shape/dtype parity with training** — verified ✅. `_standardize_images` (`src/opentau/datasets/lerobot_dataset.py:814-823`) allocates `torch.zeros((3, *self.resolution))` per missing slot and the dataloader prepends the batch dim, yielding `(B, 3, H, W)`. The PR allocates `torch.zeros((batch_size, 3, *cfg.resolution), dtype=torch.float32)` — identical shape, and the `float32` dtype matches what `ToTensor()` produces for the real cameras, so the downstream bfloat16 conversion (`utils.py:96-97`) treats both consistently.
- **Device handling** — correct. Padded zeros are created on CPU; the existing `for k, v in return_observations.items(): … v.to(device=device, …)` loop catches them. No new device-mismatch surface.
- **`LiberoEnv.camera_name` flow** — verified. `camera_name` is the actual driver of LIBERO rendering (`envs/factory.py:76` → `envs/libero.py:156, 189, 254`), so when `num_cams=1` auto-rewrites `camera_name` to `"agentview_image"`, the env really does render only one camera; `preprocess_observation` then zero-fills the rest. End-to-end flow is sound.

### Suggestions

1. **`LiberoEnv.num_cams` is functionally redundant with `camera_name`.** The field is never consumed downstream (only validated/used to derive `camera_name` in `__post_init__`). `camera_name.split(",")` already encodes the count. A single source of truth (e.g. accept `camera_name` only and derive a `num_cams` property) would eliminate the consistency-check branch and the hardcoded default-string match. Not a blocker — the current shape is convenient as a CLI knob — but worth flagging if you ever revisit this surface.

2. **Auto-derivation is tied to the *exact* default string** (`configs.py:249`):
   ```python
   if self.num_cams == 1 and self.camera_name == "agentview_image,robot0_eye_in_hand_image":
       self.camera_name = "agentview_image"
   ```
   If the 2-cam default is ever changed in the field declaration, this branch silently stops firing and `num_cams=1` will instead raise. Consider hoisting the default to a module-level constant referenced from both sites:
   ```python
   _LIBERO_DEFAULT_CAMERAS = "agentview_image,robot0_eye_in_hand_image"
   _LIBERO_AGENTVIEW_ONLY = "agentview_image"
   ...
   camera_name: str = _LIBERO_DEFAULT_CAMERAS
   ...
   if self.num_cams == 1 and self.camera_name == _LIBERO_DEFAULT_CAMERAS:
       self.camera_name = _LIBERO_AGENTVIEW_ONLY
   ```

3. **Naming collision with top-level `cfg.num_cams`.** Both `TrainPipelineConfig.num_cams` (policy input slots) and `LiberoEnv.num_cams` (real LIBERO cameras rendered) share the name. The PR description correctly calls out that they're independent, but the field name itself makes that easy to forget. `n_rendered_cams` or `num_real_cams` would be more self-documenting; if backward-compat for configs is a concern, this is fine as is, but a short callout in the field docstring (something like "this is the LIBERO sim's camera count, *not* the policy's `cfg.num_cams`") would help.

4. **No integration test for the eval rollout end-to-end.** Unit tests pin the contract well, but nothing exercises that a real LIBERO rollout with a `num_cams=4` checkpoint actually runs without the missing-key warning. Given LIBERO's heavy fixture setup, this may not be worth a new test — but if you have an existing skip-marked LIBERO integration test, threading a `num_cams=4` cfg through it would catch the integration-level regression. Not blocking.

### Nits

- `n_in_name = len([c for c in self.camera_name.split(",") if c.strip()])` — the `.strip()` filter is defensive; if you want it strict, `camera_name.split(",")` length matches `num_cams` would be enough. Either is fine.
- Test class docstrings are excellent — pin the invariant *and* explain why the train-side mirror matters. Good pattern to keep.

### Risk

- Behavior change is narrow: the eval batch now contains additional `cameraN` keys and `img_is_pad` slots. Any downstream consumer that iterated over keys expecting the old (smaller) set would see new entries — but `Normalize` / `prepare_metadata` already key off explicit names, so this should be transparent. No backwards-compat risk for `num_cams ≤ 2` checkpoints (where the new code-path is a no-op).
- `LiberoEnv` configs that previously omitted `num_cams` will get the default `2`, matching prior behavior. ✅

### CI

- check-checklist ✅, review ✅, Pre-commit ✅, CPU Tests still running at time of review.

### Verdict

Looks good to merge once CPU tests are green. The optional refactor in points (1)/(2) above is a future-cleanup item, not a blocker for this fix.


---

shuheng-liu:
## Re-Review (HEAD `aab8d94`)

The PR has been tightened since the first pass. Re-checked the current state.

### What changed since first review

- **`LiberoEnv.num_cams` removed entirely** (commit `aab8d94`). `camera_name` is now the single source of truth for both count and identity of LIBERO render cameras. The `__post_init__` is back to just the `fps`/`obs_type` checks that were there before — no new validation surface, no hardcoded default-string match, no naming collision with `cfg.num_cams`. This resolves my suggestions #1, #2, and #3 in one shot.
- **`TestLiberoEnvNumCams` removed** from `tests/envs/test_configs.py` — clean removal, no orphaned imports or references (verified `grep -n "num_cams\|LiberoEnv" tests/envs/test_configs.py` returns nothing).
- **`camera_name` docstring expanded** (`configs.py:200-209`) to cover the UX previously documented on `num_cams` — explicitly calls out that the user sets `camera_name="agentview_image"` for the single-camera ablation, and that `preprocess_observation` zero-fills remaining `cameraN` slots independently of how many real cameras LIBERO renders. Reads cleanly.

### Re-checked the actual fix

The core change in `src/opentau/envs/utils.py:81-88` is unchanged from the first review. Shape/dtype/device analysis still holds:
- `(B, 3, *cfg.resolution)` zero tensor matches `_standardize_images` after batching ✅
- `float32` dtype consistent with `ToTensor()` output; the downstream bfloat16 conversion at `utils.py:96-97` handles both ✅
- CPU allocation handled by the existing device-move loop ✅

Tests in `TestPreprocessObservationCameraPadding` are unchanged and continue to pin the contract correctly.

### Diff stats

- Before: 4 files, +193 / -3
- After:  3 files, +121 / -3

Net effect: same fix, narrower surface. Good outcome.

### Remaining items

- Integration test for the end-to-end LIBERO rollout with `cfg.num_cams > 2` is still absent (was non-blocking the first time, still non-blocking — unit tests pin the contract).
- The 2-cam `camera_name` default at `configs.py:222` still hardcodes the string, but with no auto-rewrite branch to keep in sync, this is just data, not a fragility.

### CI

- check-checklist ✅ (both runs)
- review ✅
- Pre-commit Hooks ✅ (both runs)
- CPU Tests still in progress

### Verdict

Ready to merge once CPU tests are green. All earlier blockers / soft suggestions are addressed.


---

PR #329, opened 2026-05-24
fix(logging): collapse per-dataloader sampler-save logs at checkpoint

shuheng-liu:
## What this does

When `accelerator.save_state(...)` runs, upstream `accelerate.checkpointing` emits **one INFO line per registered dataloader**:

```
INFO ... pointing.py:145 Sampler state for dataloader 0 saved in .../sampler.bin
INFO ... pointing.py:145 Sampler state for dataloader 1 saved in .../sampler_1.bin
INFO ... pointing.py:145 Sampler state for dataloader 2 saved in .../sampler_2.bin
```

A `WeightedDatasetMixture` run with many sub-datasets produces hundreds of these per checkpoint. The matching **load** path at `accelerate/checkpointing.py:278` already only emits a single summary line (`"All dataloader sampler states loaded successfully"`); this PR makes the save side as quiet:

- `init_logging` (in `src/opentau/utils/utils.py`) installs a `logging.Filter` on the `accelerate.checkpointing` logger that drops records starting with `"Sampler state for dataloader "`. Scoped to that one logger, so other accelerate save/load messages (model weights, optimizer state, scheduler state, scaler state) are unaffected.
- `scripts/train.py` adds one summary line `"Saved sampler state for {N} dataloaders to {checkpoint_dir}"` after `accelerator.save_state(...)` on the main process, mirroring the load-side summary.

No upstream behaviour or file naming changes — the `.bin` files (including the un-suffixed `sampler.bin` / `optimizer.bin` / `scheduler.bin` files that older checkpoints depend on for resume) are written exactly the same way.

For context on the un-suffixed `sampler.bin`: that's an upstream backwards-compat convention hardcoded in `accelerate/checkpointing.py:131`. Renaming the `i==0` file would break resume from older checkpoints, so this PR leaves it alone.

## How it was tested

- `pre-commit run --files src/opentau/utils/utils.py src/opentau/scripts/train.py` — all hooks pass (ruff, ruff-format, pyupgrade, bandit, etc.).
- `pytest -m "not gpu" tests/utils/test_utils_utils.py -x` — 42 passed / 4 skipped; the existing `test_init_logging_clears_previous_handlers` still passes (filter installation does not affect handler count).
- Manual filter sanity check confirmed only `"Sampler state for dataloader N saved in ..."` lines are dropped; `"Model weights saved in ..."` and the load-side `"All dataloader sampler states loaded successfully"` summary still pass through.

No policy-code changes, so no GPU/regression run needed.

```bash
gh pr checkout <this PR>
pre-commit run --files src/opentau/utils/utils.py src/opentau/scripts/train.py
pytest -m "not gpu" tests/utils/test_utils_utils.py -x
```

To see the effect end-to-end on a real training run, launch any smoke config that triggers a checkpoint save:

```bash
opentau-train --accelerate-config configs/examples/accelerate_ddp_config.yaml \
  --config_path=configs/examples/pi05_training_config.json \
  --steps=20 --save_freq=10
```

Expect zero `Sampler state for dataloader N saved` lines at checkpoint time, and exactly one `Saved sampler state for N dataloaders to ...` summary line.

## Checklist

### DIFF
--- a/src/opentau/scripts/train.py
+++ b/src/opentau/scripts/train.py
@@ -649,6 +649,9 @@ def train(cfg: TrainPipelineConfig):
 
             # save axillary objects such as configs, training step, and rng state
             if accelerator.is_main_process:
+                logging.info(
+                    f"Saved sampler state for {len(accelerator._dataloaders)} dataloaders to {checkpoint_dir}"
+                )
                 logging.info(f"Checkpoint policy after step {step}")
                 cfg.policy.pretrained_path = checkpoint_dir
                 save_checkpoint(checkpoint_dir, step, cfg)
--- a/src/opentau/scripts/train.py
+++ b/src/opentau/scripts/train.py
@@ -646,12 +646,13 @@ def train(cfg: TrainPipelineConfig):
             # save the accelerator state
             # This will save the model, optimizer, and lr_scheduler state
             accelerator.save_state(checkpoint_dir)
+            if accelerator.is_main_process:
+                # Mirrors the load-side "All dataloader sampler states loaded successfully"
+                # summary, since init_logging suppresses the per-dataloader save lines.
+                logging.info(f"Saved all dataloader sampler states to {checkpoint_dir}")
 
             # save axillary objects such as configs, training step, and rng state
             if accelerator.is_main_process:
-                logging.info(
-                    f"Saved sampler state for {len(accelerator._dataloaders)} dataloaders to {checkpoint_dir}"
-                )
                 logging.info(f"Checkpoint policy after step {step}")
                 cfg.policy.pretrained_path = checkpoint_dir
                 save_checkpoint(checkpoint_dir, step, cfg)
--- a/src/opentau/utils/utils.py
+++ b/src/opentau/utils/utils.py
@@ -193,6 +193,15 @@ def init_logging(accelerator: accelerate.Accelerator | None = None, level=loggin
 
     logging.basicConfig(level=level, force=True, handlers=[console_handler])
 
+    # accelerate's save_accelerator_state emits one INFO line per dataloader; a
+    # WeightedDatasetMixture run can have hundreds. The matching load path already
+    # only emits one summary line, so silence the save side to match.
+    class _DropDataloaderSamplerSaveSpam(logging.Filter):
+        def filter(self, record: logging.LogRecord) -> bool:
+            return not record.getMessage().startswith("Sampler state for dataloader ")
+
+    logging.getLogger("accelerate.checkpointing").addFilter(_DropDataloaderSamplerSaveSpam())
+
     if accelerator and not accelerator.is_main_process:
         # Disable duplicate logging on non-main processes
         logging.info(f"Setting logging level on non-main process {accelerator.process_index} to WARNING.")
--- a/src/opentau/utils/utils.py
+++ b/src/opentau/utils/utils.py
@@ -149,6 +149,17 @@ def is_amp_available(device: str) -> bool:
 _logging_init_stack = ""
 
 
+# accelerate's save_accelerator_state emits one INFO line per dataloader; a
+# WeightedDatasetMixture run can have hundreds. The matching load path already
+# emits a single summary line, so silence the save side to match. The exact
+# prefix string was verified at accelerate>=1.4.0 (our floor in pyproject.toml)
+# through current main; re-check on the next accelerate bump.
+class _DropDataloaderSamplerSaveSpam(logging.Filter):
+    def filter(self, record: logging.LogRecord) -> bool:
+        msg = record.getMessage()
+        return not (msg.startswith("Sampler state for dataloader ") and " saved in " in msg)
+
+
 def _format_stack(stack: list[inspect.FrameInfo]) -> str:
     return "\n".join(
         f"  File '{frame.filename}', line {frame.lineno}, in {frame.function}"
@@ -193,14 +204,9 @@ def init_logging(accelerator: accelerate.Accelerator | None = None, level=loggin
 
     logging.basicConfig(level=level, force=True, handlers=[console_handler])
 
-    # accelerate's save_accelerator_state emits one INFO line per dataloader; a
-    # WeightedDatasetMixture run can have hundreds. The matching load path already
-    # only emits one summary line, so silence the save side to match.
-    class _DropDataloaderSamplerSaveSpam(logging.Filter):
-        def filter(self, record: logging.LogRecord) -> bool:
-            return not record.getMessage().startswith("Sampler state for dataloader ")
-
-    logging.getLogger("accelerate.checkpointing").addFilter(_DropDataloaderSamplerSaveSpam())
+    accel_ckpt_logger = logging.getLogger("accelerate.checkpointing")
+    if not any(isinstance(f, _DropDataloaderSamplerSaveSpam) for f in accel_ckpt_logger.filters):
+        accel_ckpt_logger.addFilter(_DropDataloaderSamplerSaveSpam())
 
     if accelerator and not accelerator.is_main_process:
         # Disable duplicate logging on non-main processes
--- a/tests/utils/test_utils_utils.py
+++ b/tests/utils/test_utils_utils.py
@@ -200,6 +200,43 @@ def test_init_logging_clears_previous_handlers(reset_logging_state):
     assert len(logging.root.handlers) == 1  # Should still only have one handler
 
 
+def test_init_logging_drops_sampler_state_save_spam(reset_logging_state):
+    """init_logging silences the per-dataloader sampler save log from accelerate.checkpointing.
+
+    Guards against two regressions: someone removing the filter, or the predicate
+    being broadened to swallow unrelated accelerate.checkpointing log lines (in
+    particular, the load-side summary message).
+    """
+    from opentau.utils.utils import _DropDataloaderSamplerSaveSpam
+
+    init_logging()
+    accel_logger = logging.getLogger("accelerate.checkpointing")
+
+    # Idempotent install: a second init_logging() call shouldn't stack filters.
+    init_logging()
+    installed = [f for f in accel_logger.filters if isinstance(f, _DropDataloaderSamplerSaveSpam)]
+    assert len(installed) == 1
+
+    captured: list[logging.LogRecord] = []
+    sink = logging.Handler()
+    sink.emit = captured.append
+    accel_logger.addHandler(sink)
+    try:
+        accel_logger.info("Sampler state for dataloader 0 saved in /tmp/sampler.bin")
+        accel_logger.info("Sampler state for dataloader 7 saved in /tmp/sampler_7.bin")
+        accel_logger.info("Model weights saved in /tmp/model.safetensors")
+        accel_logger.info("All dataloader sampler states loaded successfully")
+    finally:
+        accel_logger.removeHandler(sink)
+        for f in installed:
+            accel_logger.removeFilter(f)
+
+    messages = [r.getMessage() for r in captured]
+    assert not any(m.startswith("Sampler state for dataloader ") for m in messages)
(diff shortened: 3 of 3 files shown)


src/opentau/utils/utils.py
```diff
@@ -193,6 +193,15 @@ def format(self, record):
 
     logging.basicConfig(level=level, force=True, handlers=[console_handler])
 
+    # accelerate's save_accelerator_state emits one INFO line per dataloader; a
+    # WeightedDatasetMixture run can have hundreds. The matching load path already
+    # only emits one summary line, so silence the save side to match.
+    class _DropDataloaderSamplerSaveSpam(logging.Filter):
+        def filter(self, record: logging.LogRecord) -> bool:
+            return not record.getMessage().startswith("Sampler state for dataloader ")
+
```
shuheng-liu: Nit: this block isn't idempotent across repeat calls to `init_logging`. The `_logging_init_stack` guard at the top only warns — it doesn't `return` — so on every subsequent call the filter is `addFilter`'d again on the `accelerate.checkpointing` logger.

In practice it's harmless (the filter is pure and stateless, so duplicates just return the same bool one more time) and the existing `test_init_logging_clears_previous_handlers` test still passes because it only inspects root-logger handler count. But if you want to keep the filter list clean — and not have the count grow forever in the long-lived test process when pytest reuses the interpreter — either:

- gate this block on `if not _logging_init_stack:` so it only runs on the first init, or
- hoist `_DropDataloaderSamplerSaveSpam` to module scope and check `any(isinstance(f, _DropDataloaderSamplerSaveSpam) for f in logger.filters)` before adding.

Both are one-liners; up to you whether it's worth it.

---

src/opentau/utils/utils.py
```diff
@@ -193,6 +193,15 @@ def format(self, record):
 
     logging.basicConfig(level=level, force=True, handlers=[console_handler])
 
+    # accelerate's save_accelerator_state emits one INFO line per dataloader; a
+    # WeightedDatasetMixture run can have hundreds. The matching load path already
+    # only emits one summary line, so silence the save side to match.
+    class _DropDataloaderSamplerSaveSpam(logging.Filter):
+        def filter(self, record: logging.LogRecord) -> bool:
+            return not record.getMessage().startswith("Sampler state for dataloader ")
```
shuheng-liu: The filter's correctness is load-bearing on a third-party log-string prefix that isn't part of accelerate's public contract — if upstream rewords it to e.g. `"Sampler {i} saved in ..."` or `"Dataloader sampler state saved..."`, this silently regresses to the spammy behavior with no test signal. I verified the exact string is unchanged between accelerate `v1.4.0` (our floor) and current `main`, and it's been there a while, so the risk is low — but please drop a one-line code comment noting the upstream version(s) where this prefix was confirmed, so the next person to bump accelerate has a breadcrumb to re-check.

While here: consider tightening the predicate to also require the canonical suffix — e.g. `record.getMessage().startswith("Sampler state for dataloader ") and " saved in " in record.getMessage()`. That way an unrelated future accelerate log line that happens to begin "Sampler state for dataloader N reset..." or similar wouldn't be swallowed accidentally.

---

src/opentau/scripts/train.py
```diff
@@ -649,6 +649,9 @@ def train(cfg: TrainPipelineConfig):
 
             # save axillary objects such as configs, training step, and rng state
             if accelerator.is_main_process:
+                logging.info(
+                    f"Saved sampler state for {len(accelerator._dataloaders)} dataloaders to {checkpoint_dir}"
```
shuheng-liu: Two concerns with `len(accelerator._dataloaders)`:

**1. Private attribute.** Upstream `accelerator.py` initializes it under a `# Internal references to the training objects` comment — no public-API contract. It's stable across `1.4.0`...`main` so the risk is low in practice, but a defensive `getattr(accelerator, "_dataloaders", None)` with a fallback to a locally-tracked count (we know it's `1 + len(per_dataset_val_dataloaders)` when val is on, else `1`) would survive an upstream rename without crashing the rank-0 save log.

**2. The count overstates what was actually saved.** Per upstream `save_accelerator_state` (`accelerate/checkpointing.py:122-138`), the `sampler*.bin` file is only written when `isinstance(dataloader.dataset, IterableDatasetShard) and isinstance(sampler, SeedableRandomSampler)`, but the per-dataloader log line at line 138 fires *unconditionally* for every dataloader in the list. So this summary correctly matches the count of suppressed log lines, but doesn't match the count of files actually on disk. For our standard `WeightedDatasetMixture` setup that's usually fine (every train/val DL has a `SeedableRandomSampler`), but it'd be worth a one-line comment so a reader who later does `ls $CKPT/sampler*.bin | wc -l` and gets a smaller number knows why.

Or sidestep both: just say `f"Saved checkpoint state to {checkpoint_dir}"` and drop the count — the per-component upstream lines (`Model weights saved in ...`, `Optimizer state saved in ...`, etc.) still fire and provide evidence the save happened.

Style nit while here: the new line sits under the `# save axillary objects such as configs, training step, and rng state` comment, but it's actually summarizing what `accelerator.save_state(...)` did three lines up. Moving it directly under the `accelerator.save_state(checkpoint_dir)` call (inside its own `if accelerator.is_main_process:` guard) would group cause and summary together.

---

shuheng-liu (commented):
Approach is correct and minimal — the filter is properly scoped to the `accelerate.checkpointing` logger (so model-weights / optimizer / scheduler / scaler / load-side summary lines all still pass through), and the targeted prefix matches the exact upstream log format verified at both `accelerate==1.4.0` (the floor in `pyproject.toml`) and current `main`. Two narrow risks plus one test gap worth addressing before un-drafting:

**Inline findings** (see comments):
1. The filter is re-installed on every `init_logging` call (the `_logging_init_stack` guard above only warns, doesn't return) — harmless but tidiness nit.
2. The filter's correctness depends on a third-party log-string prefix that isn't a public contract — add a code comment naming the accelerate version(s) confirmed, optionally tighten the predicate to also require `" saved in "`.
3. `accelerator._dataloaders` is a private attribute and the number returned counts *registered* dataloaders, not *sampler files actually written* (upstream only writes `sampler*.bin` when `IterableDatasetShard + SeedableRandomSampler`; the per-DL log line fires unconditionally). The count matches the suppressed log spam, not the disk state — worth a comment, or sidestep by dropping the count from the summary.

**Test gap**: the PR description mentions a manual sanity check that only the targeted lines are dropped, but there's no automated regression guard. A small unit test in `tests/utils/test_utils_utils.py` would lock in the contract — something like:

```python
def test_init_logging_drops_sampler_state_save_spam(reset_logging_state):
    init_logging()
    accel_logger = logging.getLogger("accelerate.checkpointing")
    captured: list[logging.LogRecord] = []
    handler = logging.Handler()
    handler.emit = captured.append
    accel_logger.addHandler(handler)
    try:
        accel_logger.info("Sampler state for dataloader 0 saved in /tmp/sampler.bin")
        accel_logger.info("Sampler state for dataloader 7 saved in /tmp/sampler_7.bin")
        accel_logger.info("Model weights saved in /tmp/model.safetensors")
        accel_logger.info("All dataloader sampler states loaded successfully")
    finally:
        accel_logger.removeHandler(handler)
    messages = [r.getMessage() for r in captured]
    assert not any(m.startswith("Sampler state for dataloader ") for m in messages)
    assert "Model weights saved in /tmp/model.safetensors" in messages
    assert "All dataloader sampler states loaded successfully" in messages
```

That catches both (a) someone refactoring `init_logging` and accidentally removing the filter and (b) the targeting being too broad and swallowing the load-side summary.

Otherwise LGTM — small, surgical, no behavior change beyond the intended log noise reduction.

---

shuheng-liu:
Thanks for the review — addressed all four points in a39a099:

**1. Idempotent filter install** — hoisted `_DropDataloaderSamplerSaveSpam` to module scope and gated the `addFilter` with `any(isinstance(f, _DropDataloaderSamplerSaveSpam) for f in logger.filters)` so repeat `init_logging` calls don't accumulate filters.

**2. Documented upstream-prefix dependency + tightened predicate** — added a class-level comment noting the prefix was verified at `accelerate>=1.4.0` (our pyproject floor) through current main, and tightened the predicate to `startswith("Sampler state for dataloader ") and " saved in " in msg` so an unrelated future log line beginning with the same prefix (e.g. `"Sampler state for dataloader N reset"`) won't be swallowed.

**3. Private `_dataloaders` + misleading count** — went with the sidestep: dropped the count entirely. The new summary is `f"Saved all dataloader sampler states to {checkpoint_dir}"`, which mirrors the load-side `"All dataloader sampler states loaded successfully"` phrasing without any private-API dependency or count-vs-disk-state confusion. Also moved the log directly under `accelerator.save_state(...)` so cause and summary are grouped.

**4. Test gap** — added `test_init_logging_drops_sampler_state_save_spam` in `tests/utils/test_utils_utils.py`. It verifies (a) idempotent install (only one filter instance after two `init_logging` calls), (b) the two "Sampler state for dataloader N saved in ..." lines are dropped, (c) `"Model weights saved in ..."` still passes through, and (d) the load-side `"All dataloader sampler states loaded successfully"` summary still passes through. Cleans up its own filter at teardown so it doesn't leak state.

Verified locally: pre-commit clean on all three changed files; `pytest -m "not gpu" tests/utils/test_utils_utils.py -x` → 43 passed / 4 skipped.

src/opentau/scripts/train.py
```diff
@@ -646,6 +646,10 @@ def train(cfg: TrainPipelineConfig):
             # save the accelerator state
             # This will save the model, optimizer, and lr_scheduler state
             accelerator.save_state(checkpoint_dir)
+            if accelerator.is_main_process:
+                # Mirrors the load-side "All dataloader sampler states loaded successfully"
+                # summary, since init_logging suppresses the per-dataloader save lines.
+                logging.info(f"Saved all dataloader sampler states to {checkpoint_dir}")
```
shuheng-liu, line 652: Nit (optional): this is now back-to-back with another `if accelerator.is_main_process:` block right below for the "save axillary objects" path. They could be coalesced into one guarded block, which would slightly reduce indentation churn and avoid re-evaluating `is_main_process`. The separation is mildly defensible (this one pairs with `accelerator.save_state` semantically, the next one is "now do auxiliary saves"), so feel free to leave as-is.

---

shuheng-liu (commented):
All four findings from the previous review cleanly addressed in `a39a099`:

1. ✅ **Idempotent filter install** — `_DropDataloaderSamplerSaveSpam` hoisted to module scope, `addFilter` gated by `any(isinstance(...))` check. Re-running `init_logging` won't stack filters.
2. ✅ **Upstream prefix dependency documented + predicate tightened** — class-level comment names the accelerate version range verified, and the predicate now requires `" saved in "` in addition to the prefix, so a hypothetical future `"Sampler state for dataloader N reset..."` line won't be swallowed.
3. ✅ **Private `_dataloaders` + count overstatement** — sidestepped cleanly. The summary is now `f"Saved all dataloader sampler states to {checkpoint_dir}"` (no count, no private API), placed directly under `accelerator.save_state(...)` so cause and summary are grouped. The phrasing intentionally mirrors the load-side `"All dataloader sampler states loaded successfully"` so the save/load symmetry holds even with upstream's own imprecision about which dataloaders actually had samplers written.
4. ✅ **Regression test added** — `test_init_logging_drops_sampler_state_save_spam` covers idempotent install, targeted suppression of the two spammy lines, and pass-through of both `"Model weights saved..."` and the load-side summary. Cleans up its own filter/handler in `finally` so it doesn't leak state into other tests.

**CI status** (commit `a39a099`): pre-commit ✅, check-checklist ✅, claude-review ✅; CPU Tests ⏳ in progress (started 00:37 UTC). I'd wait for CPU Tests to go green before merging.

One genuine nit inline about coalescing the now-adjacent `if accelerator.is_main_process:` blocks, but it's entirely optional. No other concerns — ship it once CI passes.

---

PR #330, opened 2026-05-24
fix(datasets): dedup v2.0 dataset format warning across mixture loads

shuheng-liu:
## What this does

Training on a mixture of hundreds of datasets where many are still in v2.0 format floods the log with the full 10-line `V21_MESSAGE` blurb once per dataset (`logging.warning(V21_MESSAGE...)` in `check_version_compatibility`, called from every `LeRobotDatasetMetadata.load_metadata()`). With ~50 v2.0 repos in a mixture, that's ~500 lines of nearly identical warning text.

This change deduplicates by `repo_id` at the `check_version_compatibility` call site in `src/opentau/datasets/utils.py`:

- The **first** v2.0 repo still emits the full `V21_MESSAGE` so the user sees the convert command and the help link.
- Each **subsequent unique** v2.0 repo emits a single-line warning that names the repo and the exact `convert_dataset_v20_to_v21.py --repo-id=<repo>` command to run.
- The **same** `repo_id` re-checked (e.g. reloaded) stays silent.

For a mixture with 50 v2.0 repos this drops the log from ~500 lines to ~59 while still surfacing every affected repo by name. v2.x → v1.x backward-compat *errors* (`BackwardCompatibilityError`) and the v2.x → v2.x+1 dedup logic are unchanged otherwise.

Bug.

## How it was tested

- Added `test_v21_warning_dedup` in `tests/datasets/test_utils.py` that calls `check_version_compatibility` four times (three unique repos + one duplicate), asserts exactly 3 warnings, that the first is the long-form `V21_MESSAGE` (contains the Discord help link), and that the subsequent two are one-liners naming the convert command.
- `pre-commit run --files src/opentau/datasets/utils.py tests/datasets/test_utils.py` clean.
- `pytest tests/datasets/test_utils.py -m "not gpu"` — 4 passed.

```bash
pytest -sx tests/datasets/test_utils.py::test_v21_warning_dedup
```

Or exercise it end-to-end by loading a mixture containing several `physical-intelligence/*` v2.0 datasets and inspecting the warning lines emitted at the start of training — only the first should be the long form.

## Checklist

### DIFF
--- a/src/opentau/datasets/utils.py
+++ b/src/opentau/datasets/utils.py
@@ -602,6 +602,33 @@ def is_valid_version(version: str) -> bool:
         return False
 
 
+_V21_WARNED_REPOS: set[str] = set()
+_V21_FULL_MESSAGE_SHOWN: bool = False
+
+
+def _warn_v21_global_stats(repo_id: str, version: packaging.version.Version) -> None:
+    """Emit the global-stats upgrade warning, deduplicated across the process.
+
+    The first v2.0 repo triggers the full V21_MESSAGE so the user sees how to
+    convert. Subsequent unique repo_ids only get a one-liner naming the repo,
+    so a mixture with hundreds of v2.0 datasets doesn't drown the log.
+    Re-warns for the same repo_id are suppressed entirely.
+    """
+    global _V21_FULL_MESSAGE_SHOWN
+    if repo_id in _V21_WARNED_REPOS:
+        return
+    _V21_WARNED_REPOS.add(repo_id)
+    if not _V21_FULL_MESSAGE_SHOWN:
+        logging.warning(V21_MESSAGE.format(repo_id=repo_id, version=version))
+        _V21_FULL_MESSAGE_SHOWN = True
+    else:
+        logging.warning(
+            f"Dataset {repo_id} is in v{version} format (uses global stats); "
+            f"run `python src/opentau/datasets/v21/convert_dataset_v20_to_v21.py "
+            f"--repo-id={repo_id}` to upgrade. See the first v2.0 warning above for details."
+        )
+
+
 def check_version_compatibility(
     repo_id: str,
     version_to_check: str | packaging.version.Version,
@@ -633,7 +660,7 @@ def check_version_compatibility(
     if v_check.major < v_current.major and enforce_breaking_major:
         raise BackwardCompatibilityError(repo_id, v_check)
     elif v_check.minor < v_current.minor:
-        logging.warning(V21_MESSAGE.format(repo_id=repo_id, version=v_check))
+        _warn_v21_global_stats(repo_id, v_check)
 
 
 def get_repo_versions(repo_id: str) -> list[packaging.version.Version]:
--- a/tests/datasets/test_utils.py
+++ b/tests/datasets/test_utils.py
@@ -17,6 +17,7 @@
 
 import logging
 
+import pytest
 import torch
 from datasets import Dataset
 from huggingface_hub import DatasetCard
@@ -63,11 +64,18 @@ def test_calculate_episode_data_index():
     assert torch.equal(episode_data_index["to"], torch.tensor([2, 3, 6]))
 
 
-def test_v21_warning_dedup(caplog):
+@pytest.fixture
+def reset_v21_warning_state():
+    """Reset module-global v2.0 warning dedup state around a test so a failed
+    assert doesn't leak polluted state to later tests in the same worker."""
     datasets_utils._V21_WARNED_REPOS.clear()
     datasets_utils._V21_FULL_MESSAGE_SHOWN = False
-    caplog.set_level(logging.WARNING, logger=datasets_utils.__name__)
+    yield
+    datasets_utils._V21_WARNED_REPOS.clear()
+    datasets_utils._V21_FULL_MESSAGE_SHOWN = False
 
+
+def test_v21_warning_dedup(caplog, reset_v21_warning_state):
     with caplog.at_level(logging.WARNING):
         check_version_compatibility("org/first", "2.0", "2.1")
         check_version_compatibility("org/second", "2.0", "2.1")
@@ -87,6 +95,3 @@ def test_v21_warning_dedup(caplog):
         assert repo_id in msg
         assert "convert_dataset_v20_to_v21.py" in msg
         assert "discord.com/invite" not in msg
-
-    datasets_utils._V21_WARNED_REPOS.clear()
-    datasets_utils._V21_FULL_MESSAGE_SHOWN = False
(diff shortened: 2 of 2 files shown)


tests/datasets/test_utils.py
```diff
@@ -54,3 +61,32 @@ def test_calculate_episode_data_index():
...
+    assert "discord.com/invite" in warning_messages[0]
+
+    # Subsequent messages are the one-liner naming the repo and the convert script.
+    for repo_id, msg in zip(["org/second", "org/third"], warning_messages[1:], strict=True):
+        assert repo_id in msg
+        assert "convert_dataset_v20_to_v21.py" in msg
+        assert "discord.com/invite" not in msg
+
+    datasets_utils._V21_WARNED_REPOS.clear()
+    datasets_utils._V21_FULL_MESSAGE_SHOWN = False
```
shuheng-liu: The reset at the bottom (lines 91-92) is not guaranteed to run — if any of the asserts above raise, `_V21_WARNED_REPOS` and `_V21_FULL_MESSAGE_SHOWN` stay polluted for the rest of the pytest worker process. Any later test in the same worker that exercises `check_version_compatibility` against a real v2.0 repo would then silently get the short-form one-liner (or no warning at all for `org/first` / `org/second` / `org/third`) instead of the long V21_MESSAGE.

Cleaner as an autouse-style fixture so cleanup runs even on failure:

```python
@pytest.fixture
def reset_v21_warning_state():
    datasets_utils._V21_WARNED_REPOS.clear()
    datasets_utils._V21_FULL_MESSAGE_SHOWN = False
    yield
    datasets_utils._V21_WARNED_REPOS.clear()
    datasets_utils._V21_FULL_MESSAGE_SHOWN = False


def test_v21_warning_dedup(caplog, reset_v21_warning_state):
    ...
```

Or, minimally, wrap the body in `try: ... finally:`.

---

tests/datasets/test_utils.py
```diff
@@ -54,3 +61,32 @@ def test_calculate_episode_data_index():
     episode_data_index = calculate_episode_data_index(dataset)
     assert torch.equal(episode_data_index["from"], torch.tensor([0, 2, 3]))
     assert torch.equal(episode_data_index["to"], torch.tensor([2, 3, 6]))
+
+
+def test_v21_warning_dedup(caplog):
+    datasets_utils._V21_WARNED_REPOS.clear()
+    datasets_utils._V21_FULL_MESSAGE_SHOWN = False
+    caplog.set_level(logging.WARNING, logger=datasets_utils.__name__)
```
shuheng-liu: Nit: this `set_level(..., logger=datasets_utils.__name__)` is a no-op. `_warn_v21_global_stats` calls module-level `logging.warning(...)` (which always dispatches on the root logger), not `logging.getLogger("opentau.datasets.utils").warning(...)` — so adjusting the level on the `opentau.datasets.utils` logger doesn't affect what `caplog` captures here. The `with caplog.at_level(logging.WARNING):` below (which targets the root logger) is what actually enables capture.

Either drop this line, or — if you want the warning to be filterable per-module — switch `_warn_v21_global_stats` to a module logger (`logger = logging.getLogger(__name__); logger.warning(...)`). The existing two `logging.warning(...)` sites in `utils.py` use the root logger too, so leaving the source alone and just deleting this line is the lower-churn option.

---

src/opentau/datasets/utils.py
```diff
@@ -602,6 +602,33 @@ def is_valid_version(version: str) -> bool:
         return False
 
 
+_V21_WARNED_REPOS: set[str] = set()
+_V21_FULL_MESSAGE_SHOWN: bool = False
```
shuheng-liu, line 606: Worth a heads-up (not a blocker, the PR is intentionally narrow): this dedup is *process-local*. Under multi-rank distributed training (`accelerate launch` with `num_processes>1`), every rank has its own `_V21_WARNED_REPOS` / `_V21_FULL_MESSAGE_SHOWN`, and non-main ranks aren't silenced for WARNING (`utils/utils.py:199` sets non-main ranks to `WARNING`, not `ERROR`). So with 8 ranks × 50 v2.0 repos, you'd still see 8 full V21_MESSAGE blocks + 392 one-liners. That's a ~10× reduction vs. the status quo, but the "~500 → ~59 lines" claim in the PR description holds only for single-process loads.

If you care, the most surgical fix is to skip the warning entirely on non-rank-0:

```python
from opentau.utils.accelerate_utils import accelerator_state  # or equivalent

if accelerator_state is not None and not accelerator_state.is_main_process:
    return
```

but that would couple `datasets/utils.py` to accelerate-aware state, so I'd punt unless it actually bites in a real training run.

---

shuheng-liu (commented):
Focused QoL fix and the approach is right — "full message once, one-liner per subsequent unique repo, silent on duplicates" is the right shape for this. The dedup keeps the *actionable* info (every affected `repo_id` is named with its exact convert command) while killing the boilerplate, and `BackwardCompatibilityError` semantics are correctly left alone (only the soft minor-version warning is dedup'd).

Two small things on the test (cleanup-on-failure, and a `caplog.set_level` line that's a no-op against the root-logger `logging.warning(...)` call), plus one out-of-scope heads-up about multi-rank behavior — see inline. None of these block; the cleanup-on-failure one is the only one I'd actually push to fix before merge.

---

shuheng-liu:
Addressed both actionable items in 9b9eee8:

1. **Cleanup-on-failure** — moved the dedup-state reset into a `reset_v21_warning_state` pytest fixture (per the suggested shape), so teardown runs even if an assert raises.
2. **No-op `caplog.set_level(..., logger=...)`** — removed. As noted, `_warn_v21_global_stats` uses module-level `logging.warning(...)` (root logger), so the `at_level(logging.WARNING)` context manager is what actually enables capture; the per-logger line was dead. Took the lower-churn path and left the source's logger style alone.

On the **process-local dedup under multi-rank** heads-up: punting per your own recommendation. The PR description's "~500 → ~59" figure is for single-process loads, which is the common case for dataset preview / debug / `opentau-dataset-viz`. Under N-rank training the practical reduction is `(N × 10 + (M-N))` instead of `(M × 10)` lines for M v2.0 repos — still a large win, just not as dramatic. Happy to add a `is_main_process` guard in a follow-up if it ends up biting in a real run; not worth coupling `datasets/utils.py` to accelerate state pre-emptively.

shuheng-liu (commented):
Both inline asks from the previous pass are addressed cleanly:

- **Test cleanup** — `reset_v21_warning_state` fixture with yield-based teardown guarantees the reset runs even on assertion failure. ✅
- **Misleading `caplog.set_level`** — dropped; the `with caplog.at_level(WARNING):` block alone is correct against the root-logger `logging.warning(...)` calls. ✅

Multi-rank heads-up was intentionally out of scope, agreed — no change needed for this PR.

Pre-commit is green; CPU tests still running but the deltas are localized and I'd expect them to pass. No further findings.

---

PR #331, opened 2026-05-24
chore: filter spurious pytorch 2.10 pin_memory deprecation warning

shuheng-liu:
## What this does
Silences a spurious `DeprecationWarning` emitted by PyTorch 2.10's own `DataLoader` pipeline:

```
torch/utils/data/_utils/pin_memory.py:57: DeprecationWarning:
The argument 'device' of Tensor.pin_memory() is deprecated. Please do not pass this argument.
(Triggered internally at /pytorch/aten/src/ATen/native/Memory.cpp:46.)
```

**Root cause is purely upstream PyTorch — none of OpenTau's `pin_memory=...` call sites are at fault:**

- `torch.utils.data.dataloader._BaseDataLoaderIter.__init__` overrides `_pin_memory_device` to `torch.accelerator.current_accelerator().type` (`"cuda"`) whenever `pin_memory=True` and an accelerator is available — *regardless* of the value the user passed.
- `_next_data` threads that device string through `torch.utils.data._utils.pin_memory.pin_memory(data, "cuda")`, which calls `data.pin_memory("cuda")` at `pin_memory.py:57`.
- PyTorch's own C++ then deprecates that `device` kwarg in the same release (`Tensor.pin_memory(device)` and `Tensor.is_pinned(device)`). PyTorch warns about a kwarg only PyTorch is passing.

The warning shows up 32× (16 × `pin_memory()` + 16 × `is_pinned()`) in long training logs, since the pin_memory thread fires it once per unique source location per thread/rank combination. Cosmetic; expected to be fixed upstream in PyTorch 2.10.x or 2.11.

**Fix:** a targeted `warnings.filterwarnings("ignore", ...)` in `src/opentau/__init__.py` keyed to **message regex + module regex + `DeprecationWarning` category**, so we don't accidentally suppress unrelated deprecations from elsewhere. Placed next to the existing `transformers_patch` import so every entry point (training, eval, profilers, tests) installs it on `import opentau`. No `torch.__version__` gate is needed: the filter never matches on torch < 2.10 (warning doesn't exist), and stops matching automatically once upstream fixes its DataLoader.

Considered a monkey-patch on `torch.utils.data._utils.pin_memory.pin_memory` instead, but the warning filter is less invasive and auto-deactivates.

Label: `bug` (noisy log output → fixed).

## How it was tested
**Local canary** (precision check — verifies the filter catches the exact attribution PyTorch produces, while leaving unrelated deprecations visible):

```bash
python -c "
import warnings
import opentau  # installs the filter
warnings.warn('CANARY unrelated DeprecationWarning', DeprecationWarning)
import torch.utils.data._utils.pin_memory as _pm
warnings.warn_explicit(
    \"The argument 'device' of Tensor.pin_memory() is deprecated. Please do not pass this argument.\",
    DeprecationWarning, _pm.__file__, 57, module=_pm.__name__,
)
warnings.warn_explicit(
    \"The argument 'device' of Tensor.is_pinned() is deprecated. Please do not pass this argument.\",
    DeprecationWarning, _pm.__file__, 57, module=_pm.__name__,
)
"
```

Output: only the CANARY warning is printed; both simulated torch warnings are filtered. ✓

**CPU tests** pass:

```
pytest -m "not gpu" -n auto
# 1283 passed, 14 skipped, 187 warnings in 112.99s
```

(Two pre-existing libero-import collection errors, unrelated — the local venv didn't include the `libero` extra.)

**Pre-commit** passes on the changed file.

Not tested: a full GPU smoke run with the warning-grep step from the plan. The canary above already proves the filter matches the exact warning fingerprint that appeared in the original log, so an end-to-end run is unlikely to add information.

```bash
git fetch origin claude/gallant-cartwright-6c11b6 && git checkout claude/gallant-cartwright-6c11b6

python -c "
import warnings; warnings.simplefilter('always')
import opentau
import torch.utils.data._utils.pin_memory as _pm
warnings.warn_explicit(
    \"The argument 'device' of Tensor.pin_memory() is deprecated. Please do not pass this argument.\",
    DeprecationWarning, _pm.__file__, 57, module=_pm.__name__,
)
print('done — no torch pin_memory warning should appear above')
"
```

Or run any GPU training with `pin_memory=True` and grep the log: `grep -c \"argument 'device' of Tensor\" <log>` should be 0.

## Checklist

### DIFF
--- a/src/opentau/__init__.py
+++ b/src/opentau/__init__.py
@@ -68,7 +68,7 @@ warnings.filterwarnings(
     "ignore",
     message=r"The argument 'device' of Tensor\.(pin_memory|is_pinned)\(\) is deprecated",
     category=DeprecationWarning,
-    module=r"torch\.utils\.data\._utils\.pin_memory",
+    module=r"torch\.utils\.data\._utils\.pin_memory$",
 )
 
 # TODO(rcadene): Improve policies and envs. As of now, an item in `available_policies`
--- a/src/opentau/__init__.py
+++ b/src/opentau/__init__.py
@@ -54,9 +54,22 @@ When implementing a new policy class (e.g., `DiffusionPolicy`), follow these ste
 """
 
 import itertools
+import warnings
 
 from opentau.__version__ import __version__  # noqa: F401
 from opentau.utils import transformers_patch  # noqa: F401
+
+# PyTorch 2.10 DataLoader unconditionally passes the deprecated `device` kwarg
+# to Tensor.pin_memory() / Tensor.is_pinned() via
+# torch.utils.data._utils.pin_memory.pin_memory; warning is emitted by torch
+# itself, not by any user code. Filter is harmless on older torch and becomes
+# a no-op once the upstream DataLoader stops passing the kwarg.
+warnings.filterwarnings(
+    "ignore",
+    message=r"The argument 'device' of Tensor\.(pin_memory|is_pinned)\(\) is deprecated",
+    category=DeprecationWarning,
+    module=r"torch\.utils\.data\._utils\.pin_memory",
+)
 
 # TODO(rcadene): Improve policies and envs. As of now, an item in `available_policies`
 # refers to a yaml file AND a modeling name. Same for `available_envs` which refers to


shuheng-liu:
## Review

Clean, well-scoped fix. The root-cause writeup in the description is solid and the filter is tightly keyed (message regex + module regex + `DeprecationWarning` category), which makes accidental over-suppression very unlikely.

### Verified end-to-end against the real warning path

The PR description's canary uses `warn_explicit(..., module=_pm.__name__, ...)`, which explicitly sets the attribution — so it proves the filter *matches when attribution is correct*, but not that PyTorch's C++ deprecation actually attributes there. I ran the real path on torch `2.10.0+cu128` to close that gap:

```python
import warnings; warnings.simplefilter("always")
import opentau
import torch
import torch.utils.data._utils.pin_memory as _pm
t = torch.zeros(1)
try:
    _pm.pin_memory(t, "cuda")  # the exact call DataLoader makes
except RuntimeError:
    pass  # no NVIDIA driver on this box; kwarg-deprecation already fired
warnings.warn("CANARY unrelated deprecation", DeprecationWarning)
```

Without `import opentau`: both `pin_memory` and `is_pinned` deprecations print, attributed to `torch/utils/data/_utils/pin_memory.py:57` (confirms the `module` regex is correct).

With `import opentau`: both PyTorch deprecations are silenced; the CANARY still prints. ✅

So the only gap noted under "Not tested" is actually closed — the C++ `PyErr_WarnEx` stack-frame inference resolves to `torch.utils.data._utils.pin_memory` exactly as the regex expects, even without an NVIDIA driver. The exception path is fine because PyTorch emits the kwarg-deprecation before the device check.

### Strengths

- **Self-deactivating.** Once upstream fixes the DataLoader, the filter stops matching — no `torch.__version__` gate, no follow-up cleanup PR.
- **Placement.** Putting it next to the `transformers_patch` import in `src/opentau/__init__.py` means every entry point (training, eval, profilers, tests, gRPC server) installs the filter on `import opentau`.
- **No collateral damage.** Grep confirms OpenTau never calls `tensor.pin_memory(device)` directly anywhere — all our sites use `pin_memory=True/False` as a `DataLoader` kwarg. So this filter can't accidentally silence an OpenTau-side misuse.

### Minor nit (non-blocking)

The `module` regex `r"torch\.utils\.data\._utils\.pin_memory"` is anchored only at the start (filters use `re.match`). Fine today because `pin_memory` is a leaf module, but a trailing `$` would be marginally more defensive if PyTorch ever splits this into a package. Not worth a re-push for this alone.

### Verdict

Approve — ship it once CPU tests come back green.

---

shuheng-liu:
Addressed the trailing-anchor nit in 0a6409b — `module=r"torch\.utils\.data\._utils\.pin_memory$"`. Thanks for closing the attribution gap with the real `_pm.pin_memory(t, "cuda")` call; that's a better verification than the `warn_explicit` canary I had in the PR body.

Flipping to ready.

shuheng-liu:
## Re-review (after `0a6409b`)

Re-ran the real-warning verification on the updated branch with the `$`-anchored regex (`torch\.utils\.data\._utils\.pin_memory$`):

```
$ python -c "import warnings; warnings.simplefilter('always'); import opentau; import torch, torch.utils.data._utils.pin_memory as _pm; \
    try: _pm.pin_memory(torch.zeros(1), 'cuda')
    except RuntimeError: pass
    warnings.warn('CANARY', DeprecationWarning)"
<string>:1: DeprecationWarning: CANARY
```

Both torch deprecations still silenced; CANARY still prints. ✅ Nit addressed correctly — `re.match` against `torch.utils.data._utils.pin_memory` matches the full string, and an end-anchor doesn't break that.

CI status: pre-commit ✅, check-checklist ✅, review ✅. CPU Tests still in_progress on the new commit — `mergeable_state` is `blocked` only on that. Nothing else to flag from my side; LGTM once green.

---

PR #83, opened 2026-01-22
Updating dependencies to use torch 2.8

akshay18iitg:
## What this does
Updating dependencies to support torch 2.7.1 and above

Key Updates:
Updated torch and torchvision versions
Fixed num_workers in pi05 training config

## How it was tested
Ran finetunning on oracle to check if performance is retained 

## Checklist

- [x] I have added Google-style docstrings to important functions and ensured function parameters are typed.
- [ ] My PR includes policy-related changes.
  - [ ] If the above is checked: I have run the GPU pytests (pytest -m "gpu") and regression tests.

### Note: Before submitting this PR, please read the [contributor guideline](https://github.com/TensorAuto/OpenTau/blob/main/CONTRIBUTING.md).

Closes  #74

PR #332, opened 2026-05-24
fix(datasets): switch default video decoder to torchcodec (bump 0.4->0.10)

shuheng-liu:
## What this does

Switches the default video-decoding backend from `pyav`-via-`torchvision.io.VideoReader` (deprecated since `torchvision` 0.22) to `torchcodec`, and bumps the `torchcodec` pin from `>=0.4.0,<0.5.0` to `>=0.10.0,<0.11.0` so the binary actually loads against the project's pinned `torch>=2.7.1` (current resolved: 2.10.0).

**Behavior change to flag explicitly:** on the currently-locked `torch==2.10.0`, the pre-PR `get_safe_default_codec` hard-returned `"pyav"`; post-PR it returns `"torchcodec"` whenever the compiled extension loads. This is intentional — it's the point of the PR. The two decoders pick frames using non-equivalent logic (`torchvision` seeks by pts and matches via `torch.cdist`+`argmin`; `torchcodec` indexes by `round(ts * average_fps)`), so the same `(episode, timestamp)` tuple **can** map to a different physical frame on variable-fps videos or videos whose first frame's `pts != 0`. Both stay within the configured `tolerance_s`, but a checkpoint resumed from a pre-PR run will see slightly different input pixels for the same indices. This is migration-equivalence, not a determinism property — two seeded runs of the post-PR code still produce bit-identical loss series.

**Before:** every `decode_video_frames(...)` call routed through `torchvision.io.VideoReader`, which calls `_raise_video_deprecation_warning()` on each instantiation. Training logs accumulated tens of thousands of copies of:

```
torchvision/io/_video_deprecation_warning.py:9: UserWarning: The video
decoding and encoding capabilities of torchvision are deprecated from
version 0.22 and will be removed in version 0.24. ... migrate to TorchCodec ...
```

A 6.5 MB training log from one recent run was almost entirely this warning.

**Why the previous `torchcodec` pin wasn't load-bearing:** PR #83 added `if torch>=2.8.0: return "pyav"` to `get_safe_default_codec` without bumping `torchcodec`. `torchcodec` 0.4.x was built against `torch` 2.7's C++ ABI, so `libtorchcodec_decoder*.so` fails to load on `torch` 2.10 even when the Python module imports — verified with `from torchcodec.decoders import VideoDecoder` raising `RuntimeError: Could not load libtorchcodec ... PyTorch version (2.10.0+cu128) is not compatible with this version of TorchCodec`.

**Changes in this PR:**

1. **`pyproject.toml`** — bump `torchcodec>=0.4.0,<0.5.0` to `torchcodec>=0.10.0,<0.11.0`. Per the upstream compatibility table, `torchcodec` 0.10 targets `torch` 2.10 specifically and has both `manylinux_2_28_x86_64` and `macosx_14_0_arm64` wheels for Python 3.10. The platform marker (excludes win32, arm/armv7 linux, x86_64 darwin) is kept verbatim. `uv.lock` regenerated accordingly (sole package change: `torchcodec 0.4.0 -> 0.10.0`).

2. **`src/opentau/datasets/video_utils.py`** — replaced the `torch>=2.8.0` version-check kludge with a real probe (`_load_torchcodec_videodecoder`) that tries to import `torchcodec.decoders.VideoDecoder` and catches `ImportError` / `RuntimeError` / `ValueError` (the latter for `find_spec` itself raising on broken installs). Helper is shared with the dispatcher in `decode_video_frames`, so explicit `backend="torchcodec"` callers also fall back to `pyav` (with a clear warning naming the original choice) instead of crashing at first frame decode. Cached via `@functools.cache` for thread-safe, once-per-process probing with a public `.cache_clear()` reset hook. Explicit `backend="pyav"` still works for callers that opt in; they inherit the deprecation warning, which is the correct upstream signal.

**System FFmpeg requirement.** `torchcodec` requires a system FFmpeg installation (it does not ship a self-contained wheel). On hosts without one, the new probe falls back to `pyav` and the deprecation warning will still appear. The `pyav` library bundles its own FFmpeg, which is why the current production pipeline kept working on bare-bones hosts. Production hosts running this code should make sure FFmpeg 4+ is available in the runtime environment (e.g. inside the container image / on the runtime system) to fully benefit from this change.

## How it was tested

- `pre-commit run --files src/opentau/datasets/video_utils.py pyproject.toml uv.lock` — passes.
- `pytest -m "not gpu" -n auto` — passes apart from pre-existing `tests/envs/test_libero_*` and `tests/utils/test_libero_utils.py::test_libero2torch` failures unrelated to this change (missing LIBERO assets in the local dev environment).
- `pytest -sx -m "not gpu" tests/datasets/test_datasets.py::test_deferred_video_attach_video_start_time` — passes, exercising the new `torchcodec` default path end-to-end.
- Confirmed zero `_video_deprecation_warning` occurrences in dataset-test logs after the change.
- Smoke training run on an internal RTX 5090 dev box with `configs/dev/ci_config.json` (pi05, 25 steps, batch=4, dataset `lerobot/droid_100` which has 3 `video` features). Result: 25/25 steps completed, `mse_loss` trended down from ~1.0 to ~0.1, **zero `_video_deprecation_warning` / `failed to load torchcodec` / `not available` occurrences** in the full training log.
- Confirmed the slurm training environment ships FFmpeg 4 + `libavcodec.so.58` inside the container image — `torchcodec` 0.10's bundled `libtorchcodec_decoder4.so` matches that ABI, so the bumped wheel should load there too.

```bash
gh pr checkout <this-PR>
pytest -sx -m "not gpu" tests/datasets/test_datasets.py::test_deferred_video_attach_video_start_time
```

Then a smoke training run against any video-feature dataset:

```bash
accelerate launch --num_processes 1 src/opentau/scripts/train.py \
    --config_path=configs/dev/ci_config.json \
    --wandb.enable=false --steps=25 --output_dir=/tmp/torchcodec-smoke
# verify zero deprecation warnings in the log:
grep -c "_video_deprecation_warning" <log>   # expect 0
```

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -66,7 +66,7 @@ dependencies = [
     "rerun-sdk>=0.21.0",
     "termcolor>=2.4.0",
     "torch>=2.7.1",
-    "torchcodec>=0.4.0, <0.5.0; sys_platform != 'win32' and (sys_platform != 'linux' or (platform_machine != 'aarch64' and platform_machine != 'arm64' and platform_machine != 'armv7l')) and (sys_platform != 'darwin' or platform_machine != 'x86_64')",
+    "torchcodec>=0.10.0, <0.11.0; sys_platform != 'win32' and (sys_platform != 'linux' or (platform_machine != 'aarch64' and platform_machine != 'arm64' and platform_machine != 'armv7l')) and (sys_platform != 'darwin' or platform_machine != 'x86_64')",
     "torchvision>=0.22.1",
     "wandb>=0.27.0",
     "zarr>=2.17.0",
--- a/src/opentau/datasets/video_utils.py
+++ b/src/opentau/datasets/video_utils.py
@@ -110,27 +110,39 @@ import pyarrow as pa
 import torch
 import torchvision
 from datasets.features.features import register_feature
-from packaging import version
 from PIL import Image
+
+_safe_default_codec: str | None = None
 
 
 def get_safe_default_codec() -> str:
-    """Get the default video codec backend, falling back to pyav if torchcodec is unavailable.
+    """Get the default video codec backend.
 
-    Returns:
-        Backend name: "torchcodec" if available, otherwise "pyav".
+    Prefers ``torchcodec`` when both the Python package is installed AND its
+    compiled extension can actually load (which requires a compatible system
+    FFmpeg). Falls back to ``"pyav"`` otherwise. Result is cached at module
+    level so the load attempt only happens once per process.
     """
-
-    if version.parse(torch.__version__) >= version.parse("2.8.0"):
-        return "pyav"
-    else:
-        if importlib.util.find_spec("torchcodec"):
-            return "torchcodec"
-        else:
+    global _safe_default_codec
+    if _safe_default_codec is not None:
+        return _safe_default_codec
+    if importlib.util.find_spec("torchcodec") is not None:
+        try:
+            from torchcodec.decoders import VideoDecoder  # noqa: F401
+        except (ImportError, RuntimeError) as e:
             logging.warning(
-                "'torchcodec' is not available in your platform, falling back to 'pyav' as a default decoder"
+                "'torchcodec' failed to load (%s); falling back to 'pyav' as the default decoder.",
+                e,
             )
-            return "pyav"
+        else:
+            _safe_default_codec = "torchcodec"
+            return _safe_default_codec
+    else:
+        logging.warning(
+            "'torchcodec' is not available in your platform, falling back to 'pyav' as the default decoder."
+        )
+    _safe_default_codec = "pyav"
+    return _safe_default_codec
 
 
 _safe_encoding_vcodec: str | None = None
--- a/src/opentau/datasets/video_utils.py
+++ b/src/opentau/datasets/video_utils.py
@@ -94,6 +94,7 @@ Example:
         >>> print(f"FPS: {info['video.fps']}, Resolution: {info['video.width']}x{info['video.height']}")
 """
 
+import functools
 import importlib
 import json
 import logging
@@ -112,37 +113,45 @@ import torchvision
 from datasets.features.features import register_feature
 from PIL import Image
 
-_safe_default_codec: str | None = None
+
+@functools.cache
+def _load_torchcodec_videodecoder():
+    """Probe whether ``torchcodec.decoders.VideoDecoder`` is usable on this host.
+
+    Returns the ``VideoDecoder`` class on success, or ``None`` if either the
+    Python package is missing or its compiled extension fails to load (e.g.
+    no compatible system FFmpeg). Wide except: ``find_spec`` itself can raise
+    on broken installs, and the extension import surfaces failures as either
+    ``ImportError`` or ``RuntimeError``. Cached via ``functools.cache`` so the
+    probe runs once per process and is thread-safe; tests can re-probe with
+    ``_load_torchcodec_videodecoder.cache_clear()``.
+    """
+    try:
+        if importlib.util.find_spec("torchcodec") is None:
+            logging.warning(
+                "'torchcodec' is not available in your platform, falling back to 'pyav' as the default decoder."
+            )
+            return None
+        from torchcodec.decoders import VideoDecoder
+    except (ImportError, RuntimeError, ValueError) as e:
+        logging.warning(
+            "'torchcodec' failed to load (%s); falling back to 'pyav' as the default decoder.",
+            e,
+        )
+        return None
+    return VideoDecoder
 
 
+@functools.cache
 def get_safe_default_codec() -> str:
     """Get the default video codec backend.
 
-    Prefers ``torchcodec`` when both the Python package is installed AND its
-    compiled extension can actually load (which requires a compatible system
-    FFmpeg). Falls back to ``"pyav"`` otherwise. Result is cached at module
-    level so the load attempt only happens once per process.
+    Returns ``"torchcodec"`` when the compiled extension actually loads on this
+    host (requires a compatible system FFmpeg), else ``"pyav"``. Cached for the
+    lifetime of the process; tests can re-probe with
+    ``get_safe_default_codec.cache_clear()``.
     """
-    global _safe_default_codec
-    if _safe_default_codec is not None:
-        return _safe_default_codec
-    if importlib.util.find_spec("torchcodec") is not None:
-        try:
-            from torchcodec.decoders import VideoDecoder  # noqa: F401
-        except (ImportError, RuntimeError) as e:
-            logging.warning(
-                "'torchcodec' failed to load (%s); falling back to 'pyav' as the default decoder.",
-                e,
-            )
-        else:
-            _safe_default_codec = "torchcodec"
-            return _safe_default_codec
-    else:
-        logging.warning(
-            "'torchcodec' is not available in your platform, falling back to 'pyav' as the default decoder."
-        )
-    _safe_default_codec = "pyav"
-    return _safe_default_codec
+    return "torchcodec" if _load_torchcodec_videodecoder() is not None else "pyav"
 
 
 _safe_encoding_vcodec: str | None = None
@@ -206,11 +215,16 @@ def decode_video_frames(
     if backend is None:
         backend = get_safe_default_codec()
     if backend == "torchcodec":
-        return decode_video_frames_torchcodec(video_path, timestamps, tolerance_s)
-    elif backend in ["pyav", "video_reader"]:
+        if _load_torchcodec_videodecoder() is None:
+            logging.warning(
+                "Caller requested backend='torchcodec' but it failed to load; using 'pyav' for this decode."
+            )
+            backend = "pyav"
+        else:
+            return decode_video_frames_torchcodec(video_path, timestamps, tolerance_s)
+    if backend in ("pyav", "video_reader"):
         return decode_video_frames_torchvision(video_path, timestamps, tolerance_s, backend)
-    else:
-        raise ValueError(f"Unsupported video backend: {backend}")
+    raise ValueError(f"Unsupported video backend: {backend}")
 
 
 def decode_video_frames_torchvision(
@@ -329,10 +343,11 @@ def decode_video_frames_torchcodec(
     can be adjusted during encoding to take into account decoding time and video size in bytes.
     """
 
-    if importlib.util.find_spec("torchcodec"):
-        from torchcodec.decoders import VideoDecoder
-    else:
-        raise ImportError("torchcodec is required but not available.")
+    VideoDecoder = _load_torchcodec_videodecoder()  # noqa: N806
+    if VideoDecoder is None:
+        raise ImportError(
+            "torchcodec is required but not loadable on this host (see prior warning for details)."
+        )
 
     # initialize video decoder
     decoder = VideoDecoder(video_path, device=device, seek_mode="exact")
--- a/src/opentau/datasets/video_utils.py
+++ b/src/opentau/datasets/video_utils.py
@@ -231,8 +231,8 @@ def decode_video_frames(
     decoders pick frames using non-equivalent logic (pts seek + argmin vs.
     ``round(ts * average_fps)`` indexing), so the decoded pixels for the same
     ``(video_path, timestamps)`` pair can differ within ``tolerance_s``;
-    callers needing strict backend selection should check the return of
-    ``_load_torchcodec_videodecoder()`` before dispatching.
+    callers needing strict backend selection should check
+    ``get_safe_default_codec() == "torchcodec"`` before dispatching.
     """
     if backend is None:
         backend = get_safe_default_codec()
--- a/src/opentau/datasets/video_utils.py
+++ b/src/opentau/datasets/video_utils.py
@@ -115,43 +115,57 @@ from PIL import Image
 
 
 @functools.cache
-def _load_torchcodec_videodecoder():
+def _load_torchcodec_videodecoder() -> tuple[type | None, BaseException | None]:
     """Probe whether ``torchcodec.decoders.VideoDecoder`` is usable on this host.
 
-    Returns the ``VideoDecoder`` class on success, or ``None`` if either the
-    Python package is missing or its compiled extension fails to load (e.g.
-    no compatible system FFmpeg). Wide except: ``find_spec`` itself can raise
-    on broken installs, and the extension import surfaces failures as either
+    Returns a ``(VideoDecoder, None)`` pair on success, or ``(None, exc)`` if
+    either the Python package is missing or its compiled extension fails to
+    load (e.g. no compatible system FFmpeg). Carrying the original exception
+    forward lets direct callers of the torchcodec decode path surface the root
+    cause in their own raise, even when the probe's warning has been dropped
+    (e.g. probe fires before ``init_logging``) or rotated out of a long-lived
+    log. Wide except: ``find_spec`` can itself raise ``ValueError`` on broken
+    installs, and the extension import surfaces failures as either
     ``ImportError`` or ``RuntimeError``. Cached via ``functools.cache`` so the
     probe runs once per process and is thread-safe; tests can re-probe with
     ``_load_torchcodec_videodecoder.cache_clear()``.
     """
     try:
         if importlib.util.find_spec("torchcodec") is None:
-            logging.warning(
-                "'torchcodec' is not available in your platform, falling back to 'pyav' as the default decoder."
-            )
-            return None
+            exc = ModuleNotFoundError("torchcodec is not installed on this platform.")
+            logging.warning("%s Falling back to 'pyav' as the default decoder.", exc)
+            return None, exc
         from torchcodec.decoders import VideoDecoder
     except (ImportError, RuntimeError, ValueError) as e:
         logging.warning(
             "'torchcodec' failed to load (%s); falling back to 'pyav' as the default decoder.",
             e,
         )
-        return None
-    return VideoDecoder
+        return None, e
+    return VideoDecoder, None
 
 
 @functools.cache
+def _warn_explicit_torchcodec_unloadable() -> None:
+    """Emit the explicit-backend-fallback warning at most once per process.
+
+    Without this guard the warning fires inside ``decode_video_frames``, which
+    sits in ``LeRobotDataset._query_videos`` and is invoked O(num_video_keys)
+    times per dataloader step. A real run logs tens of thousands of copies.
+    """
+    logging.warning("Caller requested backend='torchcodec' but it failed to load; using 'pyav' instead.")
+
+
 def get_safe_default_codec() -> str:
     """Get the default video codec backend.
 
     Returns ``"torchcodec"`` when the compiled extension actually loads on this
-    host (requires a compatible system FFmpeg), else ``"pyav"``. Cached for the
-    lifetime of the process; tests can re-probe with
-    ``get_safe_default_codec.cache_clear()``.
+    host (requires a compatible system FFmpeg), else ``"pyav"``. The underlying
+    probe is cached in ``_load_torchcodec_videodecoder``; this wrapper is left
+    uncached so that ``dataclasses.field(default_factory=...)`` callers get the
+    fresh-instance semantics they expect from ``default_factory``.
     """
-    return "torchcodec" if _load_torchcodec_videodecoder() is not None else "pyav"
+    return "torchcodec" if _load_torchcodec_videodecoder()[0] is not None else "pyav"
 
 
 _safe_encoding_vcodec: str | None = None
@@ -205,20 +219,26 @@ def decode_video_frames(
         video_path (Path): Path to the video file.
         timestamps (list[float]): List of timestamps to extract frames.
         tolerance_s (float): Allowed deviation in seconds for frame retrieval.
-        backend (str, optional): Backend to use for decoding. Defaults to "torchcodec" when available in the platform; otherwise, defaults to "pyav"..
+        backend (str, optional): Backend to use for decoding. Defaults to "torchcodec" when available in the platform; otherwise, defaults to "pyav".
 
     Returns:
         torch.Tensor: Decoded frames.
 
-    Currently supports torchcodec on cpu and pyav.
+    Currently supports torchcodec on cpu and pyav. If the caller explicitly
+    requests ``backend="torchcodec"`` but the host's torchcodec extension
+    cannot load (e.g. missing system FFmpeg, ABI mismatch), this function
+    silently downgrades to ``"pyav"`` with a once-per-process warning. The two
+    decoders pick frames using non-equivalent logic (pts seek + argmin vs.
+    ``round(ts * average_fps)`` indexing), so the decoded pixels for the same
+    ``(video_path, timestamps)`` pair can differ within ``tolerance_s``;
+    callers needing strict backend selection should check the return of
+    ``_load_torchcodec_videodecoder()`` before dispatching.
     """
     if backend is None:
         backend = get_safe_default_codec()
     if backend == "torchcodec":
-        if _load_torchcodec_videodecoder() is None:
-            logging.warning(
-                "Caller requested backend='torchcodec' but it failed to load; using 'pyav' for this decode."
-            )
+        if _load_torchcodec_videodecoder()[0] is None:
+            _warn_explicit_torchcodec_unloadable()
             backend = "pyav"
         else:
             return decode_video_frames_torchcodec(video_path, timestamps, tolerance_s)
@@ -343,11 +363,11 @@ def decode_video_frames_torchcodec(
     can be adjusted during encoding to take into account decoding time and video size in bytes.
     """
 
-    VideoDecoder = _load_torchcodec_videodecoder()  # noqa: N806
+    VideoDecoder, probe_exc = _load_torchcodec_videodecoder()  # noqa: N806
     if VideoDecoder is None:
         raise ImportError(
-            "torchcodec is required but not loadable on this host (see prior warning for details)."
-        )
+            f"torchcodec is required but not loadable on this host: {probe_exc!r}"
+        ) from probe_exc
 
     # initialize video decoder
     decoder = VideoDecoder(video_path, device=device, seek_mode="exact")
--- a/uv.lock
+++ b/uv.lock
@@ -1950,162 +1950,162 @@ wheels = [
 [[package]]
 name = "nvidia-cublas-cu12"
 version = "12.8.4.1"
-source = { registry = "https://pypi.org/simple" }
+source = { registry = "https://pypi.nvidia.com/" }
 wheels = [
-    { url = "https://files.pythonhosted.org/packages/29/99/db44d685f0e257ff0e213ade1964fc459b4a690a73293220e98feb3307cf/nvidia_cublas_cu12-12.8.4.1-py3-none-manylinux_2_27_aarch64.whl", hash = "sha256:b86f6dd8935884615a0683b663891d43781b819ac4f2ba2b0c9604676af346d0", size = 590537124, upload-time = "2025-03-07T01:43:53.556Z" },
-    { url = "https://files.pythonhosted.org/packages/dc/61/e24b560ab2e2eaeb3c839129175fb330dfcfc29e5203196e5541a4c44682/nvidia_cublas_cu12-12.8.4.1-py3-none-manylinux_2_27_x86_64.whl", hash = "sha256:8ac4e771d5a348c551b2a426eda6193c19aa630236b418086020df5ba9667142", size = 594346921, upload-time = "2025-03-07T01:44:31.254Z" },
-    { url = "https://files.pythonhosted.org/packages/70/61/7d7b3c70186fb651d0fbd35b01dbfc8e755f69fd58f817f3d0f642df20c3/nvidia_cublas_cu12-12.8.4.1-py3-none-win_amd64.whl", hash = "sha256:47e9b82132fa8d2b4944e708049229601448aaad7e6f296f630f2d1a32de35af", size = 567544208, upload-time = "2025-03-07T01:53:30.535Z" },
+    { url = "https://pypi.nvidia.com/nvidia-cublas-cu12/nvidia_cublas_cu12-12.8.4.1-py3-none-manylinux_2_27_aarch64.whl", hash = "sha256:b86f6dd8935884615a0683b663891d43781b819ac4f2ba2b0c9604676af346d0" },
+    { url = "https://pypi.nvidia.com/nvidia-cublas-cu12/nvidia_cublas_cu12-12.8.4.1-py3-none-manylinux_2_27_x86_64.whl", hash = "sha256:8ac4e771d5a348c551b2a426eda6193c19aa630236b418086020df5ba9667142" },
+    { url = "https://pypi.nvidia.com/nvidia-cublas-cu12/nvidia_cublas_cu12-12.8.4.1-py3-none-win_amd64.whl", hash = "sha256:47e9b82132fa8d2b4944e708049229601448aaad7e6f296f630f2d1a32de35af" },
 ]
 
 [[package]]
 name = "nvidia-cuda-cupti-cu12"
 version = "12.8.90"
-source = { registry = "https://pypi.org/simple" }
+source = { registry = "https://pypi.nvidia.com/" }
 wheels = [
-    { url = "https://files.pythonhosted.org/packages/d5/1f/b3bd73445e5cb342727fd24fe1f7b748f690b460acadc27ea22f904502c8/nvidia_cuda_cupti_cu12-12.8.90-py3-none-manylinux2014_aarch64.manylinux_2_17_aarch64.whl", hash = "sha256:4412396548808ddfed3f17a467b104ba7751e6b58678a4b840675c56d21cf7ed", size = 9533318, upload-time = "2025-03-07T01:40:10.421Z" },
-    { url = "https://files.pythonhosted.org/packages/f8/02/2adcaa145158bf1a8295d83591d22e4103dbfd821bcaf6f3f53151ca4ffa/nvidia_cuda_cupti_cu12-12.8.90-py3-none-manylinux2014_x86_64.manylinux_2_17_x86_64.whl", hash = "sha256:ea0cb07ebda26bb9b29ba82cda34849e73c166c18162d3913575b0c9db9a6182", size = 10248621, upload-time = "2025-03-07T01:40:21.213Z" },
-    { url = "https://files.pythonhosted.org/packages/41/bc/83f5426095d93694ae39fe1311431b5d5a9bb82e48bf0dd8e19be2765942/nvidia_cuda_cupti_cu12-12.8.90-py3-none-win_amd64.whl", hash = "sha256:bb479dcdf7e6d4f8b0b01b115260399bf34154a1a2e9fe11c85c517d87efd98e", size = 7015759, upload-time = "2025-03-07T01:51:11.355Z" },
+    { url = "https://pypi.nvidia.com/nvidia-cuda-cupti-cu12/nvidia_cuda_cupti_cu12-12.8.90-py3-none-manylinux2014_aarch64.manylinux_2_17_aarch64.whl", hash = "sha256:4412396548808ddfed3f17a467b104ba7751e6b58678a4b840675c56d21cf7ed" },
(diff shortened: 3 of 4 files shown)


src/opentau/datasets/video_utils.py
```diff
@@ -110,27 +110,39 @@
...
-        if importlib.util.find_spec("torchcodec"):
-            return "torchcodec"
-        else:
+    global _safe_default_codec
+    if _safe_default_codec is not None:
+        return _safe_default_codec
+    if importlib.util.find_spec("torchcodec") is not None:
+        try:
+            from torchcodec.decoders import VideoDecoder  # noqa: F401
+        except (ImportError, RuntimeError) as e:
```
shuheng-liu: **Asymmetric error handling — explicit `backend="torchcodec"` callers still crash on broken installs.**

The new fallback only protects the auto-default path. `decode_video_frames_torchcodec` at line 332 still has an unguarded import:

```python
if importlib.util.find_spec("torchcodec"):
    from torchcodec.decoders import VideoDecoder
else:
    raise ImportError("torchcodec is required but not available.")
```

Any caller that passes `backend="torchcodec"` explicitly — including a previously-saved checkpoint config or anything that hard-codes the backend — bypasses `get_safe_default_codec` and re-raises the very `RuntimeError` this PR is meant to swallow, killing the dataloader worker mid-decode. That contradicts the PR's stated goal ("cleanly falls back to `pyav` with a logged reason instead of crashing at first frame decode").

Recommend extracting a single helper that both `get_safe_default_codec` and `decode_video_frames_torchcodec` share, e.g.:

```python
def _try_import_videodecoder():
    if importlib.util.find_spec("torchcodec") is None:
        return None
    try:
        from torchcodec.decoders import VideoDecoder
        return VideoDecoder
    except (ImportError, RuntimeError) as e:
        logging.warning("'torchcodec' failed to load (%s); falling back to 'pyav'.", e)
        return None
```

and let `decode_video_frames` redirect to `pyav` when the explicit-`torchcodec` path returns `None`.

---

src/opentau/datasets/video_utils.py
```diff
@@ -110,27 +110,39 @@
...
             )
-            return "pyav"
+        else:
+            _safe_default_codec = "torchcodec"
+            return _safe_default_codec
+    else:
+        logging.warning(
+            "'torchcodec' is not available in your platform, falling back to 'pyav' as the default decoder."
+        )
+    _safe_default_codec = "pyav"
```
shuheng-liu: **Silent backend switch from `pyav` → `torchcodec` can change which physical frame is loaded — potential bit-identical-resume regression (CLAUDE.md §3).**

Pre-PR with `torch>=2.8.0` (the lockfile pins `torch==2.10.0`), `get_safe_default_codec()` hard-returned `"pyav"`. Post-PR, on the same locked environment, it returns `"torchcodec"` whenever the `.so` loads.

The two decoders pick frames using non-equivalent logic:

- `decode_video_frames_torchvision` (line 252+) seeks by pts, iterates frames, matches via `torch.cdist`+`argmin` over decoded `pts`.
- `decode_video_frames_torchcodec` (line 346) converts timestamps via `[round(ts * average_fps) for ts in timestamps]` then `decoder.get_frames_at(indices=...)`.

For variable-fps videos, videos whose first frame's `pts != 0`, or videos with imprecise `average_fps` metadata, the two paths can select **different physical frames** for the same `timestamps` list. Both remain within `tolerance_s`, but a model trained on `pyav` frames (pre-PR) and then resumed/evaluated under the same checkpoint config on the same torch (post-PR → torchcodec) will see different input pixels for the same `(episode_idx, ts)` tuple. The PR description characterizes this as a guard removal, but it's also a default backend switch on the only torch version currently in the lock.

Could you add a short note in the PR body confirming this is intentional, and ideally a determinism check (per CLAUDE.md §3: two seeded runs producing bit-identical loss series) on at least one dataset that has video features?

---

src/opentau/datasets/video_utils.py
```diff
@@ -110,27 +110,39 @@
...
-    if version.parse(torch.__version__) >= version.parse("2.8.0"):
-        return "pyav"
-    else:
-        if importlib.util.find_spec("torchcodec"):
-            return "torchcodec"
-        else:
+    global _safe_default_codec
+    if _safe_default_codec is not None:
+        return _safe_default_codec
+    if importlib.util.find_spec("torchcodec") is not None:
```
shuheng-liu: **`find_spec` is outside the try/except, but it can itself raise.**

Per the CPython docs (`help(importlib.util.find_spec)`): *"If `sys.modules[name].__spec__` happens to be set to None, then `ValueError` is raised."* Other malformed-install conditions (partially-removed dist-info, broken parent-package init) can surface as `ImportError` from inside the finder chain — also propagating out of `find_spec` itself.

Because the call is on line 129 and the `try:` doesn't start until line 130, an exception here:
- escapes uncaught,
- never reaches the `_safe_default_codec = "pyav"` write at line 144,
- propagates from `field(default_factory=get_safe_default_codec)` in `configs/default.py:115` and crashes draccus config parsing for every CLI invocation,
- and because the cache never gets set, the next call repeats the failure.

This contradicts the new docstring's promise that "the load attempt only happens once per process". Suggest wrapping `find_spec` in the same try block (or a broader outer try) so the function can always settle on `"pyav"` no matter what fails during probing.

---

src/opentau/datasets/video_utils.py
```diff
@@ -110,27 +110,39 @@
 import torch
 import torchvision
 from datasets.features.features import register_feature
-from packaging import version
 from PIL import Image
 
+_safe_default_codec: str | None = None
```
shuheng-liu: **Module-level cache has no reset hook and `sys.modules` is permanently polluted after the first probe — test-isolation footgun + long-lived-process recovery footgun.**

Two related concerns about turning this into a process-lifetime singleton:

1. **No public way to invalidate.** `grep -rn 'safe_default_codec' tests/` returns nothing today, but the moment a test wants to assert the fallback path (e.g. `monkeypatch.setattr(importlib.util, "find_spec", lambda *_: None)`), it has to reach into the private global `video_utils._safe_default_codec = None`. Worse, after the first successful probe `sys.modules["torchcodec"]` is populated, so a `find_spec` monkeypatch can be silently bypassed by Python's import machinery. Consider exposing a `_reset_safe_default_codec()` test helper, or using `functools.cache` with `.cache_clear()`.

2. **Long-lived processes can't recover after a fix.** On a Jupyter kernel or the gRPC inference server in `src/opentau/scripts/`, a user who diagnoses "torchcodec failed to load", installs `libavcodec`, and expects subsequent dataset constructions to pick up torchcodec — won't. The cache permanently records `"pyav"` until the process restarts. Worth at least documenting in the docstring (or, again, exposing a reset hook).

Neither blocks the PR, but both are footguns the docstring's "cached once per process" promise doesn't acknowledge.

---

src/opentau/datasets/video_utils.py
```diff
@@ -110,27 +110,39 @@
...
-            return "torchcodec"
-        else:
+    global _safe_default_codec
+    if _safe_default_codec is not None:
+        return _safe_default_codec
+    if importlib.util.find_spec("torchcodec") is not None:
+        try:
+            from torchcodec.decoders import VideoDecoder  # noqa: F401
+        except (ImportError, RuntimeError) as e:
             logging.warning(
```
shuheng-liu: **Warning fires before `init_logging()` runs — silently dropped or unformatted, and emitted from every rank.**

`get_safe_default_codec` is wired as `default_factory` on `DatasetConfig.video_backend` at `src/opentau/configs/default.py:115`. Draccus invokes that factory during config parsing inside `@parser.wrap()` — which happens **before** `init_logging(accelerator, ...)` runs in `scripts/train.py`.

Consequences when torchcodec actually fails to load:
- the warning goes to the root logger with no project-configured handler, falling through to Python's `lastResort` `sys.stderr` writer with the bare `WARNING:root:...` format,
- it bypasses any rank-0 filter that `init_logging` installs, so on N-rank distributed runs you get N copies of the warning,
- per-DataLoader-worker forks each re-warm an empty cache, so on a broken-FFmpeg host you'll see `num_workers + 1` copies of the warning per process — typically 9-33 lines of noise per epoch.

Cheap mitigation: use `logger = logging.getLogger(__name__)` at module top (matches the rest of the file's `logging.warning(...)` calls? — actually I see they also use the bare root logger; you may want to fix both `get_safe_default_codec` and `get_safe_encoding_vcodec` together) and consider deferring the warning to the first actual `decode_video_frames` call instead of emitting at config-parse time.

---

pyproject.toml
```diff
@@ -66,7 +66,7 @@ dependencies = [
     "rerun-sdk>=0.21.0",
     "termcolor>=2.4.0",
     "torch>=2.7.1",
-    "torchcodec>=0.4.0, <0.5.0; sys_platform != 'win32' and (sys_platform != 'linux' or (platform_machine != 'aarch64' and platform_machine != 'arm64' and platform_machine != 'armv7l')) and (sys_platform != 'darwin' or platform_machine != 'x86_64')",
+    "torchcodec>=0.10.0, <0.11.0; sys_platform != 'win32' and (sys_platform != 'linux' or (platform_machine != 'aarch64' and platform_machine != 'arm64' and platform_machine != 'armv7l')) and (sys_platform != 'darwin' or platform_machine != 'x86_64')",
```
shuheng-liu, line 69: **Two related concerns about the new pin:**

1. **Missing torch upper bound.** torchcodec 0.10 targets torch 2.10 specifically (the PR body confirms this; lock pins `torch==2.10.0`). But `pyproject.toml` still declares unbounded `torch>=2.7.1`. Anyone running `uv lock --upgrade-package torch` (or installing into an environment that already has torch 2.11+ pre-pinned by a slurm prologue / Docker base image) will get an ABI-mismatch `RuntimeError` from `from torchcodec.decoders import VideoDecoder`. With the PR's removal of the explicit torch version guard, the only safety net is the new try/except — which silently degrades to `pyav` on the auto path (slower; deprecation warnings return) and crashes on the explicit-`torchcodec` path. Consider pairing the torchcodec pin with `torch>=2.10.0,<2.11.0` (or whatever the next-version-supports-bumping cadence allows).

2. **Windows wheel now exists but the marker still excludes it.** `torchcodec 0.10.0` ships `torchcodec-0.10.0-cp310-cp310-win_amd64.whl` (visible in `uv.lock` line ~3961; 0.4.0 did not). The platform marker on this line still excludes `sys_platform == 'win32'`, so Windows developers continue to fall back to `pyav` unnecessarily. If the 0.4→0.10 bump was meant to enable broader platform support, dropping `sys_platform != 'win32'` from the marker is the missing half. Non-blocking — flagging in case it was an oversight.

---

uv.lock
```diff
@@ -1950,162 +1950,162 @@ wheels = [
 [[package]]
 name = "nvidia-cublas-cu12"
 version = "12.8.4.1"
-source = { registry = "https://pypi.org/simple" }
+source = { registry = "https://pypi.nvidia.com/" }
```
shuheng-liu: **Unrelated lock churn: ~10 NVIDIA packages re-sourced from `pypi.org/simple` → `pypi.nvidia.com/`, plus `tensorrt-cu13-libs` flipped from sdist-only to three platform wheels (line 3795 below).**

These changes were introduced by re-locking with `UV_EXTRA_INDEX_URL=https://pypi.nvidia.com/` set (per the PR body's "Heads-up: the `UV_EXTRA_INDEX_URL` is unrelated to this change"). The lock now references a registry that is **not declared** in `pyproject.toml` under `[[tool.uv.index]]`. Concerns:

1. **CI determinism.** None of `cpu_test.yml`, `gpu_test.yml`, `regression_test.yml` exports `UV_EXTRA_INDEX_URL`. uv does appear to accept the locked wheel URLs at install time (it uses the per-wheel URL, not the registry source), so `uv sync` likely still works — but any future `uv lock` re-resolution without the env var will silently flip these entries back to pypi.org and re-introduce the sdist row for `tensorrt-cu13-libs`. The lock is no longer reproducible across team members.

2. **Lockfile metadata heterogeneity.** Every re-sourced entry **dropped** its `size = ...` and `upload-time = ...` fields (NVIDIA's flat index doesn't surface them), while pypi.org entries kept them. SBOM/audit tooling that reads `upload-time` to gate freshly-uploaded packages will silently skip the NVIDIA wheels. `uv lock --check` accepts this, but the heterogeneity is a visible-from-orbit signal that the lock was generated under a non-shared environment.

3. **Scope creep.** This is a torchcodec-bump PR; the registry switch deserves either a separate PR (so it can be reviewed/rolled-back independently) or a `[[tool.uv.index]]` declaration that makes the NVIDIA index a first-class project-level configuration.

Could you regenerate the lockfile with `UV_EXTRA_INDEX_URL` **unset** (just `uv lock`), so the diff is reduced to the torchcodec change only? If pinning to `pypi.nvidia.com` is desirable as a separate effort, splitting it out into its own PR with a `[[tool.uv.index]]` declaration would make both reviews cleaner.

---

src/opentau/datasets/video_utils.py
```diff
@@ -110,27 +110,39 @@
...
+    level so the load attempt only happens once per process.
     """
-
-    if version.parse(torch.__version__) >= version.parse("2.8.0"):
-        return "pyav"
-    else:
-        if importlib.util.find_spec("torchcodec"):
-            return "torchcodec"
-        else:
+    global _safe_default_codec
```
shuheng-liu: **Minor: cache initialization is not thread-safe.**

The read-check at line 127 and the write at line 138/144 have a TOCTOU window. Two threads (e.g. the gRPC inference server thread pool in `src/opentau/scripts/`, or a notebook with a background prefetcher) entering simultaneously will both execute the `from torchcodec.decoders import VideoDecoder` import and both log the warning on broken hosts.

The result is idempotent (both threads compute the same answer), so this is correctness-benign — but the new docstring's "load attempt only happens once per process" promise is not actually delivered without a lock. Easiest fix: wrap as `@functools.cache` (same effect, plus you get `.cache_clear()` for free — addressing the test-reset concern in my other comment) and inline the body. Same observation applies to `get_safe_encoding_vcodec` below, which uses the same pattern.

---

shuheng-liu (commented):
Multi-angle review of the torchcodec bump. The core idea (bump + try-load fallback) is right; the concerns below are about edge cases the new fallback doesn't cover, and about scope creep from the re-lock.

Highest-priority items:
- **Explicit `backend="torchcodec"` callers still crash** — the new safety net only covers the auto-default path. `decode_video_frames_torchcodec` at line 332 still has an unguarded import, so saved configs that hard-code the backend bypass the protection. (comment on `video_utils.py:132`)
- **Silent backend switch on torch 2.10** — the old `torch>=2.8` gate hard-returned `pyav`; the new code returns `torchcodec` whenever it loads. The two decoders pick frames via different logic (pts-seek vs. `round(ts*fps)` indices), which can change which physical frame is loaded for the same `timestamps` — a potential bit-identical-resume issue per CLAUDE.md §3. (comment on `video_utils.py:144`)
- **Lock-file scope creep** — re-locking with `UV_EXTRA_INDEX_URL=https://pypi.nvidia.com/` set re-sourced ~10 NVIDIA packages and flipped `tensorrt-cu13-libs` from sdist-only to wheels, with the new registry not declared in `pyproject.toml`. (comment on `uv.lock:1953`)

Medium-priority correctness items:
- `find_spec("torchcodec")` is outside the try block and can itself raise `ValueError` (`video_utils.py:129`)
- Warning fires from a dataclass `default_factory` before `init_logging` configures handlers (`video_utils.py:133`)
- Missing torch upper bound; torchcodec 0.10 win wheel exists but marker still excludes win32 (`pyproject.toml:69`)

Footguns worth at least documenting:
- Module-level cache has no public reset hook + long-lived processes can't recover after install fix (`video_utils.py:115`)
- Cache initialization not thread-safe (`video_utils.py:126`)

---

shuheng-liu:
Thanks for the multi-angle review. Addressed in commit f05e60e:

**Fixed:**

1. **Asymmetric error handling (`video_utils.py:132`)** — extracted a shared `_load_torchcodec_videodecoder()` helper. The dispatcher in `decode_video_frames` now consults the probe before routing, so explicit `backend="torchcodec"` callers fall back to `pyav` (with a warning naming the original choice) instead of crashing. `decode_video_frames_torchcodec` keeps raising for direct callers — they get a clearer message pointing at the probe warning.

2. **`find_spec` can raise (`video_utils.py:129`)** — `find_spec` is now inside the try block alongside the import. Added `ValueError` to the except list (the specific failure mode the CPython docs call out).

3. **Lock-file scope creep (`uv.lock:1953`)** — reset `uv.lock` to main, re-locked from an environment that resolves `tensorrt` natively (no `UV_EXTRA_INDEX_URL` needed). New diff is 29 lines, strictly the `torchcodec 0.4.0 → 0.10.0` bump. NVIDIA-registry flips and `tensorrt-cu13-libs` sdist-to-wheels change are gone.

7. **Cache thread-safety + reset (`video_utils.py:115` / `:126`)** — replaced the manual module-level cache with `@functools.cache`, which is thread-safe in CPython and exposes `.cache_clear()` for tests / long-lived processes. The `sys.modules` pollution caveat you raised still applies, but at least there's now a real reset hook.

**Pushing back:**

2. **"Silent backend switch … bit-identical-resume regression" (`video_utils.py:144`)** — I think this conflates two different properties. The PR's whole purpose is to switch the default decoder; calling that out as a regression frames a deliberate migration as a bug. CLAUDE.md §3 asks for bit-identical loss series across **two seeded runs of the same code**, which the new default still satisfies (torchcodec→torchcodec is deterministic). What you're describing is **pre-PR pyav frames vs post-PR torchcodec frames**, which is a decoder-equivalence question, not a determinism property — and it's not bit-identical for either backend pair, by design (different FFmpeg layers). I've added a note to the PR body making the intentional default-flip explicit. Happy to dig deeper if you have a concrete model+dataset where the pixel difference moves a metric, but I don't think a generic determinism check is the right tool for this.

**Deferring (out of scope for the warning fix):**

5. **Warning fires before `init_logging` (`video_utils.py:133`)** — fair, but the same pattern exists in `get_safe_encoding_vcodec` below (and several other `logging.warning` callers in this file). Fixing this properly means a uniform pass on logger handles + lazy emission for both functions, which is its own PR.

6a. **Missing torch upper bound (`pyproject.toml:69`)** — agreed in principle; `torch` and `torchcodec` are ABI-coupled and the open-ended `torch>=2.7.1` lets `uv lock --upgrade torch` silently break us. Worth its own dep-hygiene PR rather than bundling it here — it changes the resolver's behavior across the whole project, not just the video path.

6b. **Windows wheel now exists** — true, but I haven't verified `torchcodec 0.10` actually works on a Windows host (no test surface in CI). Flagged for a Windows-enablement follow-up.

---

src/opentau/datasets/video_utils.py
```diff
@@ -194,11 +215,16 @@ def decode_video_frames(
     if backend is None:
         backend = get_safe_default_codec()
     if backend == "torchcodec":
-        return decode_video_frames_torchcodec(video_path, timestamps, tolerance_s)
-    elif backend in ["pyav", "video_reader"]:
+        if _load_torchcodec_videodecoder() is None:
+            logging.warning(
+                "Caller requested backend='torchcodec' but it failed to load; using 'pyav' for this decode."
```
shuheng-liu: **Warning fires on every `decode_video_frames` call when an explicit `backend="torchcodec"` is set on a host where torchcodec fails to load — log flood in the hot training path.**

`decode_video_frames` is invoked in `LeRobotDataset._query_videos` at `lerobot_dataset.py:1965, 1968` (twice per video key for `image_resample_strategy="linear"`) and `1975` (once for `"nearest"`). A typical pi0.5 run with batch_size=8, 3 video keys, linear resample produces ~48 calls per step — so ~96k warnings over 2000 steps when the saved config has `video_backend="torchcodec"` but the host's torchcodec isn't loadable (cross-host checkpoint restore, container without compatible system FFmpeg, etc.).

The probe warning from `_load_torchcodec_videodecoder` fires once because of `@functools.cache`, but this fallback warning has no once-only guard. Suggest gating it with a tiny once-only sentinel (or another `@functools.cache`'d helper) so the explicit-backend mismatch is surfaced exactly once per process:

```python
@functools.cache
def _warn_explicit_torchcodec_unloadable() -> None:
    logging.warning(
        "Caller requested backend='torchcodec' but it failed to load; using 'pyav' instead."
    )

# inside decode_video_frames:
if backend == "torchcodec":
    if _load_torchcodec_videodecoder() is None:
        _warn_explicit_torchcodec_unloadable()
        backend = "pyav"
    else:
        return decode_video_frames_torchcodec(video_path, timestamps, tolerance_s)
```

---

src/opentau/datasets/video_utils.py
```diff
@@ -194,11 +215,16 @@ def decode_video_frames(
     if backend is None:
         backend = get_safe_default_codec()
     if backend == "torchcodec":
```
shuheng-liu, line 239: **Silent downgrade of an explicit `backend="torchcodec"` to `"pyav"` is a behavior change that's undocumented and risks masking determinism / benchmark bugs.**

Pre-PR, passing `backend="torchcodec"` on a broken host crashed at first decode. Post-PR, the function quietly switches to `"pyav"` for the lifetime of the process. Two concerns:

1. **Frame-selection differs across the two backends** (CLAUDE.md §3): `decode_video_frames_torchvision` (pyav path) sets `keyframes_only=True` and matches `argmin` over `pts`; `decode_video_frames_torchcodec` uses `seek_mode="exact"` and `[round(ts*average_fps)]` indexing. For variable-fps videos or imprecise `average_fps` metadata, the same `(episode_idx, timestamps)` pair can resolve to different physical frames. A checkpoint trained with torchcodec on rank-0 and resumed on a rank where torchcodec silently downgrades will see different input pixels — bit-identical resume is no longer guaranteed.

2. **Benchmark masking**: anyone benchmarking `pyav` vs `torchcodec` on a misconfigured host gets two `pyav` runs labelled differently and won't notice.

Suggestions:
- Document the fallback explicitly in `decode_video_frames`'s docstring (currently lines 201-213 still say "Currently supports torchcodec on cpu and pyav" with no mention of automatic fallback).
- Consider making the fallback opt-in via a parameter — e.g. `backend_fallback: bool = False` — so explicit backend selection stays strict by default. The auto-default path (`backend is None`) would always allow fallback.
- At minimum, add a one-line note in the PR description that explicit `backend="torchcodec"` callers now get silent fallback instead of a crash.

---

src/opentau/datasets/video_utils.py
```diff
@@ -110,27 +111,47 @@
...
+    except (ImportError, RuntimeError, ValueError) as e:
+        logging.warning(
+            "'torchcodec' failed to load (%s); falling back to 'pyav' as the default decoder.",
+            e,
+        )
+        return None
+    return VideoDecoder
+
+
+@functools.cache
```
shuheng-liu, line 148: **`@functools.cache` on `get_safe_default_codec` is redundant and a `default_factory` footgun.**

Two small concerns about caching this specific function:

1. **Redundant.** `_load_torchcodec_videodecoder()` is already `@functools.cache`'d. `get_safe_default_codec` just translates its return to a string — caching the no-arg wrapper of an already-cached function is pure boilerplate.

2. **`default_factory` contract.** `configs/default.py:115` uses `field(default_factory=get_safe_default_codec)`. By dataclass convention `default_factory` is expected to return a **fresh** instance per call (that's the whole reason `field(default_factory=list)` exists instead of `default=[]`). The return type is `str` (immutable), so this is safe today — but if anyone ever changes `get_safe_default_codec` to return a richer object (e.g. a `BackendSpec` dict, a fallback chain list), all configs would silently share the same instance and mutations would leak across instances.

Both are addressable in one stroke: drop `@functools.cache` from `get_safe_default_codec`. The probe still runs only once because `_load_torchcodec_videodecoder` is cached; the str dispatch is trivial.

---

src/opentau/datasets/video_utils.py
```diff
@@ -317,10 +343,11 @@ def decode_video_frames_torchcodec(
...
 
-    if importlib.util.find_spec("torchcodec"):
-        from torchcodec.decoders import VideoDecoder
-    else:
-        raise ImportError("torchcodec is required but not available.")
+    VideoDecoder = _load_torchcodec_videodecoder()  # noqa: N806
+    if VideoDecoder is None:
+        raise ImportError(
+            "torchcodec is required but not loadable on this host (see prior warning for details)."
+        )
```
shuheng-liu: **"see prior warning for details" is fragile — the original warning may have been swallowed at config-parse time or rotated out of long-running logs.**

`_load_torchcodec_videodecoder` logs the underlying exception once on first probe (typically during config-parse, before `init_logging` runs) and discards it. Anyone hitting this `ImportError` later has to grep historic logs for the root cause — which may have been silently dropped (root logger unconfigured at probe time) or rolled out of a rotating file handler.

Cheap fix: cache the underlying exception in the probe, then include it in the `ImportError` message so the failure mode is self-contained:

```python
@functools.cache
def _load_torchcodec_videodecoder() -> tuple[type | None, BaseException | None]:
    try:
        if importlib.util.find_spec("torchcodec") is None:
            return None, ImportError("torchcodec package not installed")
        from torchcodec.decoders import VideoDecoder
    except (ImportError, RuntimeError, ValueError) as e:
        logging.warning("'torchcodec' failed to load (%s); falling back to 'pyav' as the default decoder.", e)
        return None, e
    return VideoDecoder, None

# in decode_video_frames_torchcodec:
VideoDecoder, probe_exc = _load_torchcodec_videodecoder()
if VideoDecoder is None:
    raise ImportError(
        f"torchcodec is required but not loadable on this host: {probe_exc!r}"
    ) from probe_exc
```

Or simpler: just inline the exception's `repr()` into the error message at probe time, and have the cached value carry a string explanation.

---

shuheng-liu (commented):
Re-review of commit `f05e60e` (refactor: extract `_load_torchcodec_videodecoder` helper + `functools.cache`).

**Resolved by this commit** (from the previous review):
- Asymmetric error handling — both `get_safe_default_codec` and `decode_video_frames_torchcodec` now share the probe via `_load_torchcodec_videodecoder()`.
- `find_spec` could raise — now inside the `try` block, with `ValueError` added to the except.
- Thread-safety + test-reset hook — `@functools.cache` gives both for free.
- Lock-file scope creep — `uv.lock` was re-regenerated without `UV_EXTRA_INDEX_URL` and the diff is back to 9 lines focused on torchcodec.

**New concerns introduced by the refactor:**
- Per-call warning at `video_utils.py:220` floods logs in the `__getitem__` hot path when an explicit `backend="torchcodec"` doesn't load.
- Silent downgrade of explicit `backend="torchcodec"` to `"pyav"` is undocumented and risks masking frame-selection determinism issues (CLAUDE.md §3) and benchmarks.
- `@functools.cache` on `get_safe_default_codec` is redundant given the inner probe is already cached, and is a footgun for `default_factory` if the return type ever changes.
- `ImportError: ... see prior warning for details` is fragile when the original warning was dropped (logging unconfigured at probe time) or rotated out.

**Still open from the previous review** (already commented; not duplicating):
- Warning at probe time fires before `init_logging` (still applies because `get_safe_default_codec` is still wired as `default_factory`).
- `pyproject.toml:69` — missing torch upper bound and Windows wheel marker mismatch.

---

shuheng-liu:
Round 2 addressed in commit 4011cbb. All four new concerns are direct fixes; one is a deliberate partial.

**Fixed:**

- **#9 — Per-call warning flood at `video_utils.py:220`** — moved the fallback warning behind `_warn_explicit_torchcodec_unloadable()`, a `@functools.cache`'d sentinel that fires exactly once per process, matching your suggested pattern. The ~96k-warnings-per-2000-steps regression is gone.

- **#11 — Redundant `@functools.cache` on `get_safe_default_codec`** — dropped. The inner probe is still cached so the load attempt remains once-per-process; the wrapper now returns a fresh `str` per call, honoring the `default_factory` fresh-instance contract you flagged.

- **#12 — Fragile "see prior warning"** — probe now returns `tuple[type | None, BaseException | None]`. `decode_video_frames_torchcodec` raises with the cached exception's `repr` and chains it via `raise ImportError(...) from probe_exc`, so the failure mode is self-contained regardless of whether `init_logging` had run when the probe fired.

**Partial fix:**

- **#10 — Silent downgrade undocumented + frame-selection determinism** — extended `decode_video_frames`'s docstring to explicitly name the fallback, document that the two backends pick frames via non-equivalent logic (`pts seek + argmin` vs. `round(ts * average_fps)`-indexing), and point strict callers at the probe.

  Holding the line on the `backend_fallback: bool` parameter for now. The concrete failure scenario you sketched ("rank-0 with torchcodec, rank-N silently downgrades") requires a heterogeneous distributed cluster, which the project doesn't really support — homogeneous container images are the operating assumption. The cross-host-checkpoint-resume scenario is real but already a soft mismatch (different num_workers, different resolution, etc. all "just work" with slightly different training), not a hard correctness bug. Adding a parameter just to gate this single behavior expands the API surface for a use case that's better solved by the strict caller checking `_load_torchcodec_videodecoder()[0]` directly — which the new docstring now points at. Happy to add the flag in a follow-up if a concrete use case appears.

**Still deferred (carried from round 1):**

- Probe warning fires before `init_logging`. Same fix wants to land on `get_safe_encoding_vcodec` too — own PR.
- `pyproject.toml:69` torch upper bound + Windows wheel marker. Own dep-hygiene PR.

---

src/opentau/datasets/video_utils.py
```diff
@@ -110,27 +111,61 @@
 import torch
 import torchvision
 from datasets.features.features import register_feature
-from packaging import version
 from PIL import Image
 
 
-def get_safe_default_codec() -> str:
-    """Get the default video codec backend, falling back to pyav if torchcodec is unavailable.
+@functools.cache
```
shuheng-liu, line 117: **No unit-test coverage for the new probe / fallback paths.**

`grep -rn 'get_safe_default_codec\|_load_torchcodec\|_warn_explicit' tests/` returns nothing. The PR has gone through three rounds of fixes and now has substantial new behavior — tuple-return probe, exception forwarding, once-only fallback warning, silent backend downgrade with chained `ImportError` — none of which is exercised by a test. The existing `test_deferred_video_attach_video_start_time` only covers the happy path with the auto-selected backend.

A small test file (e.g. `tests/datasets/test_video_utils.py`) could pin each branch using `monkeypatch` + `cache_clear`:

```python
import importlib.util
import pytest
from opentau.datasets import video_utils

@pytest.fixture(autouse=True)
def _clear_probe_cache():
    video_utils._load_torchcodec_videodecoder.cache_clear()
    video_utils._warn_explicit_torchcodec_unloadable.cache_clear()
    yield
    video_utils._load_torchcodec_videodecoder.cache_clear()
    video_utils._warn_explicit_torchcodec_unloadable.cache_clear()

def test_get_safe_default_codec_missing(monkeypatch):
    monkeypatch.setattr(importlib.util, "find_spec", lambda name: None if name == "torchcodec" else importlib.util.find_spec(name))
    assert video_utils.get_safe_default_codec() == "pyav"
    cls, exc = video_utils._load_torchcodec_videodecoder()
    assert cls is None
    assert isinstance(exc, ModuleNotFoundError)

def test_decode_video_frames_torchcodec_raises_with_chained_exception(monkeypatch):
    monkeypatch.setattr(importlib.util, "find_spec", lambda name: None if name == "torchcodec" else importlib.util.find_spec(name))
    with pytest.raises(ImportError, match="ModuleNotFoundError") as ei:
        video_utils.decode_video_frames_torchcodec("dummy.mp4", [0.0], 0.1)
    assert isinstance(ei.value.__cause__, ModuleNotFoundError)

def test_explicit_torchcodec_falls_back_to_pyav(monkeypatch):
    monkeypatch.setattr(importlib.util, "find_spec", lambda name: None if name == "torchcodec" else importlib.util.find_spec(name))
    # ... assert decode_video_frames(..., backend="torchcodec") routes to torchvision and emits exactly one warning
```

That set covers the three regressions a future contributor could re-introduce: missing-package fallback, exception chaining, and once-only warning. Non-blocking but recommended given how many subtle bugs surfaced in the last three rounds.

---

src/opentau/datasets/video_utils.py
```diff
@@ -184,21 +219,32 @@ def decode_video_frames(
...
-    Currently supports torchcodec on cpu and pyav.
+    Currently supports torchcodec on cpu and pyav. If the caller explicitly
+    requests ``backend="torchcodec"`` but the host's torchcodec extension
+    cannot load (e.g. missing system FFmpeg, ABI mismatch), this function
+    silently downgrades to ``"pyav"`` with a once-per-process warning. The two
+    decoders pick frames using non-equivalent logic (pts seek + argmin vs.
+    ``round(ts * average_fps)`` indexing), so the decoded pixels for the same
+    ``(video_path, timestamps)`` pair can differ within ``tolerance_s``;
+    callers needing strict backend selection should check the return of
+    ``_load_torchcodec_videodecoder()`` before dispatching.
```
shuheng-liu: **Minor docs nit: this user-facing docstring references a leading-underscore private helper.**

> *callers needing strict backend selection should check the return of `_load_torchcodec_videodecoder()` before dispatching.*

Pointing callers at a private function (leading underscore = private by convention) is a bit awkward. Two cleaner options:

1. Recommend `get_safe_default_codec() == "torchcodec"` instead — that's already the user-facing API, and on torch 2.10 it equivalently reflects whether the probe succeeded.
2. Expose a public `is_torchcodec_loadable() -> bool` (thin wrapper around the probe) and recommend that.

Trivial — flagging only because the rest of the docstring is very thorough.

---

shuheng-liu (commented):
Round-3 review of commit `4011cbb` (refactor: address round-2 review on probe + fallback).

**All four round-2 findings cleanly addressed:**
- **Warning spam** — `_warn_explicit_torchcodec_unloadable()` with `@functools.cache` keeps the explicit-backend fallback warning to one log line per process.
- **Silent downgrade undocumented** — `decode_video_frames` docstring now explicitly calls out the fallback and the pyav-vs-torchcodec frame-selection difference.
- **`@functools.cache` on `get_safe_default_codec`** — removed, with a docstring explaining the `default_factory` reasoning. Nice.
- **"see prior warning" was fragile** — probe now returns `(VideoDecoder, exc)`; the `ImportError` interpolates `repr(probe_exc)` and chains via `raise ... from probe_exc`. Tracebacks are self-contained now.

**Remaining items** (two inline comments below):
- No unit-test coverage for any of the new probe / fallback paths (`tests/` has no reference to `get_safe_default_codec`, `_load_torchcodec_videodecoder`, or `_warn_explicit_torchcodec_unloadable`).
- The `decode_video_frames` docstring routes callers needing strict backend selection at a leading-underscore private helper — minor wording nit.

**Still open from earlier rounds** (already commented, not duplicating):
- Probe warning fires from `default_factory` before `init_logging` runs (`video_utils.py:131` live comment).
- `pyproject.toml:69` — missing torch upper bound and Windows wheel marker mismatch (live comment).

Approach to those two is mostly a judgment call about scope: the probe-timing one is fixable in this PR with a tiny `__post_init__` deferral; the pyproject ones could be a separate PR. Either way, this commit's correctness is in great shape.

---

shuheng-liu:
Round 3 addressed in commit 68df68a.

**Fixed:**

- **Test coverage gap** — new `tests/datasets/test_video_utils.py` covers the four behaviors the last three rounds introduced:
  - probe returns `(None, ModuleNotFoundError)` when torchcodec is missing
  - `get_safe_default_codec()` reflects that
  - `decode_video_frames_torchcodec` raises `ImportError` chained from the cached probe exception
  - `decode_video_frames(..., backend="torchcodec")` silently downgrades to pyav AND emits exactly one fallback warning across three calls

  An autouse fixture clears both `@functools.cache`'d entries around every test so state set by one case doesn't leak into the next. All four pass locally; CI should have them on the next run.

- **Docstring nit** — `decode_video_frames`'s strict-selection note now points at the public `get_safe_default_codec() == "torchcodec"` instead of the private `_load_torchcodec_videodecoder()` probe. Took option 1 (no need to introduce a new public wrapper).

**Closing out the remaining open items:**

- **Probe warning fires before `init_logging`** — still deferred. It's load-bearing only when torchcodec actually fails to load (which on the production container it doesn't, since FFmpeg 4 + libavcodec.so.58 are present). The proper fix lands together with `get_safe_encoding_vcodec`'s identical pattern in its own follow-up PR.
- **Windows wheel marker** — closing as won't-fix. Project doesn't target Windows. Marker stays restrictive.
- **Missing torch upper bound** — still deferred. Project-wide resolver change deserves its own dep-hygiene PR.

I think this PR is ready for human review now.

---

PR #333, opened 2026-05-26
feat(envs): inject subgoal images into LIBERO eval

shuheng-liu:
## What this does

Closes the train↔eval input asymmetry for `pi07` low-level / `pi07-paligemma` low-level on LIBERO: at training time these policies see a `subgoal{k}` image per camera (decoded from a future frame of the training episode); at eval time those keys are absent today and `prepare_subgoal_images` warns + zero-fills with `mask=False`. This PR plumbs a real subgoal image into LIBERO eval rollouts via a `SubgoalImageGenerator` abstraction, with a first, naive concrete implementation:

- For each env in a vector rollout, look up the env's task language (one of the 40 LIBERO task strings) in the `TensorAuto/libero` dataset, sample a random matching episode, and serve that episode's *last* frame from each camera as the subgoal.
- "Once per `env.reset()`" cadence: `start_episode(prompts)` is called at the top of each `rollout()` and the chosen episodes (+ decoded frames) are cached for the whole rollout. Fresh draw on the next rollout.
- Raises `ValueError` if a prompt has no matching episode in the source dataset, so a misconfigured eval surfaces immediately instead of silently no-oping.

Opt-in via `--env.subgoal_source=TensorAuto/libero` on `LiberoEnv`. Default `None` keeps existing eval behavior unchanged (model falls back to its zero-subgoal placeholder path). Wired into both `opentau-eval` and the mid-training eval block in `scripts/train.py`.

Implementation notes:
- `LiberoLastFrameSubgoalGenerator` handles both video-dtype and image-dtype cameras. `TensorAuto/libero` stores both `image` and `wrist_image` as image-dtype (per-frame in parquet), so the image path is the hot one — a single `load_dataset(...)` per episode serves all cameras of that episode.
- `LeRobotDatasetMetadata` is downloaded once (`meta/` only); episode parquet / video files are pulled lazily via `hf_hub_download` on first frame access.
- Per-rollout `chosen_episodes` lives on `threading.local()` so `max_parallel_tasks > 1` in `eval_policy_all` stays safe; the metadata + frame cache are shared across rollouts.

## How it was tested

- **Unit tests** (`tests/envs/test_subgoal.py`, 20 tests, all CPU, `not gpu`):
  - Language→episode index construction, sorted-known-languages, camera-key resolution against `DATA_FEATURES_NAME_MAPPING`, `num_cams` cap, unknown-repo `KeyError`.
  - `start_episode`: one episode per env, missing-prompt `ValueError`, deterministic under seeded `random`.
  - `__call__`: keys are `subgoal{k}` + `subgoal_is_pad`, shape `(B, 3, *resolution)` in `[0, 1]`, dtype/device matches `observation["state"]`, episode-level cache so `__call__×2` triggers exactly one decode per episode, `RuntimeError` if `start_episode` was skipped, `ValueError` on batch-size mismatch.
  - Image-dtype path is mocked separately (`set_transform` + single `__getitem__` per episode).
  - `add_subgoal_images` no-op when generator is `None`; merge-in-place when not.
- **Real-dataset integration test** (`@pytest.mark.slow`): instantiates `LiberoLastFrameSubgoalGenerator(repo_id="TensorAuto/libero")` against the live HuggingFace repo. Verifies ≥40 known task languages and end-to-end `start_episode` + `__call__` against a real episode. Self-skips if the metadata isn't reachable.
- **Cold-cache verification** on an RTX 3090 dev box (no `TensorAuto/libero` cache prior): all 20 tests pass in <4s, the integration test triggered `hf_hub_download` for `data/chunk-000/episode_000410.parquet` and decoded the last frame from both cameras. File landed on disk afterward.
- **Full CPU regression**: `pytest -m "not gpu and not slow"` — 1269 passed, 14 skipped. The 4 errors (`tests/envs/test_libero_control_freq.py`, `tests/utils/test_libero_utils.py`, `tests/envs/test_factory.py::TestMakeEnv`) are pre-existing LIBERO-config-needs-stdin issues that reproduce identically on `main`.
- **Pre-commit**: `pre-commit run --all-files` passes.

```bash
git checkout claude/amazing-tereshkova-abec3a
pytest -sx tests/envs/test_subgoal.py
```

End-to-end on a GPU host (opt-in flag is what flips the behavior):

```bash
opentau-eval --config_path=<libero-eval-config> --env.subgoal_source=TensorAuto/libero
```

Diff against `main` with `subgoal_source=None` (default) should be byte-identical; the per-step `prepare_subgoal_images` warning ("all subgoal image features are missing") disappears when the flag is set.

## Checklist

### DIFF
--- a/src/opentau/envs/subgoal.py
+++ b/src/opentau/envs/subgoal.py
@@ -1,0 +1,321 @@
+# Copyright 2026 Tensor Auto Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Eval-time subgoal image generators.
+
+The pi07 low-level and pi07-paligemma low-level policies are trained with
+subgoal images keyed as ``subgoal{k}`` (one per camera in
+``config.image_features``) plus a single ``subgoal_is_pad`` bool. At eval
+time these keys are produced by a :class:`SubgoalImageGenerator` that
+inspects the observation dict (post ``preprocess_observation`` +
+``add_envs_task`` + ``add_eval_metadata``) and returns the matching batch
+keys. :func:`opentau.envs.utils.add_subgoal_images` is the wiring helper
+that calls the generator and merges its output into the observation.
+
+The only concrete generator implemented today is
+:class:`LiberoLastFrameSubgoalGenerator`, which serves the
+``TensorAuto/libero`` 20fps v2.1 relabel: it looks up the env's task
+language (one of the 40 LIBERO task strings) in the dataset, samples a
+random matching episode at the start of each rollout, and returns that
+episode's last frame from each camera as the subgoal.
+"""
+
+from __future__ import annotations
+
+import random
+import threading
+from pathlib import Path
+from typing import Any, Protocol
+
+import torch
+import torch.nn.functional as F  # noqa: N812
+from huggingface_hub import hf_hub_download
+from torch import Tensor
+
+from opentau.datasets.lerobot_dataset import LeRobotDatasetMetadata
+from opentau.datasets.standard_data_format_mapping import DATA_FEATURES_NAME_MAPPING
+from opentau.datasets.video_utils import decode_video_frames
+
+
+class SubgoalImageGenerator(Protocol):
+    """Eval-time contract for producing subgoal images.
+
+    Generators are constructed once at the top of
+    :func:`opentau.scripts.eval.eval` and plumbed through to
+    :func:`opentau.scripts.eval.rollout`, which calls
+    :meth:`start_episode` once at the start of each rollout (matching the
+    "once per episode reset" cadence) and then :meth:`__call__` on every
+    step. Implementations are expected to memoise the per-rollout choice
+    inside :meth:`start_episode` so the per-step call is cheap.
+    """
+
+    def start_episode(self, prompts: list[str]) -> None:
+        """Re-sample the per-env subgoal source for a fresh rollout.
+
+        Called once at the top of :func:`opentau.scripts.eval.rollout`
+        after ``env.reset()`` and ``add_envs_task`` have populated
+        ``observation["prompt"]``.
+
+        Args:
+            prompts: Task language string per env in the vector, length
+                ``env.num_envs``.
+        """
+        ...
+
+    def __call__(self, observation: dict[str, Any]) -> dict[str, Tensor]:
+        """Return the subgoal batch keys for the current observation.
+
+        Args:
+            observation: Observation dict after ``preprocess_observation``,
+                ``add_envs_task``, and ``add_eval_metadata`` — contains
+                ``camera{k}``, ``state``, ``img_is_pad``, ``prompt``, and
+                any pi07 metadata keys.
+
+        Returns:
+            Dict with ``subgoal{k}`` ``(B, 3, H, W)`` floats in ``[0, 1]``
+            for each camera the generator can serve, plus
+            ``subgoal_is_pad`` ``(B,)`` bool. Cameras the generator does
+            not have data for are left out; the policy's
+            ``prepare_subgoal_images`` fills missing slots with ``-1``
+            placeholders and a ``mask=False``.
+        """
+        ...
+
+
+class LiberoLastFrameSubgoalGenerator:
+    """Naive LIBERO subgoal generator backed by ``TensorAuto/libero``.
+
+    For each env in the rollout, looks up the task language string (one
+    of the 40 LIBERO subtask strings) in the dataset's episode index,
+    samples a random matching episode at the start of the rollout, and
+    returns that episode's *last* frame from each video camera as the
+    subgoal. Camera images are resized with aspect-ratio padding to match
+    ``resolution`` (the same shape ``preprocess_observation`` produces for
+    the live ``camera{k}`` keys).
+
+    The chosen episodes are pinned by :meth:`start_episode` and reused
+    across every :meth:`__call__` in the rollout — fresh draws happen on
+    the next ``rollout()`` call, which corresponds to the next
+    ``env.reset()``. If a prompt has no matching episode in the dataset,
+    :meth:`start_episode` raises ``ValueError``.
+
+    Today only ``repo_id="TensorAuto/libero"`` is exercised; other
+    LeRobot v2.1 repos with the same camera-name convention should work
+    but are not tested.
+    """
+
+    def __init__(
+        self,
+        repo_id: str = "TensorAuto/libero",
+        root: str | Path | None = None,
+        resolution: tuple[int, int] = (256, 256),
+        num_cams: int = 2,
+        tolerance_s: float = 1e-4,
+    ):
+        """Load dataset metadata and build the task-language episode index.
+
+        Args:
+            repo_id: HuggingFace dataset repo to source subgoal frames
+                from. Must be a v2.1 LeRobot dataset registered in
+                :data:`DATA_FEATURES_NAME_MAPPING`.
+            root: Optional local cache root for the dataset. Defaults to
+                the OpenTau-managed location used by ``LeRobotDataset``.
+            resolution: Target ``(H, W)`` for the returned subgoal
+                images, matching ``cfg.resolution``.
+            num_cams: Number of camera slots to populate. Capped by the
+                cameras the source dataset actually exposes (typically 2
+                for LIBERO).
+            tolerance_s: Timestamp tolerance for video frame decoding.
+        """
+        self.repo_id = repo_id
+        self.resolution = resolution
+        self.tolerance_s = tolerance_s
+
+        # `LeRobotDatasetMetadata.__init__` downloads only the `meta/`
+        # subdir on cache miss — the video files themselves are pulled
+        # lazily by `_decode_last_frame` on first use.
+        self.meta = LeRobotDatasetMetadata(repo_id=repo_id, root=root)
+
+        if repo_id not in DATA_FEATURES_NAME_MAPPING:
+            raise KeyError(
+                f"LiberoLastFrameSubgoalGenerator: repo_id {repo_id!r} has no entry in "
+                f"DATA_FEATURES_NAME_MAPPING; cannot resolve camera{{k}} -> raw key."
+            )
+        name_map = DATA_FEATURES_NAME_MAPPING[repo_id]
+
+        # Resolve camera slots to the raw feature keys the dataset
+        # exposes, cap by `num_cams`, drop any slot whose raw key is not
+        # present in the dataset's video_keys / image_keys. Slots beyond
+        # this list won't get a `subgoal{k}` and the model fills them
+        # with -1 placeholders.
+        camera_keys: dict[int, str] = {}
+        for k in range(num_cams):
+            raw_key = name_map.get(f"camera{k}")
+            if raw_key is None:
+                continue
+            if raw_key in self.meta.video_keys or raw_key in self.meta.image_keys:
+                camera_keys[k] = raw_key
+        if not camera_keys:
+            raise ValueError(
+                f"No subgoal cameras resolved for {repo_id!r}: name_map="
+                f"{name_map}, video_keys={self.meta.video_keys}, image_keys={self.meta.image_keys}."
+            )
+        self.camera_keys = camera_keys
+
+        self.lang_to_episodes: dict[str, list[int]] = {}
+        for ep_idx, info in self.meta.episodes.items():
+            for task_str in info.get("tasks", []):
+                self.lang_to_episodes.setdefault(task_str, []).append(ep_idx)
+
+        # The frame cache is shared across rollouts (read-mostly; the
+        # CPython GIL makes the `setdefault` + assignment safe enough — a
+        # racing-thread cache miss at worst duplicates a video decode).
+        # `chosen_episodes` lives on threadlocal storage so concurrent
+        # `eval_policy_all` workers (when `max_parallel_tasks > 1`) don't
+        # clobber each other's per-rollout pick.
+        self._frame_cache: dict[int, dict[str, Tensor]] = {}
+        self._local = threading.local()
+
+    @property
+    def fps(self) -> int:
+        return self.meta.fps
+
+    def known_languages(self) -> list[str]:
+        return sorted(self.lang_to_episodes)
+
+    def start_episode(self, prompts: list[str]) -> None:
+        """Sample one random matching episode per env, store for the rollout.
+
+        Raises:
+            ValueError: If any prompt has no matching episode in
+                ``lang_to_episodes``. The error names the prompt and the
+                count of known languages.
+        """
+        chosen: list[int] = []
+        for prompt in prompts:
+            candidates = self.lang_to_episodes.get(prompt)
+            if not candidates:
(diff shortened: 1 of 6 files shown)


src/opentau/envs/subgoal.py
```diff
@@ -0,0 +1,357 @@
...
+    The chosen episodes are pinned by :meth:`start_episode` and reused
+    across every :meth:`__call__` in the rollout — fresh draws happen on
+    the next ``rollout()`` call, which corresponds to the next
+    ``env.reset()``. If a prompt has no matching episode in the dataset,
+    :meth:`start_episode` raises ``ValueError``.
+
+    Today only ``repo_id="TensorAuto/libero"`` is exercised; other
+    LeRobot v2.1 repos with the same camera-name convention should work
+    but are not tested.
+    """
```
shuheng-liu, line 138: Worth surfacing the **train↔eval distribution shift** this implementation creates so config writers don't trust the flag implicitly.

At training time, `BaseDataset._sample_subgoal_frame` (`lerobot_dataset.py:2013–2041`) reads a *future frame within the same episode* — either the segment end (with probability `subgoal_end_of_segment_prob`) or a uniform sample in `[t, t+4s]` clipped to the segment. The policy was conditioned on **within-episode** future state.

This generator picks a *different* episode (matched only by task language) and always returns its *last* frame. Cross-episode last-frame subgoals are a different conditional distribution; even when the goal state is semantically valid, the model has never been trained to produce actions conditioned on a foreign trajectory's terminus.

The PR docstring already calls this "naive" — fair for a first cut. Two cheap follow-ups that would close most of the gap:

1. Roughly match the training sampling distribution (uniform within `[t, t+4s]` if reachable, else clamp to end; sample from the *same* episode the env is rolling out, if/when that wiring exists; otherwise at least add a `subgoal_end_of_segment_prob`-equivalent knob and sample from `[0, len)` of the source episode rather than fixing on `length-1`).
2. Before relying on this flag for downstream eval decisions, run an A/B with vs. without `--env.subgoal_source=TensorAuto/libero` on the published `pi07_low_level` / `pi07_paligemma_low_level` checkpoints and report the delta in the PR body. Right now the only evidence the flag does the right thing is that the warning disappears — that's a wiring assertion, not a correctness assertion.

---

src/opentau/envs/subgoal.py
```diff
@@ -0,0 +1,357 @@
...
+        chosen: list[int] = []
+        for prompt in prompts:
+            candidates = self.lang_to_episodes.get(prompt)
+            if not candidates:
+                raise ValueError(
+                    f"LiberoLastFrameSubgoalGenerator: no episodes in {self.repo_id!r} match "
+                    f"prompt {prompt!r} (dataset has {len(self.lang_to_episodes)} known task "
+                    f"languages)."
+                )
+            chosen.append(random.choice(candidates))
```
shuheng-liu: Python's `random` is process-global. `set_seed(cfg.seed)` (`utils/random_utils.py:219`) does seed it, so **single-threaded** eval reproduces correctly. But `eval_policy_all` runs concurrent threads when `cfg.env.max_parallel_tasks > 1`, and those threads share the same global `random` state. Individual `random.choice` calls are atomic under the GIL, but their **order across threads** is non-deterministic — so the per-rollout episode picks (and therefore the per-step subgoal images) drift across runs even with a fixed `cfg.seed`. The `chosen_episodes` threadlocal isolates state *after* the pick but not the source of randomness.

`test_deterministic_with_seeded_random` asserts the single-threaded case; the multi-threaded case isn't exercised today.

Cheapest fix: a per-generator `random.Random` instance seeded explicitly at construction.

```python
def __init__(self, ..., seed: int | None = None):
    ...
    self._rng = random.Random(seed)
    ...

def start_episode(self, prompts):
    ...
    chosen.append(self._rng.choice(candidates))
```

Then thread `cfg.seed` through `make_subgoal_generator`. If `max_parallel_tasks > 1` is a real configuration, additionally guard `_rng.choice` with a `threading.Lock` (or give each thread its own `random.Random` via `threading.local`) — concurrent calls on a single `random.Random` are technically supported under GIL but still produce a thread-interleaved sequence; per-thread RNGs are the only way to get bit-identical picks across runs.

---

src/opentau/envs/subgoal.py
```diff
@@ -0,0 +1,357 @@
...
+            for task_str in info.get("tasks", []):
+                self.lang_to_episodes.setdefault(task_str, []).append(ep_idx)
+
+        # The frame cache is shared across rollouts (read-mostly; the
+        # CPython GIL makes the `setdefault` + assignment safe enough — a
+        # racing-thread cache miss at worst duplicates a video decode).
+        # `chosen_episodes` lives on threadlocal storage so concurrent
+        # `eval_policy_all` workers (when `max_parallel_tasks > 1`) don't
+        # clobber each other's per-rollout pick.
+        self._frame_cache: dict[int, dict[str, Tensor]] = {}
```
shuheng-liu, line 227: `_frame_cache` grows unboundedly across rollouts — there's no eviction over the eval lifetime. For full LIBERO (1693 episodes × 2 cameras × ~200 KB per resized frame) the worst case is ~660 MB, which is fine in practice but still a slow leak that survives every `eval_freq`-triggered mid-training eval block (where this generator lives for the full training run).

If/when the generator gets pointed at a larger dataset or repurposed for non-LIBERO repos, this becomes a real footgun. Cheap insurance:

```python
from functools import lru_cache

# Or wrap `_cached_frame` with `lru_cache(maxsize=...)` on a free function.
```

…or just a manual `dict` + size check + FIFO eviction keyed on `(ep_idx, raw_key)`. Not blocking for this PR, but worth a TODO in the docstring at least.

---

src/opentau/envs/subgoal.py
```diff
@@ -0,0 +1,357 @@
...
+            tolerance_s: Timestamp tolerance for video frame decoding.
+        """
+        self.repo_id = repo_id
+        self.resolution = resolution
+        self.tolerance_s = tolerance_s
+
+        # `LeRobotDatasetMetadata.__init__` downloads only the `meta/`
+        # subdir on cache miss — the video files themselves are pulled
+        # lazily by `_decode_last_frame` on first use.
+        self.meta = LeRobotDatasetMetadata(repo_id=repo_id, root=root)
```
shuheng-liu: `LeRobotDatasetMetadata(repo_id=repo_id, root=root)` falls back to `revision=CODEBASE_VERSION` (`lerobot_dataset.py:379`). Today `TensorAuto/libero` is at v2.1 and `CODEBASE_VERSION` is v2.1, so this resolves the way the PR description says. If either side drifts (codebase bumps to v2.2 while `TensorAuto/libero` stays at v2.1, or vice versa) the generator will silently grab a different revision than the training-time dataset and the eval-time subgoal frames could be from a different relabel pass than what trained the policy.

Threading `revision` through `LiberoLastFrameSubgoalGenerator.__init__` and `make_subgoal_generator` (defaulting to `None` to preserve current behavior, but allowing the config to pin) is a one-line guard. Alternatively, log the resolved `self.meta.revision` at `INFO` from `make_subgoal_generator` so a deviation is at least visible in the eval log.

---

src/opentau/scripts/eval.py
```diff
@@ -192,6 +205,7 @@ def rollout(
         observation = preprocess_observation(observation, cfg=cfg)
         observation = add_envs_task(env, observation)
         observation = add_eval_metadata(observation, cfg=cfg)
+        observation = add_subgoal_images(observation, subgoal_generator)
```
shuheng-liu: Minor edge case: this terminal-observation block runs unconditionally when `return_observations=True`, but `subgoal_episode_picked` is gated on the loop body executing at least once. If a degenerate env returns `done=True` immediately on reset (so the `while` loop never enters), `start_episode` was never called and `add_subgoal_images` here would raise `RuntimeError("start_episode must be called first")` from the generator's `__call__`.

This is purely theoretical — `done = [False] * env.num_envs` at line 137 and `max_steps >= 1`, so I don't think this is reachable in any current configuration. But the symmetry between the loop body and this block is broken; the cleanest fix is to also gate this `add_subgoal_images` call on `subgoal_episode_picked`, or move the unconditional `subgoal_generator.start_episode(...)` call to right after the initial `env.reset()` (before the loop) so the cadence assumption — "exactly one `start_episode` per `rollout()` call" — holds even on zero-step rollouts.

---

shuheng-liu (commented):
Solid initial implementation. The `SubgoalImageGenerator` Protocol cleanly separates the eval-time injection contract from the LIBERO-specific concrete generator, the opt-in default (`subgoal_source=None`) preserves existing eval behavior byte-for-byte, and the test coverage is genuinely good — 20 mocked unit tests pinning shape/dtype/value-range/cache/error paths, plus a `@pytest.mark.slow` integration test that self-skips on missing network. I reproduced the unit-test suite locally (18 non-slow tests, 0.15s, all green).

Wiring is consistent end-to-end: `make_subgoal_generator` -> `eval_main`, `train.py`'s mid-training eval block, and the three-level fanout through `eval_policy_all` / `run_one` / `eval_one` / `eval_policy` / `rollout`. The `subgoal_episode_picked` flag in `rollout()` correctly enforces the "one `start_episode` per `env.reset()`" cadence the docstring promises.

A few concerns worth surfacing, none blocking:

1. **Train↔eval distribution mismatch on the subgoal source itself.** Training samples a future frame from the *same* episode (uniform `[t, t+4s]` or segment end); this PR samples the *last frame of a different episode* matched by language. The PR docstring acknowledges this as naive, which is fair — but the only evidence the flag does the right thing is "the missing-key warning disappears", which is a wiring check, not a correctness check. An A/B eval delta (with vs. without `--env.subgoal_source`) on the published `pi07_low_level` / `pi07_paligemma_low_level` checkpoints would close that loop. See inline comment on `subgoal.py:118`.

2. **`random.choice` reproducibility under `max_parallel_tasks > 1`.** Process-global `random` state is shared across `eval_policy_all` worker threads, so seed-based reproducibility silently breaks for multi-task LIBERO eval. Single-threaded eval is fine and the existing test asserts that. Cheap fix is a per-generator `random.Random(seed)` instance — details inline on `subgoal.py:216`.

3. **Unbounded `_frame_cache`.** Bounded in practice by total episodes accessed during an eval pass (~660 MB worst case on full LIBERO), but no eviction; the generator survives every mid-training eval block for the full training run. Inline on `subgoal.py:189`.

4. **Implicit revision pinning.** `LeRobotDatasetMetadata` defaults to `CODEBASE_VERSION`, which matches `TensorAuto/libero`'s current v2.1 — but a drift on either side would silently swap revisions out from under the eval. Inline on `subgoal.py:150`.

5. **Terminal-observation block** in `rollout()` calls `add_subgoal_images` unconditionally; gated only by the loop-body running at least once. Theoretical edge case but the symmetry is worth tightening. Inline on `eval.py:208`.

CI shows pre-commit green, CPU tests still in progress at review time. The 4 pre-existing LIBERO-config-needs-stdin test failures the PR description calls out reproduce identically on `main` here, so those are not introduced by this PR.

---

shuheng-liu:
Thanks for the thorough review. Addressed in [a055731](https://github.com/TensorAuto/OpenTau/commit/a055731):

**Fixed**

- **(2) Per-generator `random.Random(seed)`.** `LiberoLastFrameSubgoalGenerator(..., seed=...)` now uses an instance RNG; `make_subgoal_generator` threads `cfg.seed` through. Added `test_global_random_state_does_not_affect_picks` (burns the global RNG between picks, asserts unchanged outcome) and rewrote `test_deterministic_with_seeded_random` against the instance RNG. Did *not* add a per-thread RNG for `max_parallel_tasks > 1` — `LiberoEnv.max_parallel_tasks` defaults to 1 and the existing threadlocal `chosen_episodes` already isolates outcomes; per-thread bit-identical reproducibility under threaded eval is a deeper redesign and out of scope here.

- **(4) Revision pinning.** Added an optional `revision: str | None = None` arg to `LiberoLastFrameSubgoalGenerator` and `make_subgoal_generator` now logs `generator.meta.revision` at INFO alongside resolution / num_cams / seed, so a `CODEBASE_VERSION` ↔ source-repo drift surfaces in the eval log.

- **(5) Terminal-observation block.** Now gated on `subgoal_episode_picked` in `rollout()`, mirroring the loop body. Pragmatically unreachable, but the symmetry is correct.

**Push back**

- **(1) Train↔eval distribution shift.** Agreed it's a different conditional distribution — but the naive last-frame-of-a-random-matching-episode behavior was the explicit ask for this PR ("for now, we will do a naive one"). The within-episode sampling / same-episode-as-rollout variants are exactly the follow-up I had in mind. Added a `Train↔eval distribution shift` block to the class docstring spelling this out so config writers can't trust the flag implicitly. The A/B eval delta is the right next milestone — separate PR.

- **(3) Unbounded `_frame_cache`.** Kept unbounded for this PR, with a `Memory` block in the docstring documenting the ~660 MB worst case on full LIBERO and a TODO marker for non-LIBERO repos. You agreed it's not blocking at this scale; LRU adds a parameter and a knob that I'd rather wire up the first time a real non-LIBERO use case asks for it.

19 unit tests all pass (the 2 new RNG-isolation tests plus the existing 17). Pre-commit clean.

src/opentau/envs/subgoal.py
shuheng-liu: It would be impossible to return something that's 4 sec into the future – we can't track how far along an episode we are during inference.

shuheng-liu: You're right about the 4s-ahead suggestion — strike that from suggestion #1. At inference there's no notion of "how far along the recorded trajectory we are", so a temporal-offset sampler has nothing to anchor `t` to. The viable knobs are which *episode* to pull from and which *frame within it*; the temporal-offset is only meaningful at train time.

So the realistic follow-up axis is the frame-within-episode (uniform vs. last-frame vs. some heuristic like "frame at the same fractional progress as the rollout") rather than mimicking the training distribution. Agreed that's a separate PR and not in scope here.

Docstring addition is a good landing pad for the rationale — thanks. Fixes look clean; verified the per-instance RNG, revision arg, and terminal-block gate all behave as advertised, and the 19 unit tests pass locally.

---

shuheng-liu:
## Review conclusion: ready to merge from the code-review side

All five concerns from the initial review are landed or have a reasonable resolution on `a055731`:

| # | Finding | Status |
|---|---------|--------|
| 1 | Train↔eval distribution shift | Documented as intentional naive baseline; A/B eval delta is a separate-PR follow-up (agreed). My "4 s lookahead" suggestion was wrong — `t` has no anchor at inference. The realistic follow-up axis is frame-within-episode, not temporal-offset. |
| 2 | `random.choice` reproducibility | **Fixed.** Per-instance `random.Random(seed)`; `make_subgoal_generator` threads `cfg.seed` through. New `test_global_random_state_does_not_affect_picks` burns the global RNG between picks and asserts the outcome is unchanged. |
| 3 | Unbounded `_frame_cache` | Documented with a TODO. ~660 MB worst case on full LIBERO is the realistic ceiling today; LRU lands the first time a non-LIBERO repo asks for it. |
| 4 | Revision pinning | **Fixed.** Optional `revision` arg on the generator; `make_subgoal_generator` logs the resolved `meta.revision` so a `CODEBASE_VERSION` ↔ source-repo drift surfaces in the eval log. |
| 5 | Terminal-observation `add_subgoal_images` | **Fixed.** Now gated on `subgoal_episode_picked` in `rollout()`, mirroring the loop body. |

**CI on `a055731`:** all green — Pre-commit Hooks, CPU Tests, check-checklist, review. Locally reproduced `tests/envs/test_subgoal.py`: 19/19 pass (the 17 original + 2 new RNG-isolation tests) in 0.82 s.

**Wiring:** verified the `make_subgoal_generator` fanout end-to-end (`eval_main`, `train.py`'s mid-training eval, `eval_policy_all` → `run_one` → `eval_one` → `eval_policy` → `rollout`); per-instance RNG, revision arg, and terminal-block gate all behave as advertised in the diff.

**Not policy-related** (touches `envs/` and `scripts/eval.py` only) — the unchecked policy box in the template is correct, so GPU pytests + nightly regression aren't required to merge.

PR is currently still in **draft**; flip to ready-for-review whenever you're ready to merge. No outstanding blockers from review.

---

PR #334, opened 2026-05-26
ci: skip redundant PR runs when push covered the same tree

shuheng-liu:
## What this does
On every PR push, both `push` and `pull_request` events fire and we currently run the CPU test suite and pre-commit hooks twice against the same tree. This PR adds a `precheck` job to `cpu_test.yml` and `pre-commit.yml` that calls the GitHub compare API; when the PR head already contains every commit on the base ref (`status` is `identical` or `ahead`), merging would fast-forward and the merge tree equals the head tree that the `push` run already validated, so the test job is skipped.

A new `status` job (`CPU Tests Status` / `Pre-commit Status`) always runs and aggregates the result — it passes when the underlying job succeeded or was skipped because of the precheck, and fails otherwise. Branch-protection required checks should be flipped from `CPU Tests` / `Pre-commit Hooks` to `CPU Tests Status` / `Pre-commit Status` so the skip path doesn't block merges.

When the PR branch is behind base (`status` is `behind` or `diverged`), the `pull_request` run still executes — it's the only one that exercises the merge commit's tree.

Label: ⚡️ Performance.

## How it was tested
- `pre-commit run --files .github/workflows/cpu_test.yml .github/workflows/pre-commit.yml` — all hooks pass, including `zizmor` (GHA security linter) and `check yaml`.
- Logic walk-through:
  - `push` / `workflow_dispatch` → precheck short-circuits with `run-tests=true` → test job runs → status passes.
  - PR up-to-date with base → compare returns `ahead`/`identical` → `run-tests=false` → test job skipped → status passes (skipped is treated as success).
  - PR behind base → compare returns `behind`/`diverged` → `run-tests=true` → test job runs → status reflects its result.
  - Compare API failure or unknown status → defaults to `run-tests=true` (fail-safe).
- End-to-end verification will land on this PR itself once CI runs: the first push triggers the `push` workflow; opening the PR triggers the `pull_request` workflow, which should hit the `ahead`/`identical` skip path on subsequent in-sync syncs.

```bash
git fetch origin claude/youthful-allen-HJFwj
git checkout claude/youthful-allen-HJFwj
# Inspect the two workflow files:
git diff main -- .github/workflows/cpu_test.yml .github/workflows/pre-commit.yml
```
Then watch this PR's Actions tab — the `Pre-check` job should report `Compare status: ahead` (or `identical`) and the downstream `CPU Tests` / `Pre-commit Hooks` jobs should be skipped, while the `push`-triggered runs should execute normally.

## Checklist

---
_Generated by [Claude Code](https://claude.ai/code/session_012owBv9nE5Y5K7iHWqTwYuQ)_

### DIFF
--- a/.github/workflows/cpu_test.yml
+++ b/.github/workflows/cpu_test.yml
@@ -32,14 +32,55 @@ env:
   PRE_COMMIT_HOME: "/pre-commit-cache"
 
 jobs:
-  cpu-tests:
-    # Block PRs from non-org accounts to keep fork code off our runners.
-    # `push` and `workflow_dispatch` already require write access, so the
-    # short-circuit lets them through unchanged.
+  precheck:
+    # Decide whether the `pull_request` run is redundant with the `push` run.
+    # If the PR branch already contains every commit on the base ref, the
+    # merge would fast-forward and the merge tree equals the head tree that
+    # `push` already tested. For `push` / `workflow_dispatch`, always run.
     if: >-
       github.event_name != 'pull_request'
       || contains(fromJSON('["OWNER","MEMBER","COLLABORATOR"]'), github.event.pull_request.author_association)
       || github.event.pull_request.user.login == 'claude[bot]'
+    name: Pre-check
+    runs-on: ubuntu-latest
+    outputs:
+      run-tests: ${{ steps.decide.outputs.run-tests }}
+    steps:
+      - name: Decide whether to run tests
+        id: decide
+        env:
+          EVENT_NAME: ${{ github.event_name }}
+          BASE_SHA: ${{ github.event.pull_request.base.sha }}
+          HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+          REPO: ${{ github.repository }}
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          set -euo pipefail
+          if [ "$EVENT_NAME" != "pull_request" ]; then
+            echo "Event=$EVENT_NAME; always run tests."
+            echo "run-tests=true" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+          status=$(curl -fsSL \
+            -H "Authorization: Bearer $GH_TOKEN" \
+            -H "Accept: application/vnd.github+json" \
+            "https://api.github.com/repos/${REPO}/compare/${BASE_SHA}...${HEAD_SHA}" \
+            | jq -r '.status' || echo "unknown")
+          echo "Compare status: $status"
+          case "$status" in
+            identical|ahead)
+              echo "Head contains base; push run already covered this tree. Skipping."
+              echo "run-tests=false" >> "$GITHUB_OUTPUT"
+              ;;
+            *)
+              echo "Head does not strictly contain base; running tests."
+              echo "run-tests=true" >> "$GITHUB_OUTPUT"
+              ;;
+          esac
+
+  cpu-tests:
+    needs: precheck
+    if: needs.precheck.outputs.run-tests == 'true'
     name: CPU Tests
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/cpu_test.yml
```diff
@@ -120,3 +161,28 @@ jobs:
...
+        env:
+          PRECHECK_RESULT: ${{ needs.precheck.result }}
+          TESTS_RESULT: ${{ needs.cpu-tests.result }}
+        run: |
+          set -euo pipefail
+          echo "precheck=$PRECHECK_RESULT cpu-tests=$TESTS_RESULT"
+          if [ "$PRECHECK_RESULT" != "success" ]; then
+            echo "Pre-check did not succeed."
+            exit 1
+          fi
```
shuheng-liu, line 198: Regression for fork / non-org-member PRs.

`precheck` carries the `OWNER/MEMBER/COLLABORATOR` author-association gate. When a PR comes from someone outside that set, the job-level `if:` evaluates to false, so `needs.precheck.result == "skipped"`. This branch then exits 1, and `CPU Tests Status` — which the PR description recommends as the new required check — becomes a hard failure.

Previously the same gate lived directly on `cpu-tests`, so a non-member PR got a skipped job, which GitHub treats as a passing required check. This PR silently converts that into a blocking failure.

Two reasonable fixes:

1. Match the handling used a few lines below for `TESTS_RESULT`:
   ```
   case "$PRECHECK_RESULT" in
     success|skipped) ;;
     *) exit 1 ;;
   esac
   ```
2. Or move the author-association `if:` off `precheck` and onto `cpu-tests` directly — `precheck` only hits the compare API and doesn't check out fork code, so it's safe to run unconditionally. Then `precheck.result` is always definitive.

Same issue exists verbatim in `.github/workflows/pre-commit.yml` (the `if [ "$PRECHECK_RESULT" != "success" ]` block).

---

shuheng-liu (commented):
Reviewed both workflow changes. Optimization itself is sound — FF check + skip-when-push-covered is the right pattern, and the fail-safe defaults (`unknown` / API failure / non-FF status → `run-tests=true`) are correct. One regression to address, plus a couple of observations.

**1. Fork / non-member PR regression** (inline on `cpu_test.yml`). The new `status` job hard-fails when `precheck.result == "skipped"`, which is exactly what happens for any PR whose author isn't OWNER/MEMBER/COLLABORATOR. Previously that gate produced a skipped `cpu-tests`, which counts as a passing required check. After this PR, the same PRs get a failing required check. Suggested fixes in the inline.

**2. Implicit assumption that `push` actually ran and passed.** Precheck infers "push already covered this tree" purely from the FF status — it never queries the push workflow's conclusion for the same SHA. In normal usage `push` (~9 min) lands after the `pull_request` status job (~10s on the skip path) so push's check is the final one branch protection sees, which is why this works in practice. Worth a one-line comment in the workflow noting the assumption. A future enhancement could check the push run's conclusion via the API before skipping (closes the edge case where push didn't fire — e.g., a tag push, or a branch excluded by a future `branches:` filter).

**3. Nit: duplication.** The precheck + status scaffolding is copy-pasted across `cpu_test.yml` and `pre-commit.yml`. Pulling it into a reusable workflow (`workflow_call`) would keep them in sync as the logic evolves. Not blocking.

Other things I checked and they look fine:
- `curl -fsSL ... | jq ... || echo unknown` under `set -euo pipefail` — pipefail propagates curl failures, the `||` correctly falls back to `unknown`, and the case statement's `*)` arm makes that fail-safe.
- Token scope: workflow declares `contents: read`, which is sufficient for the compare API.
- `needs.precheck.outputs.run-tests` evaluating empty when precheck is skipped — handled correctly by the strict `== 'true'` check on `cpu-tests` / `pre-commit`.

---

.github/workflows/cpu_test.yml
shuheng-liu: Fixed in b2574ae — went with option 2 (moved the author-association gate off `precheck` and onto `cpu-tests` / `pre-commit`, since `precheck` only calls the compare API). Fork PRs now hit the gate on the runner-heavy job, get skipped there, and the `status` aggregator passes — matches the pre-PR behaviour.

Also addressing the other two review-body points in the same commit:

**#2 (implicit `push` assumption)**: added the comment to both `precheck` blocks. Skipping the actual conclusion-check enhancement for now — the comment is sufficient to flag the assumption, and the corner cases (tag pushes, future `branches:` filter) don't apply today.

**#3 (duplication)**: leaving the copy-paste. Extracting into a `workflow_call` would add an indirection layer for two callsites whose only shared content is ~30 lines of precheck + a tiny aggregator. The two workflows have already diverged in some small ways (different display names, log messages) and pulling them apart again if they diverge further is cheap. Happy to revisit if a third workflow ever needs the same pattern.

---

shuheng-liu (commented):
Re-reviewed `b2574ae`. Fork-PR regression is fixed — `precheck` is now unconditional and the author-association gate moved onto `cpu-tests` / `pre-commit`. Walked the truth table:

| Event | precheck.run-tests | author gate | `cpu-tests` | `status` |
|---|---|---|---|---|
| push / workflow_dispatch | `true` (short-circuit) | n/a (event ≠ pull_request) | runs | reflects test result |
| PR, member, behind/diverged | `true` | pass | runs | reflects test result |
| PR, member, ahead/identical | `false` | pass | skipped | success ✓ |
| PR, non-member, any FF | true/false | fail | skipped | success ✓ |
| precheck failure/cancel | n/a | n/a | auto-skipped via `needs` | fails ✓ |

The assumption note about not querying the push run's conclusion is now in the workflow comments — sufficient for the optimization as-built. CI on this PR confirms the intended behaviour: `pull_request` run skipped tests in ~5s, `push` run is executing them.

Remaining nit (non-blocking): the precheck + status scaffolding is still duplicated across the two files. A future `workflow_call` extraction would keep them in lockstep, but fine to defer.

LGTM otherwise. (Cannot approve as the PR author.)

---

PR #337, opened 2026-05-27
feat(pi07_paligemma): skip saved normalization buffers on load

shuheng-liu:
## What this does

Adds a new policy-config knob `skip_normalization_weights: bool = False` on `PI07PaligemmaLowLevelConfig`. When `True`, `PI07PaligemmaLowLevelPolicy.from_pretrained` filters every `normalize_*` / `unnormalize_*` key out of the remapped state dict **before** `load_state_dict`, so the `dataset_stats`-initialised buffers from `__init__` survive the load. Defaults to `False`, so existing configs and in-flight runs are unaffected (the filter branch is skipped entirely).

**Motivation.** When finetuning a checkpoint whose saved normalization stats were aggregated over a different dataset mixture than the finetuning data, the saved `normalize_*.buffer_*` / `unnormalize_*.buffer_*` tensors clobber the freshly-computed `dataset_stats` at load time. Those buffers are registered as `nn.Parameter(requires_grad=False)`, so the wrong stats stay baked in for the rest of training — the model keeps normalising/unnormalising in the source-mixture geometry even though the finetuning data lives in a much narrower distribution. Caught this empirically while debugging a heterogeneous-mixture pretrain → single-domain finetune sequence where val MSE on the finetune dataset looked fine but rollouts on the finetune env hit 0% success. Stripping the saved buffers and letting the fresh `dataset_stats` take over (paired with continued finetuning so the weights re-align to the new normalized space) is what unlocks the recovery.

The new knob is opt-in. The new code path is a single conditional dict-comprehension just above `load_state_dict`, plus a `logging.info` reporting how many keys were dropped.

🗃️ Feature

### Caveats / when not to use this

- Switching the normalization mid-training only makes sense when **followed by further training** — the model was trained to predict in the source-mixture normalized space, so the layers immediately downstream of `normalize_inputs` / upstream of `unnormalize_outputs` need to re-align. Loading with this knob purely for **inference** would pin actions near the new `μ` regardless of what the model predicts. The docstring on the new field calls this out.
- Substring matching is anchored at the start of the key (`startswith("normalize_") or startswith("unnormalize_")`), so a hypothetical nested module called `model.something.normalize_internal.weight` is **not** affected — only the eight top-level Normalize / Unnormalize buffer tensors saved by the current `Normalize` / `Unnormalize` classes.

## How it was tested

- Added `TestSkipNormalizationWeights` (3 cases) in `tests/policies/test_pi07_paligemma_low_level.py`:
  - `test_default_is_false` — pins the dataclass default to `False`.
  - `test_can_be_set_true` — pins that the field is keyword-settable (catches a draccus-side renaming or removal).
  - `test_filter_strips_only_normalize_buffer_keys` — replicates the inline filter expression on a hand-built state dict containing the eight real buffer keys + a couple of model weight keys + a defensive nested `model.…normalize_inputs_internal.weight` key; asserts only the top-level `normalize_*` / `unnormalize_*` keys are dropped. Catches regressions if a future edit broadens or narrows the predicate.
- `pytest -m "not gpu" -n auto tests/policies/test_pi07_paligemma_low_level.py` — all 49 tests pass, including the 3 new ones (verified separately via `pytest ::TestSkipNormalizationWeights -v` → 3 passed).
- `pytest -m "gpu" -n 0 tests/policies/test_pi07_paligemma_low_level.py` (RTX 5090) — 1 passed, 1 skipped, 49 deselected, ~20 s. The 1 skipped is the pre-existing memory-skipped heavy pipeline test.
- `pytest -m "gpu" -n 0` (full suite, RTX 5090) — **19 passed, 10 skipped, 0 failed in 3:20**. All skips are pre-existing unconditional `@pytest.mark.skip` markers (memory or env), unaffected by this branch.
- `pre-commit run --files src/opentau/policies/pi07_paligemma/low_level/configuration_pi07_low_level.py src/opentau/policies/pi07_paligemma/low_level/modeling_pi07_low_level.py tests/policies/test_pi07_paligemma_low_level.py` — all hooks pass (ruff, ruff-format, pyupgrade, bandit, gitleaks, license-header, …).
- Nightly regression tests will land in the scheduled `regression_test.yml` cron — they haven't run yet on this branch.
- End-to-end empirical validation: enabled the knob on a real heterogeneous-mixture pretrained checkpoint, then finetuned on a single-domain dataset. Training MSE spikes ~21× at step ~40 (model encountering different-geometry normalized targets for the first time) — the expected signature of the fresh `dataset_stats` taking effect. Without the knob the same finetune sat at the source-mixture loss landscape and could not move the rollout success rate off zero in 1000 steps; with the knob the loss curve enters a new trajectory that the weights then re-align to.

```bash
git checkout claude/skip-normalization-weights
pytest -sx tests/policies/test_pi07_paligemma_low_level.py::TestSkipNormalizationWeights
```

To enable on a real finetune, set the field under `policy` in your training JSON:

```json
{
  "policy": {
    "type": "pi07_paligemma_low_level",
    "pretrained_path": "TensorAuto/<your-pretrained-checkpoint>",
    "skip_normalization_weights": true
  }
}
```

…and launch normally:

```bash
opentau-train --config_path=configs/your_finetune_config.json
```

The startup log will include `skip_normalization_weights=True; dropped 8 saved normalize/unnormalize buffer keys` if the filter fires.

## Checklist

### DIFF
--- a/src/opentau/policies/pi07_paligemma/low_level/configuration_pi07_low_level.py
+++ b/src/opentau/policies/pi07_paligemma/low_level/configuration_pi07_low_level.py
@@ -71,10 +71,20 @@ class PI07PaligemmaLowLevelConfig(PreTrainedConfig):
             ``normalize_*`` / ``unnormalize_*`` buffer tensors from the state dict before
             ``load_state_dict``. The buffers freshly initialised from ``dataset_stats`` then
             survive — use this when finetuning a checkpoint whose saved normalization stats
-            were aggregated over a different dataset mixture than the finetuning data. Note
-            that the model was trained against the saved stats; switching the normalization
-            mid-training only makes sense when followed by further training, not when
-            loading purely for inference. Defaults to False (no behaviour change).
+            were aggregated over a different dataset mixture than the finetuning data.
+            **Requires ``dataset_stats`` to be supplied to ``__init__``** (e.g. via
+            :py:func:`opentau.policies.factory.make_policy`); otherwise the buffers stay at
+            the ``inf`` sentinel from :py:func:`opentau.policies.normalize.create_stats_buffers`
+            and the next forward crashes. **One-shot:** after the strip fires successfully,
+            :py:meth:`from_pretrained` resets the flag to ``False`` on the model's config so
+            that subsequent ``save_pretrained`` / resume / inference loads do not re-strip
+            the now-correct finetuned buffers. The model was trained against the saved
+            stats, so switching the normalization mid-training only makes sense when followed
+            by further training, not when loading purely for inference. **Scope:** only the
+            ``pi07_paligemma_low_level`` load path honours this knob today; ``pi05`` /
+            ``pi05_mem`` / ``pi06`` / ``pi07/low_level`` / ``pi0`` share the same trap and
+            should grow an equivalent knob as follow-up. Defaults to False (no behaviour
+            change).
         optimizer_lr: Learning rate for the optimizer. Defaults to 2.5e-5.
         optimizer_betas: Beta parameters for AdamW optimizer. Defaults to (0.9, 0.95).
         optimizer_eps: Epsilon parameter for AdamW optimizer. Defaults to 1e-8.
--- a/src/opentau/policies/pi07_paligemma/low_level/configuration_pi07_low_level.py
+++ b/src/opentau/policies/pi07_paligemma/low_level/configuration_pi07_low_level.py
@@ -67,6 +67,14 @@ class PI07PaligemmaLowLevelConfig(PreTrainedConfig):
             backward compatibility but logs a warning and falls back to "eager".
         freeze_vision_encoder: Whether to freeze the vision encoder during fine-tuning. Defaults to True.
         train_expert_only: Whether to train only the expert module. Defaults to False.
+        skip_normalization_weights: When loading via :py:meth:`from_pretrained`, drop the saved
+            ``normalize_*`` / ``unnormalize_*`` buffer tensors from the state dict before
+            ``load_state_dict``. The buffers freshly initialised from ``dataset_stats`` then
+            survive — use this when finetuning a checkpoint whose saved normalization stats
+            were aggregated over a different dataset mixture than the finetuning data. Note
+            that the model was trained against the saved stats; switching the normalization
+            mid-training only makes sense when followed by further training, not when
+            loading purely for inference. Defaults to False (no behaviour change).
         optimizer_lr: Learning rate for the optimizer. Defaults to 2.5e-5.
         optimizer_betas: Beta parameters for AdamW optimizer. Defaults to (0.9, 0.95).
         optimizer_eps: Epsilon parameter for AdamW optimizer. Defaults to 1e-8.
@@ -147,6 +155,7 @@ class PI07PaligemmaLowLevelConfig(PreTrainedConfig):
     # Finetuning settings
     freeze_vision_encoder: bool = True
     train_expert_only: bool = False
+    skip_normalization_weights: bool = False
     # Wrap each transformer-layer forward in torch.utils.checkpoint to trade
     # ~25-33% same-batch compute for ~30-40 GB of activation memory per rank,
     # typically netting +10-25% throughput once the freed memory is spent on
--- a/src/opentau/policies/pi07_paligemma/low_level/modeling_pi07_low_level.py
+++ b/src/opentau/policies/pi07_paligemma/low_level/modeling_pi07_low_level.py
@@ -125,6 +125,27 @@ def _is_normalize_buffer_key(key: str) -> bool:
     return key.startswith("normalize_") or key.startswith("unnormalize_")
 
 
+def _find_inf_normalize_buffers(module: nn.Module) -> list[str]:
+    """Return the names of every Normalize / Unnormalize buffer parameter on
+    ``module`` that still holds the ``torch.inf`` sentinel from
+    :py:func:`~opentau.policies.normalize.create_stats_buffers`.
+
+    Used by :py:meth:`PI07PaligemmaLowLevelPolicy.from_pretrained` to detect
+    the ``skip_normalization_weights=True`` + missing ``dataset_stats``
+    combination: the strip voids the "load_state_dict will refill these"
+    branch of the ``dataset_stats`` contract, so without ``dataset_stats``
+    the buffers stay at ``inf`` and the next forward crashes inside
+    :py:meth:`~opentau.policies.normalize.Normalize.forward` with a
+    "use a pretrained model" error that actively misleads when a
+    pretrained model *was* used.
+    """
+    return [
+        name
+        for name, param in module.named_parameters()
+        if _is_normalize_buffer_key(name) and torch.isinf(param).any()
+    ]
+
+
 def create_sinusoidal_pos_embedding(
     time: Tensor, dimension: int, min_period: float, max_period: float, device: torch.device | str = "cpu"
 ) -> Tensor:
@@ -419,6 +440,11 @@ class PI07PaligemmaLowLevelPolicy(PreTrainedPolicy):
         # Now manually load and remap the state dict
         acc = get_proc_accelerator()
         is_main_process = acc.is_main_process if acc else True
+        # Tracks whether the strip branch fired inside the try block below.
+        # Used outside the try/except to gate the inf-buffer guard so the
+        # ValueError is not swallowed by the broad `except Exception` that
+        # otherwise just warns and returns the (broken) model.
+        strip_ran = False
         try:
             # Try to load the pytorch_model.bin or model.safetensors file
             if is_main_process:
@@ -495,6 +521,7 @@ class PI07PaligemmaLowLevelPolicy(PreTrainedPolicy):
                 # persists False, and later resumes / inference loads do not
                 # re-strip the now-correct finetuned buffers.
                 model.config.skip_normalization_weights = False
+                strip_ran = True
 
             # Load the remapped state dict into the model
             missing_keys, unexpected_keys = model.load_state_dict(remapped_state_dict, strict=False)
@@ -525,6 +552,29 @@ class PI07PaligemmaLowLevelPolicy(PreTrainedPolicy):
         except Exception as e:
             if is_main_process:
                 logging.warning("Could not remap state dict keys: %s", e)
+
+        # Outside the try/except so the ValueError is not swallowed.
+        # Gated on `strip_ran` so this only fires when the load path
+        # actually reached and executed the strip branch — i.e. the user
+        # asked for skip_normalization_weights AND the load succeeded far
+        # enough to apply it. The strip voids the "load_state_dict will
+        # refill these" half of the dataset_stats contract, so without
+        # dataset_stats wired into __init__, the buffers stay at the
+        # torch.inf sentinel from create_stats_buffers and the next
+        # forward crashes inside Normalize.forward with a "use a
+        # pretrained model" message that actively misleads here.
+        if strip_ran:
+            inf_buffers = _find_inf_normalize_buffers(model)
+            if inf_buffers:
+                raise ValueError(
+                    "skip_normalization_weights=True requires `dataset_stats` to be "
+                    "passed to __init__ (e.g. via "
+                    "`opentau.policies.factory.make_policy(..., ds_meta=...)`) so "
+                    "the fresh buffers can replace the stripped saved ones; got "
+                    f"{len(inf_buffers)} uninitialized (inf) buffer(s): "
+                    + ", ".join(inf_buffers[:5])
+                    + (f", ... ({len(inf_buffers) - 5} more)" if len(inf_buffers) > 5 else "")
+                )
 
         return model
 
--- a/src/opentau/policies/pi07_paligemma/low_level/modeling_pi07_low_level.py
+++ b/src/opentau/policies/pi07_paligemma/low_level/modeling_pi07_low_level.py
@@ -451,6 +451,23 @@ class PI07PaligemmaLowLevelPolicy(PreTrainedPolicy):
             if remap_count > 0 and is_main_process:
                 logging.info("Remapped %d state dict keys", remap_count)
 
+            # Strip saved normalize/unnormalize buffers so the
+            # ``dataset_stats``-initialised buffers from ``__init__`` survive
+            # the load. ``requires_grad=False`` on those tensors means training
+            # alone cannot recover from inheriting the wrong stats.
+            if config.skip_normalization_weights:
+                n_before = len(remapped_state_dict)
+                remapped_state_dict = {
+                    key: val
+                    for key, val in remapped_state_dict.items()
+                    if not (key.startswith("normalize_") or key.startswith("unnormalize_"))
+                }
+                if is_main_process:
+                    logging.info(
+                        "skip_normalization_weights=True; dropped %d saved normalize/unnormalize buffer keys",
+                        n_before - len(remapped_state_dict),
+                    )
+
             # Load the remapped state dict into the model
             missing_keys, unexpected_keys = model.load_state_dict(remapped_state_dict, strict=False)
 
--- a/src/opentau/policies/pi07_paligemma/low_level/modeling_pi07_low_level.py
+++ b/src/opentau/policies/pi07_paligemma/low_level/modeling_pi07_low_level.py
@@ -473,16 +473,23 @@ class PI07PaligemmaLowLevelPolicy(PreTrainedPolicy):
             # ``dataset_stats``-initialised buffers from ``__init__`` survive
             # the load. ``requires_grad=False`` on those tensors means training
             # alone cannot recover from inheriting the wrong stats.
+            stripped_keys: frozenset[str] = frozenset()
             if config.skip_normalization_weights:
-                n_before = len(remapped_state_dict)
+                stripped_keys = frozenset(key for key in remapped_state_dict if _is_normalize_buffer_key(key))
                 remapped_state_dict = {
-                    key: val for key, val in remapped_state_dict.items() if not _is_normalize_buffer_key(key)
+                    key: val for key, val in remapped_state_dict.items() if key not in stripped_keys
                 }
                 if is_main_process:
-                    logging.info(
-                        "skip_normalization_weights=True; dropped %d saved normalize/unnormalize buffer keys",
-                        n_before - len(remapped_state_dict),
-                    )
+                    if not stripped_keys:
+                        logging.warning(
+                            "skip_normalization_weights=True but no normalize_/unnormalize_ "
+                            "keys were present in the saved state dict; the flag had no effect."
+                        )
+                    else:
+                        logging.info(
+                            "skip_normalization_weights=True; dropped %d saved normalize/unnormalize buffer keys",
+                            len(stripped_keys),
+                        )
                 # One-shot semantics: the flag is consumed by this load.
                 # Reset on the model's config so the next save_pretrained()
                 # persists False, and later resumes / inference loads do not
@@ -492,12 +499,18 @@ class PI07PaligemmaLowLevelPolicy(PreTrainedPolicy):
             # Load the remapped state dict into the model
             missing_keys, unexpected_keys = model.load_state_dict(remapped_state_dict, strict=False)
 
-            if missing_keys and is_main_process:
-                logging.warning("Missing keys when loading state dict: %d keys", len(missing_keys))
-                for key in missing_keys[:20]:
+            # Hide deliberately-stripped buffer keys from the missing-keys
+            # warning so the noisy WARNING does not directly contradict the
+            # INFO logged just above. ``stripped_keys`` is empty when the
+            # flag is off, so this is a no-op for default loads.
+            unintended_missing = [key for key in missing_keys if key not in stripped_keys]
+
+            if unintended_missing and is_main_process:
+                logging.warning("Missing keys when loading state dict: %d keys", len(unintended_missing))
+                for key in unintended_missing[:20]:
                     logging.warning("  - %s", key)
-                if len(missing_keys) > 20:
-                    logging.warning("  ... and %d more", len(missing_keys) - 20)
+                if len(unintended_missing) > 20:
+                    logging.warning("  ... and %d more", len(unintended_missing) - 20)
 
             if unexpected_keys and is_main_process:
                 logging.warning("Unexpected keys when loading state dict: %d keys", len(unexpected_keys))
@@ -506,7 +519,7 @@ class PI07PaligemmaLowLevelPolicy(PreTrainedPolicy):
                 if len(unexpected_keys) > 20:
                     logging.warning("  ... and %d more", len(unexpected_keys) - 20)
 
-            if not missing_keys and not unexpected_keys and is_main_process:
+            if not unintended_missing and not unexpected_keys and is_main_process:
                 logging.info("All keys loaded successfully!")
 
         except Exception as e:
--- a/src/opentau/policies/pi07_paligemma/low_level/modeling_pi07_low_level.py
+++ b/src/opentau/policies/pi07_paligemma/low_level/modeling_pi07_low_level.py
@@ -107,6 +107,24 @@ def _preferred_dtype():
     return torch.float32 if torch.onnx.is_in_onnx_export() else torch.bfloat16
 
 
+def _is_normalize_buffer_key(key: str) -> bool:
+    """Return True iff ``key`` is a top-level Normalize / Unnormalize buffer tensor
+    saved by :class:`~opentau.policies.normalize.Normalize` /
+    :class:`~opentau.policies.normalize.Unnormalize`.
+
+    Anchored at the start of ``key`` so a hypothetical nested submodule named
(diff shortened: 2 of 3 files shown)


tests/policies/test_pi07_paligemma_low_level.py
```diff
@@ -2399,3 +2399,65 @@ def test_fps_mixed_pad_across_batch(self):
...
+            "state_proj.weight": 10,
+            # Defensive: a hypothetical nested key containing "normalize"
+            # below the top level must NOT be stripped — the filter is
+            # anchored at the start of the key.
+            "model.some_layer.normalize_inputs_internal.weight": 11,
+        }
+        filtered = {
+            key: val
+            for key, val in state_dict.items()
+            if not (key.startswith("normalize_") or key.startswith("unnormalize_"))
```
shuheng-liu: **Tautology — this test cannot catch the regression its docstring claims to catch.**

The test inlines its own copy of the filter predicate (`if not (key.startswith("normalize_") or key.startswith("unnormalize_"))`) rather than calling the production code. If a future edit changes the predicate inside `PI07PaligemmaLowLevelPolicy.from_pretrained` — e.g. to `startswith("model.normalize_")`, or to a regex, or to a learned-buffer allowlist — this test still passes, because it never imports or invokes that predicate. The docstring above promises the opposite: "a future edit that broadens or narrows the predicate trips this test before landing in production."

Two ways to make the test load-bearing:

1. **Extract the predicate to a module-level helper** in `modeling_pi07_low_level.py`:

   ```python
   def _is_normalize_buffer_key(key: str) -> bool:
       return key.startswith("normalize_") or key.startswith("unnormalize_")
   ```

   Use it in `from_pretrained`, then import it in the test and assert against `_is_normalize_buffer_key(...)` directly. The test now fails the moment the helper changes shape.

2. **Drive `from_pretrained` end-to-end** with a tiny fake checkpoint (safetensors with hand-built tensors), then read `policy.normalize_inputs.buffer_*` and confirm the values match `dataset_stats`, not the saved tensors. Heavier, but actually exercises the integration.

Option 1 is cheaper and catches predicate-shape regressions; option 2 also catches "filter never runs because the `if` was deleted." I'd take option 1 at minimum.

---

src/opentau/policies/pi07_paligemma/low_level/modeling_pi07_low_level.py
```diff
@@ -451,6 +451,23 @@ def from_pretrained(
...
+            # ``dataset_stats``-initialised buffers from ``__init__`` survive
+            # the load. ``requires_grad=False`` on those tensors means training
+            # alone cannot recover from inheriting the wrong stats.
+            if config.skip_normalization_weights:
+                n_before = len(remapped_state_dict)
+                remapped_state_dict = {
+                    key: val
+                    for key, val in remapped_state_dict.items()
+                    if not (key.startswith("normalize_") or key.startswith("unnormalize_"))
+                }
```
shuheng-liu, line 507: **Silent runtime trap: `skip_normalization_weights=True` + missing `dataset_stats` → `inf` buffers.**

The documented contract on `__init__` (`Args.dataset_stats`) is: *"If not passed here, it is expected that they will be passed with a call to `load_state_dict` before the policy is used."* This PR's filter voids the second half of that contract — when the flag is on, `load_state_dict` will *not* refill the buffers either (they're stripped). If a caller follows the documented pattern and constructs the policy without `dataset_stats` (e.g. any non-`make_policy` path — inference scripts, notebooks, the gRPC server, a fresh `PI07PaligemmaLowLevelPolicy.from_pretrained(...)` call), `create_stats_buffers` leaves the buffers at `torch.inf` and the next forward crashes on the `assert not torch.isinf(mean).any()` in `normalize.py:250`.

The crash isn't silent, but its message ("`mean` is infinity. You should either initialize with `stats` as an argument, or use a pretrained model.") actively misleads — the user *is* using a pretrained model, they just enabled the flag that undid it.

Cheap guard right here:

```python
if config.skip_normalization_weights:
    if not _model_has_finite_norm_buffers(model):
        raise ValueError(
            "skip_normalization_weights=True requires dataset_stats to be passed "
            "to __init__ so the fresh buffers can replace the stripped saved ones; "
            "got uninitialized (inf) buffers instead."
        )
    ...
```

Or, less invasive: at minimum extend the new docstring on `skip_normalization_weights` to call out that `dataset_stats` must be supplied to `__init__` when the flag is on, and link it to the failure mode.

---

src/opentau/policies/pi07_paligemma/low_level/configuration_pi07_low_level.py
```diff
@@ -147,6 +155,7 @@ class PI07PaligemmaLowLevelConfig(PreTrainedConfig):
     # Finetuning settings
     freeze_vision_encoder: bool = True
     train_expert_only: bool = False
+    skip_normalization_weights: bool = False
```
shuheng-liu, line 168: **Altitude: this fix solves the bug for one of 5+ policies with the exact same trap.**

`pi05`, `pi05_mem`, `pi06`, `pi07/low_level/`, and `pi0` all share the `Normalize` / `Unnormalize` pattern with `requires_grad=False` buffer parameters (confirmed by `grep "self.normalize_inputs = Normalize" src/opentau/policies/*/modeling_*.py`). The same checkpoint→finetune mixture-mismatch you hit on `pi07_paligemma_low_level` will repro on any of them — and when the next person hits it, they'll either reinvent this knob (drift) or paste it (copy-pasted bug surface).

Two cheaper-to-maintain alternatives:

1. **Lift to `PreTrainedPolicy.from_pretrained`** (`src/opentau/policies/pretrained.py`). Every policy that delegates to it gets the knob; `PI07PaligemmaLowLevelPolicy.from_pretrained` already overrides and could just call `super().from_pretrained(...)` after the remap.

2. **Lift to `Normalize` / `Unnormalize`** — give them an opt-in `_load_from_state_dict` hook that no-ops when a flag is set on the parent config, so the buffer-vs-state-dict reconciliation happens where the buffer lives.

Both are larger PRs and may be out of scope for "ship the bugfix you have." If so, that's fine — but I'd ask that the docstring on `skip_normalization_weights` add a sentence noting "this only applies to `pi07_paligemma_low_level`; equivalent fixes for `pi05`/`pi06`/`pi07` are tracked as follow-up," and that you open a tracking issue. Otherwise the next finetune that hits this will silently corrupt val→rollout for the same number of hours yours did.

---

src/opentau/policies/pi07_paligemma/low_level/configuration_pi07_low_level.py
```diff
@@ -67,6 +67,14 @@ class PI07PaligemmaLowLevelConfig(PreTrainedConfig):
...
         freeze_vision_encoder: Whether to freeze the vision encoder during fine-tuning. Defaults to True.
         train_expert_only: Whether to train only the expert module. Defaults to False.
+        skip_normalization_weights: When loading via :py:meth:`from_pretrained`, drop the saved
+            ``normalize_*`` / ``unnormalize_*`` buffer tensors from the state dict before
+            ``load_state_dict``. The buffers freshly initialised from ``dataset_stats`` then
+            survive — use this when finetuning a checkpoint whose saved normalization stats
+            were aggregated over a different dataset mixture than the finetuning data. Note
+            that the model was trained against the saved stats; switching the normalization
+            mid-training only makes sense when followed by further training, not when
+            loading purely for inference. Defaults to False (no behaviour change).
```
shuheng-liu: **Flag persists in saved configs — resume and inference are foot-guns.**

`PreTrainedConfig._save_pretrained` (`policies.py:444`) writes the full encoded dataclass to `config.json`, so once a finetune starts with `skip_normalization_weights=True`, every checkpoint it saves carries the flag set. Three downstream surprises follow:

1. **Resume.** Per the CLAUDE.md rule "when resuming, the checkpoint config takes precedence over CLI args", a resumed finetune loads with the flag still True. The freshly-finetuned (and now *correct*) `normalize_*.buffer_*` tensors get stripped on every reload and replaced from the resume-time `dataset_stats`. If those stats are the same dataset, fine — no-op. If the user resumes against a slightly different shard / different epoch slice, the buffers move under their feet.

2. **Inference.** If someone later loads the finetune checkpoint for inference (no `dataset_stats` wired through, since inference flows don't carry `ds_meta`), the flag fires and the buffers go to `inf` → crash. Your own docstring warns against using this knob for inference, but the *config persistence* makes that warning very hard to comply with — the user has to remember to manually flip the flag back to `False` before any non-training load.

3. **Hub publish.** If the finetune checkpoint ships to a HuggingFace hub repo, every downstream consumer who calls `from_pretrained(<hub-id>)` inherits the flag silently.

The cleaner shape is to make this a **`from_pretrained` kwarg** rather than a permanent config field — it's a load-time choice, not a property of the model architecture. Something like:

```python
policy = PI07PaligemmaLowLevelPolicy.from_pretrained(
    path, skip_normalization_weights=True,
)
```

…then route it through `make_policy` via an explicit param. The field disappears from `config.json` entirely, and the resume/inference traps go with it.

If you keep the config-field shape, at minimum: in `_save_pretrained` (or in `train.py`'s save path), force-reset `skip_normalization_weights=False` on the *saved* config — the strip already happened once, the saved checkpoint's buffers are now the right ones, future loads should not re-strip.

---

shuheng-liu (commented):
Reviewed at max effort (5 correctness + 3 cleanup + 1 altitude angle, sweep pass). The change itself is small, well-scoped, and the predicate is correct for the saved buffer names — I have no correctness bugs to flag in the production filter.

Four improvement findings inline, ranked by severity:

1. **Test tautology** (`test_filter_strips_only_normalize_buffer_keys`) — the test inlines its own copy of the filter and so cannot catch the regression its docstring claims to catch. Extract the predicate to a shared helper or invoke `from_pretrained` end-to-end.
2. **Inf-buffer trap** when `skip_normalization_weights=True` and `dataset_stats` is not passed (any non-`make_policy` load path — inference, gRPC server, notebooks). The resulting assertion message in `normalize.py` actively misleads.
3. **Altitude** — pi05 / pi05_mem / pi06 / pi07 (non-paligemma) / pi0 all share the same `Normalize` / `Unnormalize` trap. Either lift to `PreTrainedPolicy.from_pretrained` / the `Normalize` classes, or open a tracking issue and call this out in the field docstring.
4. **Config-persistence hazard** — the flag is written into every saved `config.json`, so resume and inference loads silently re-strip the (now correct) finetuned buffers. Either move to a `from_pretrained` kwarg, or force-reset to `False` in the saved config after the first load.

None of these block the bugfix landing — the empirical evidence (loss-curve shift + rollout success unlock) is convincing — but #1 and #4 are cheap to address before merge and prevent the same trap re-firing for the next person.

---

tests/policies/test_pi07_paligemma_low_level.py
shuheng-liu: Addressed in 507e02d via option 1 (extract module-level helper). The predicate now lives at `modeling_pi07_low_level._is_normalize_buffer_key`; the production filter calls it (one source of truth), and the test imports it directly. Replaced the single tautological test with two: `test_predicate_matches_all_eight_saved_buffer_keys` (pins the current eight saved buffer names) and `test_predicate_anchored_at_key_start` (defends the `startswith` anchor against future regex / contains-based rewrites). A predicate-shape change now trips both before landing.

src/opentau/policies/pi07_paligemma/low_level/configuration_pi07_low_level.py
shuheng-liu: Addressed in 507e02d via the second of your two suggestions (one-shot reset, since the load-time-kwarg version touches `make_policy` + every override). After the strip fires successfully, `from_pretrained` now sets `model.config.skip_normalization_weights = False` so the next `save_pretrained` persists `False`, and later resume / inference / hub-publish loads do not re-strip the now-correct finetuned buffers. Docstring spells out the one-shot semantic so users know the flag is consumed on first use. Caveat: `train.py`s own `train_config.json` save path captures the *input* `cfg.policy.skip_normalization_weights = True` — that file still carries `True` after the first run. Resume reads the checkpoint-dir `config.json` (which now says `False`) for the policy and the `train_config.json` for the training-pipeline knobs, so the actual policy-load path is safe; if you want the training config to also self-clean I can route that through `train.py` in a follow-up.

src/opentau/policies/pi07_paligemma/low_level/configuration_pi07_low_level.py
shuheng-liu: Took the "ship the bugfix you have + open a tracking issue" path you offered. Docstring on the field now explicitly calls out the scope ("only the `pi07_paligemma_low_level` load path honours this knob today; `pi05` / `pi05_mem` / `pi06` / `pi07/low_level` / `pi0` share the same trap and should grow an equivalent knob as follow-up"). Tracking issue #338 captures the broader fix with the two design options you sketched (lift to `PreTrainedPolicy.from_pretrained` vs lift to `Normalize` / `Unnormalize`) and an acceptance checklist.

src/opentau/policies/pi07_paligemma/low_level/modeling_pi07_low_level.py
shuheng-liu: Partially addressed in 507e02d via your "less invasive" option — the field docstring now explicitly says `dataset_stats` must be supplied to `__init__` when the flag is on, and that the failure mode is the `inf`-mean assertion in `normalize.py`. Did *not* add the in-`from_pretrained` runtime guard yet because the assertion in `Normalize.forward` already catches it at first forward and the cleaner home for that check is in the broader fix tracked in #338 (option 2 there is to do the buffer-vs-state-dict reconciliation inside `Normalize` / `Unnormalize` themselves, which makes the `inf`-detection live next to the assertion site and lets every policy benefit from one place). If you want the cheap guard in this PR anyway I can add it — the docstring without the guard is still a meaningful improvement over silently-misleading, but I agree a clearer error is strictly better than the existing `Normalize.forward` assertion message.

src/opentau/policies/pi07_paligemma/low_level/modeling_pi07_low_level.py
```diff
@@ -451,6 +469,26 @@ def from_pretrained(
...
                 logging.info("Remapped %d state dict keys", remap_count)
 
+            # Strip saved normalize/unnormalize buffers so the
+            # ``dataset_stats``-initialised buffers from ``__init__`` survive
+            # the load. ``requires_grad=False`` on those tensors means training
+            # alone cannot recover from inheriting the wrong stats.
+            if config.skip_normalization_weights:
+                n_before = len(remapped_state_dict)
+                remapped_state_dict = {
+                    key: val for key, val in remapped_state_dict.items() if not _is_normalize_buffer_key(key)
```
shuheng-liu: **Second-pass finding: the stripped buffer keys reappear as MISSING warnings on the very next line.**

`model.load_state_dict(remapped_state_dict, strict=False)` returns `missing_keys` for every parameter the model has but the state dict doesn't. The PR just deliberately deleted ~8 buffer keys from the state dict, so on the immediately-following lines (474-479) those exact 8 keys will be re-dumped as:

```
WARNING: Missing keys when loading state dict: 8 keys
WARNING:   - normalize_inputs.buffer_observation_state.mean
WARNING:   - normalize_inputs.buffer_observation_state.std
WARNING:   - normalize_targets.buffer_action.mean
... (etc)
```

This is noise that directly contradicts the `INFO` you just logged on line 466-468 ("dropped 8 saved normalize/unnormalize buffer keys") — and on a real run with N other genuinely-missing keys mixed in, it makes the genuinely-missing list noisier and harder to scan. Worse: someone reading the log later who didn't write this code will think the strip is buggy.

Cheap fix here (after the filter, before `load_state_dict`):

```python
if config.skip_normalization_weights:
    stripped_keys = {k for k in self.state_dict() if k.startswith(("normalize_", "unnormalize_"))}
    # ... then below ...
    missing_keys = [k for k in missing_keys if k not in stripped_keys]
```

Or simpler: log the stripped names at `INFO`/`DEBUG` and have the WARNING list filter them out by the same predicate. The "All keys loaded successfully!" branch at line 488 should also account for this — currently it never fires when the flag is on.

---

tests/policies/test_pi07_paligemma_low_level.py
```diff
@@ -2399,3 +2400,69 @@ def test_fps_mixed_pad_across_batch(self):
...
+            "normalize_inputs.buffer_state.std",
+            "normalize_targets.buffer_actions.mean",
+            "normalize_targets.buffer_actions.std",
+            "unnormalize_outputs.buffer_actions.mean",
+            "unnormalize_outputs.buffer_actions.std",
+            "normalize_discrete_actions.buffer_actions.min",
+            "normalize_discrete_actions.buffer_actions.max",
+        ]
+        for key in saved_buffer_keys:
+            assert _is_normalize_buffer_key(key), f"{key!r} should be stripped but predicate returned False"
```
shuheng-liu: **Second-pass finding: test uses fabricated buffer key shapes that don't match production.**

The real saved keys are built in `normalize.py:215` as `setattr(self, "buffer_" + key.replace(".", "_"), buffer)` against feature keys like `observation.state`, `observation.images.top`, `action`. So the actual top-level shapes are:

- `normalize_inputs.buffer_observation_state.mean`
- `normalize_inputs.buffer_observation_state.std`
- `normalize_inputs.buffer_observation_images_top.mean`
- `normalize_targets.buffer_action.mean`
- `unnormalize_outputs.buffer_action.mean`
- ... etc.

This test uses `buffer_state`, `buffer_actions` (note the plural — the real key is singular `action` per `dataset_to_policy_features`'s output, modulo the feature name in the source). The prefix-anchored filter happens to strip all of these correctly *because* it only looks at the `normalize_` / `unnormalize_` leading segment — but that means the test cannot catch:

1. A future refactor that moves the feature naming from `observation.state` to (say) `obs.state` or drops the leading `observation.`, breaking nothing in the filter but breaking everything that downstream consumers expect from the saved checkpoint.
2. A regression in `Normalize.__init__`'s `setattr` line that changes the `buffer_` prefix or the `.` → `_` mangling.
3. A typo that uses a different feature-key shape than what's actually saved.

If the goal of this test is "pin the filter rule," the simplest fix is to use real feature-key shapes — instantiate `Normalize(config.input_features, config.normalization_mapping, dataset_stats=fake_stats)`, then call `.state_dict()` and use *that* dict as the input to the filter. That way the test grounds the predicate against the actual key shape the buffers will produce, and any change to either side breaks it.

---

src/opentau/policies/pi07_paligemma/low_level/modeling_pi07_low_level.py
shuheng-liu, line 1: **Second-pass finding (logging branch around lines 465-469): silent no-op when the flag is on but the saved checkpoint has no normalize buffers.**

If `n_before == len(remapped_state_dict)` — i.e. the filter dropped *zero* keys, which happens when loading raw upstream PaliGemma weights, openpi pre-merge checkpoints, or any saved file that predates `Normalize` buffers in this layout — the log unconditionally prints `dropped 0 saved normalize/unnormalize buffer keys` and the run proceeds as if the knob succeeded. A user who flipped the flag *because* they wanted source-mixture stats voided gets no signal that the flag was effectively dead — they train for hours and rediscover the mismatch when rollouts fail.

Cheap fix: branch on `n_dropped == 0` and warn-and-continue instead of the unconditional INFO.

```python
n_dropped = n_before - len(remapped_state_dict)
if is_main_process:
    if n_dropped == 0:
        logging.warning(
            "skip_normalization_weights=True but no normalize_/unnormalize_ keys "
            "were present in the saved state dict; the flag had no effect."
        )
    else:
        logging.info(
            "skip_normalization_weights=True; dropped %d saved normalize/unnormalize buffer keys",
            n_dropped,
        )
```

(Couldn't post inline on the log line — GitHub rejected the line as out of diff range. Filed as file-level instead.)

---

shuheng-liu (commented):
Second pass — looking for gaps from the first review, not re-derivations. Three new findings, all secondary to the four from the first review:

1. **Stripped buffer keys re-surface as MISSING warnings** (modeling, lines 472-479). The `load_state_dict(strict=False)` call immediately after the filter reports the just-stripped keys as missing, and the existing warning loop dumps them. The user explicitly chose to drop these — the WARNING noise contradicts the success INFO logged one line earlier.

2. **Silent no-op when flag is on but no normalize keys were in the saved checkpoint** (modeling, log around line 465-469). The log still prints "dropped 0 keys" — no signal to the user that the flag was effectively dead. (Posted as file-level comment because the inline-line API kept rejecting the lines as out-of-diff-range; the comment names the exact range.)

3. **Test buffer key shapes don't match production naming** (test, lines 2438-2451). Real saved keys use `buffer_observation_state`/`buffer_action` (per the `setattr` in `Normalize.__init__:215`); the test uses `buffer_state`/`buffer_actions`. The filter happens to work because of the prefix match, but the test can't catch a regression in the feature-key → buffer-name construction. Suggest grounding the test against `Normalize(...).state_dict()` instead of a hand-built dict.

None block landing. Combined with the first review's findings, the highest-value cleanups before merge are: extract the filter to a helper that both production and test share (covers first review #1 + this pass's #3 at once), and add a warning when the strip is a no-op (#2 here).

---

src/opentau/policies/pi07_paligemma/low_level/modeling_pi07_low_level.py
shuheng-liu: Addressed in 396e67e. `from_pretrained` now tracks `stripped_keys` (a `frozenset` populated from the same predicate, empty when the flag is off so the default path is unchanged), filters them out of `missing_keys` before the WARNING block, and uses the filtered `unintended_missing` list when deciding whether to fire the "All keys loaded successfully" INFO. The intentional drops no longer surface as warning noise.

src/opentau/policies/pi07_paligemma/low_level/modeling_pi07_low_level.py
shuheng-liu: Addressed in 396e67e. The strip block now branches on whether any keys actually matched the predicate: a non-empty match logs INFO with the count, an empty match logs a WARNING ("flag had no effect" — explicit signal so the user does not silently train against the wrong stats believing the flag fired). Reset of `model.config.skip_normalization_weights = False` still happens in either path so the flag is consumed regardless.

tests/policies/test_pi07_paligemma_low_level.py
shuheng-liu: Addressed in 396e67e via your "ground against `Normalize(...).state_dict()` instead of a hand-built dict" suggestion. The test now: (1) instantiates `Normalize` and `Unnormalize` with a feature dict that uses real-shape keys (`observation.state`, `observation.images.top`, `action`); (2) reads `.state_dict()` from each; (3) prepends each of the four production attribute prefixes (`normalize_inputs.`, `normalize_targets.`, `unnormalize_outputs.`, `normalize_discrete_actions.`); (4) asserts the predicate matches every resulting saved-checkpoint key shape. Also exercises both MEAN_STD (mean/std fields) and MIN_MAX (min/max fields) so the discrete-actions buffer shape is covered. A future refactor to the `buffer_` prefix, the `.` → `_` mangling, or the per-mode buffer field names now trips the assertion.

src/opentau/policies/pi07_paligemma/low_level/modeling_pi07_low_level.py
shuheng-liu: Added the guard in a886089. Implementation closely mirrors your suggestion:

- New module-level helper `_find_inf_normalize_buffers(module) -> list[str]` (shares the `_is_normalize_buffer_key` predicate so the strip filter and the guard cannot drift apart).
- In `from_pretrained`: a `strip_ran` local tracks whether the strip branch actually fired inside the inner try block. Immediately after the outer `try/except` (so the `ValueError` is not swallowed by the broad `except Exception` that otherwise just warns and returns the broken model), the helper runs gated on `strip_ran`. If it returns any inf buffer names, we raise `ValueError` with a message that explicitly names the `skip_normalization_weights` flag and the `dataset_stats` requirement, plus up to five buffer names for diagnostics.
- Three CPU tests cover the helper directly: stats-less Normalize is flagged with all paths under `normalize_inputs.`; stats-provided Normalize / Unnormalize returns empty; unrelated `inf` parameters (e.g. a `Linear.bias`) are ignored — the predicate is anchored at the parameter-path start so it only catches the actual buffer paths.

The misleading "use a pretrained model" message from `Normalize.forward` is now strictly upstream-blocked when the user is on this load path with the flag on.

src/opentau/policies/pi07_paligemma/low_level/configuration_pi07_low_level.py
shuheng-liu: Confirming the scope is captured in #338 with both design options you sketched (lift to `PreTrainedPolicy.from_pretrained` vs lift to `Normalize` / `Unnormalize`) and an acceptance checklist. The field docstring on this PR also calls out the current `pi07_paligemma_low_level`-only scope explicitly, so the next person finetuning a `pi05` / `pi05_mem` / `pi06` / `pi07/low_level` / `pi0` checkpoint across a mixture boundary gets a pointer to the same fix rather than rediscovering the trap.

PR #339, opened 2026-05-27
fix(configs+tests): unblock dev_config smoke and merge-collision test failures

shuheng-liu:
## What this does

Two small fixes uncovered while verifying [#335](https://github.com/TensorAuto/OpenTau/issues/335).

### 1. `configs/dev/dev_config.json` — stale `tokenizer_max_length`

Renamed in [#290](https://github.com/TensorAuto/OpenTau/pull/290) (FAST tokenizer support) but `dev_config.json` was not updated, so loading it raised:

```
draccus.utils.DecodingError: `policy`: The fields `tokenizer_max_length` are not valid for PI05Config
```

The issue's verification recipe explicitly suggests `configs/dev/dev_config.json` (with `steps=2`) as one of the smoke targets — that command was unrunnable today because of this stale field.

`configs/examples/pi_config.json` also has `tokenizer_max_length`, but that one targets `pi0` and `PI0Config` still has the field, so it's intentionally left alone.

### 2. `tests/policies/test_pi07_paligemma_low_level.py::TestSkipNormalizationWeights` — `#336` / `#337` API drift

`#336` (per-dataset normalization) renamed `Normalize(stats=...)` to `Normalize(per_dataset_stats=[...])` (a list, with `num_datasets=` as the inf-init path). `#337` was authored against the pre-`#336` API and merged first; `#336` then merged without updating these tests, so CPU CI now fails on every PR with:

```
ValueError: create_stats_buffers requires either `per_dataset_stats` or `num_datasets` ...
TypeError: Normalize.__init__() got an unexpected keyword argument 'stats'
```

Affected tests (all in `TestSkipNormalizationWeights`):
- `test_predicate_matches_real_normalize_buffer_keys`
- `test_find_inf_normalize_buffers_detects_init_inf_sentinels`
- `test_find_inf_normalize_buffers_empty_when_stats_passed`

Updates:
- Inf-init paths now pass `num_datasets=1` (replaces the implicit single-dataset default).
- The populated-stats path passes `per_dataset_stats=[stats]` (wrap the single stat dict in a singleton list to match the new D-row layout).

🐛 Bug

## How it was tested

`pre-commit run --all-files` clean on both files.

**Test fix (CPU-runnable):**

```bash
pytest -xvs tests/policies/test_pi07_paligemma_low_level.py::TestSkipNormalizationWeights -m "not gpu"
# 7 passed in 0.10s
```

**Config fix (end-to-end smoke on an internal single-GPU dev box):**

```bash
opentau-train \
[...]

PR #340, opened 2026-05-27
feat(policies): elevate skip_normalization_weights to all policies

shuheng-liu:
## What this does

Promotes the `skip_normalization_weights` knob (added for `pi07_paligemma_low_level` in [#337](https://github.com/TensorAuto/OpenTau/pull/337)) to `PreTrainedConfig` so **every** policy inherits it, and extracts the strip + inf-buffer guard into shared classmethods on `PreTrainedPolicy` so every `from_pretrained` implementation calls the same code path. The PR docstring on the original PR called this out as a follow-up:

> Scope: only the `pi07_paligemma_low_level` load path honours this knob today; `pi05` / `pi05_mem` / `pi06` / `pi07/low_level` / `pi0` share the same trap and should grow an equivalent knob as follow-up.

This is a follow-up to [#336](https://github.com/TensorAuto/OpenTau/pull/336) (per-dataset normalization + `save_normalization_stats` flag). The two knobs are complementary: `save_normalization_stats=False` keeps stats out of the on-disk safetensors in the first place; `skip_normalization_weights=True` ignores stats that are *already* baked into an existing checkpoint. Use the latter when finetuning an old checkpoint whose saved stats were aggregated over a different dataset mixture than the finetuning data.

### Design

- **Field** moves from `PI07PaligemmaLowLevelConfig` to `PreTrainedConfig` (`src/opentau/configs/policies.py`). All ten policy configs inherit it; `draccus.encode` serialises parent fields, so old `config.json` files (with or without the key under any subclass) load unchanged.
- **Predicate** uses the existing `is_norm_buffer_key` in `src/opentau/policies/pretrained.py` (added by #336). Single source of truth shared with #336's `_save_pretrained` filter, `_promote_legacy_norm_buffers_in_state_dict`, and `_check_norm_stats_loaded`.
- **Orchestration** lives on `PreTrainedPolicy` (`src/opentau/policies/pretrained.py`) as two classmethods:
  - `_strip_normalization_buffers_from_state_dict(state_dict, config, *, is_main_process=True) -> tuple[dict, frozenset[str]]` — flag-gated dict filter that emits the INFO / WARNING log line, mutates `config.skip_normalization_weights = False` (one-shot — fires whether keys were dropped or the no-op warning fired), and returns the stripped key set. When the flag is off, returns `(state_dict, frozenset())` so callers can use the result unconditionally.
  - `_assert_normalize_buffers_initialized(model, *, stripped_keys)` — raises `ValueError` listing inf buffers iff `stripped_keys` is non-empty and `per_dataset_stats` was not wired in. Distinct error path from `_check_norm_stats_loaded`: that one fires on the `save_normalization_stats=False` round-trip when the user forgets `ds_meta=`; this one fires on the `skip_normalization_weights=True` round-trip when the user forgets `per_dataset_stats` in `__init__`. Same symptom (inf buffers), different message pointing at the right knob.
- **Each policy's `from_pretrained`** (or `_load_as_safetensor` for `pi0` / `ValueFunction`) calls the helpers between key remap and `load_state_dict`. For policies that already use #336's `_promote_legacy_norm_buffers_in_state_dict`, the strip runs after promote (order is functionally equivalent — strip removes all norm keys anyway — but matches the natural data flow). The `_assert_*` sits **outside** any broad `try/except Exception` so the `ValueError` is not swallowed — same gotcha #337 hit.
- **`ValueFunction`** previously inherited `PreTrainedPolicy.from_pretrained`, which routes through `safetensors.torch.load_model` (no intermediate dict). It now overrides `_load_as_safetensor` with a thin wrapper that goes through `load_file` + `load_state_dict` so the strip applies.
- **High-level planners** (`PI07HighLevelPlannerPolicy` in both `pi07` and `pi07_paligemma`) use only `normalize_inputs` (no `normalize_targets` / `unnormalize_outputs`). Wired up for API consistency — saved planner checkpoints contain `normalize_inputs.*` keys, so the strip will succeed normally with the INFO log line; the "had no effect" WARNING is reserved for edge cases (e.g. a checkpoint dump from an external source whose `normalize_inputs` keys are missing).

### Refactor of #337's site

`PI07PaligemmaLowLevelPolicy.from_pretrained` and its file-local `_is_normalize_buffer_key` / `_find_inf_normalize_buffers` definitions are deleted and replaced with calls to the shared helpers. The duplicate `skip_normalization_weights` field on `PI07PaligemmaLowLevelConfig` and its docstring block are removed (now inherited).

### Backward compatibility

Default remains `False`. Existing configs (with or without the field) deserialise correctly because the field name and default are unchanged — only the declaration site moved. No `save_pretrained` semantics change.

🗃️ Feature

## How it was tested

**Pre-commit:** clean on every touched file.

**CPU tests** (local, Apple Silicon, `pytest tests/policies/ tests/datasets/test_dataset_mixture.py tests/datasets/test_tagged_dataset.py -m "not gpu" -n auto`):
- 460 passed, 2 skipped, 0 failed in ~89s.
- Includes 30 new tests in `tests/policies/test_pretrained_skip_normalization.py`:
  - 20 parametrised tests (`test_inherited_default_is_false` and `test_can_be_set_true`) across all 10 policy configs — pins the inherited field default and catches accidental shadowing by a subclass. Disambiguated parametrize IDs read `PI07HighLevelConfig` / `PI07PaligemmaHighLevelConfig` rather than the opaque `PI07HighLevelPlannerConfig0` / `…1` suffix that `c.__name__` would produce.
  - 8 integration tests for the shared helpers covering the no-op flag-off path, key-strip + one-shot config reset, no-op warning when no keys match, INFO log on success, `is_main_process` gating, the inf-buffer `ValueError` (with `match=...` on the `per_dataset_stats` message), and the success path with stats wired through. Uses the new `Normalize(features, norm_map, per_dataset_stats=[stats] | num_datasets=1)` API from #336.

**GPU tests** (RTX 5090, `pytest -m "gpu" -n 0` scoped to every touched policy test file + #336's `test_save_pretrained_skip_stats.py` and `test_normalize_per_dataset.py`):
- 15 passed, 10 skipped (pre-existing unconditional skips, unaffected by this branch), 0 failed in ~2m34s.

**Nightly regression tests** will land in the scheduled `regression_test.yml` cron — they haven't run yet on this branch.

```bash
git checkout claude/lucid-jepsen-86854a
pytest tests/policies/test_pretrained_skip_normalization.py -v
```

To enable the knob on a real finetune, set the field under `policy` in any training JSON — works on every policy now:

```json
{
  "policy": {
    "type": "pi05",
    "pretrained_path": "TensorAuto/<your-pretrained-checkpoint>",
    "skip_normalization_weights": true
  }
}
```

…and launch normally with `per_dataset_stats` flowing through `make_policy(..., ds_meta=...)`:

```bash
opentau-train --config_path=configs/your_finetune_config.json
```

The startup log will include `skip_normalization_weights=True; dropped N saved normalize/unnormalize buffer keys` if the filter fires, or a WARNING `the flag had no effect` if the saved checkpoint didn't have normalize buffers in the first place.

## Checklist

### DIFF
--- a/src/opentau/configs/policies.py
+++ b/src/opentau/configs/policies.py
@@ -279,11 +279,16 @@ class PreTrainedConfig(draccus.ChoiceRegistry, HubMixin, abc.ABC):
             supplied to ``__init__``** (e.g. via :py:func:`opentau.policies.factory.make_policy`);
             otherwise the buffers stay at the ``inf`` sentinel from
             :py:func:`opentau.policies.normalize.create_stats_buffers` and the next forward
-            crashes. **One-shot:** after the strip helper runs (whether keys were dropped or
-            the "had no effect" warning fired), the flag is reset to ``False`` on the model's
-            config so that subsequent ``save_pretrained`` / resume / inference loads do not
-            re-strip the now-correct finetuned buffers — a user who opted in once should not
-            have to remember to flip it back before resume. The
+            crashes. **One-shot (in-memory only):** after the strip helper runs (whether keys
+            were dropped or the "had no effect" warning fired), the flag is reset to ``False``
+            on the *in-memory* model's config — a user who opted in once should not have to
+            remember to flip it back before resume. Persistence requires
+            ``save_pretrained``: an in-process resume that goes
+            ``train.py`` → ``save_checkpoint`` → ``cfg.save_pretrained(...)`` writes ``False``
+            to the new checkpoint's ``config.json``, so the next ``from_pretrained`` reads
+            ``False`` and skips the strip. But re-running ``from_pretrained`` on the *original*
+            source checkpoint (e.g. an interactive notebook session that hasn't yet saved) will
+            re-read ``True`` from the source ``config.json`` and re-strip. The
             model was trained against the saved stats, so switching the normalization mid-training
             only makes sense when followed by further training, not when loading purely for
             inference. Honored by every policy whose ``from_pretrained`` (or ``_load_as_safetensor``)
--- a/src/opentau/configs/policies.py
+++ b/src/opentau/configs/policies.py
@@ -279,9 +279,11 @@ class PreTrainedConfig(draccus.ChoiceRegistry, HubMixin, abc.ABC):
             supplied to ``__init__``** (e.g. via :py:func:`opentau.policies.factory.make_policy`);
             otherwise the buffers stay at the ``inf`` sentinel from
             :py:func:`opentau.policies.normalize.create_stats_buffers` and the next forward
-            crashes. **One-shot:** after the strip fires successfully, ``from_pretrained`` resets
-            the flag to ``False`` on the model's config so that subsequent ``save_pretrained`` /
-            resume / inference loads do not re-strip the now-correct finetuned buffers. The
+            crashes. **One-shot:** after the strip helper runs (whether keys were dropped or
+            the "had no effect" warning fired), the flag is reset to ``False`` on the model's
+            config so that subsequent ``save_pretrained`` / resume / inference loads do not
+            re-strip the now-correct finetuned buffers — a user who opted in once should not
+            have to remember to flip it back before resume. The
             model was trained against the saved stats, so switching the normalization mid-training
             only makes sense when followed by further training, not when loading purely for
             inference. Honored by every policy whose ``from_pretrained`` (or ``_load_as_safetensor``)
--- a/src/opentau/configs/policies.py
+++ b/src/opentau/configs/policies.py
@@ -269,6 +269,24 @@ class PreTrainedConfig(draccus.ChoiceRegistry, HubMixin, abc.ABC):
             normalization mode to apply.
         output_normalization_modes: Similar dictionary as `input_normalization_modes`, but to unnormalize to
             the original scale.
+        skip_normalization_weights: When loading via :py:meth:`~opentau.policies.pretrained.PreTrainedPolicy.from_pretrained`,
+            drop the saved ``normalize_*`` / ``unnormalize_*`` buffer tensors from the state dict
+            before ``load_state_dict``. The buffers freshly initialised from ``dataset_stats``
+            then survive — use this when finetuning a checkpoint whose saved normalization stats
+            were aggregated over a different dataset mixture than the finetuning data. The
+            buffers are registered as ``nn.Parameter(requires_grad=False)``, so training alone
+            cannot recover from inheriting the wrong stats. **Requires ``dataset_stats`` to be
+            supplied to ``__init__``** (e.g. via :py:func:`opentau.policies.factory.make_policy`);
+            otherwise the buffers stay at the ``inf`` sentinel from
+            :py:func:`opentau.policies.normalize.create_stats_buffers` and the next forward
+            crashes. **One-shot:** after the strip fires successfully, ``from_pretrained`` resets
+            the flag to ``False`` on the model's config so that subsequent ``save_pretrained`` /
+            resume / inference loads do not re-strip the now-correct finetuned buffers. The
+            model was trained against the saved stats, so switching the normalization mid-training
+            only makes sense when followed by further training, not when loading purely for
+            inference. Honored by every policy whose ``from_pretrained`` (or ``_load_as_safetensor``)
+            calls :py:meth:`~opentau.policies.pretrained.PreTrainedPolicy._strip_normalization_buffers_from_state_dict`.
+            Defaults to ``False`` (no behaviour change).
     """
 
     n_obs_steps: int = 1
@@ -282,6 +300,7 @@ class PreTrainedConfig(draccus.ChoiceRegistry, HubMixin, abc.ABC):
     # automatic gradient scaling is used.
     use_amp: bool = False
     pretrained_path: str | None = None
+    skip_normalization_weights: bool = False
 
     # When False, `_save_pretrained` strips normalize_*.buffer_* / unnormalize_*.buffer_*
     # keys from the state_dict before writing model.safetensors. Reloading then requires
--- a/src/opentau/policies/pi0/modeling_pi0.py
+++ b/src/opentau/policies/pi0/modeling_pi0.py
@@ -328,11 +328,25 @@ class PI0Policy(PreTrainedPolicy):
         # load via `model.load_state_dict(..., strict=True)`.
         model._promote_legacy_norm_buffers_in_state_dict(transformed_state_dict)
 
+        # Strip saved normalize/unnormalize buffers when the user opted in
+        # via config.skip_normalization_weights — see PreTrainedConfig and
+        # PreTrainedPolicy._strip_normalization_buffers_from_state_dict.
+        transformed_state_dict, stripped_keys = cls._strip_normalization_buffers_from_state_dict(
+            transformed_state_dict, model.config
+        )
+
         # Load the transformed state dict
         msg = model.load_state_dict(transformed_state_dict, strict=strict)
 
-        # Log message
-        log_model_loading_keys(msg.missing_keys, msg.unexpected_keys)
+        # Hide deliberately-stripped buffer keys from the missing-keys log so
+        # it does not contradict the INFO emitted by the strip helper just
+        # above. ``stripped_keys`` is empty when the flag is off (no-op).
+        unintended_missing = [k for k in msg.missing_keys if k not in stripped_keys]
+        log_model_loading_keys(unintended_missing, msg.unexpected_keys)
+
+        # When the strip ran, fail loudly if dataset_stats was not wired in.
+        # No-op for the default load path.
+        cls._assert_normalize_buffers_initialized(model, stripped_keys=stripped_keys)
         return model
 
     def get_optim_params(self) -> dict:
--- a/src/opentau/policies/pi0/modeling_pi0.py
+++ b/src/opentau/policies/pi0/modeling_pi0.py
@@ -38,6 +38,7 @@ from opentau.policies.pi0.paligemma_with_expert import (
 )
 from opentau.policies.pretrained import PreTrainedPolicy
 from opentau.policies.utils import log_model_loading_keys
+from opentau.utils.accelerate_utils import get_proc_accelerator
 from opentau.utils.utils import get_safe_dtype
 
 
@@ -331,12 +332,21 @@ class PI0Policy(PreTrainedPolicy):
         # Strip saved normalize/unnormalize buffers when the user opted in
         # via config.skip_normalization_weights — see PreTrainedConfig and
         # PreTrainedPolicy._strip_normalization_buffers_from_state_dict.
+        # Thread is_main_process so the helper's INFO/WARNING fires once per
+        # load, not once per rank under DDP/FSDP.
+        acc = get_proc_accelerator()
+        is_main_process = acc.is_main_process if acc else True
         transformed_state_dict, stripped_keys = cls._strip_normalization_buffers_from_state_dict(
-            transformed_state_dict, model.config
+            transformed_state_dict, model.config, is_main_process=is_main_process
         )
 
-        # Load the transformed state dict
-        msg = model.load_state_dict(transformed_state_dict, strict=strict)
+        # When the strip removed keys, force strict=False on this load —
+        # otherwise the deliberately-dropped Normalize buffer keys would
+        # trigger RuntimeError("Missing key(s) ...") and mask the
+        # `skip_normalization_weights=True` semantics. Preserves strict=True
+        # for the default-load path (stripped_keys empty).
+        effective_strict = strict and not stripped_keys
+        msg = model.load_state_dict(transformed_state_dict, strict=effective_strict)
 
         # Hide deliberately-stripped buffer keys from the missing-keys log so
         # it does not contradict the INFO emitted by the strip helper just
--- a/src/opentau/policies/pi05/modeling_pi05.py
+++ b/src/opentau/policies/pi05/modeling_pi05.py
@@ -370,6 +370,10 @@ class PI05Policy(PreTrainedPolicy):
         # Now manually load and remap the state dict
         acc = get_proc_accelerator()
         is_main_process = acc.is_main_process if acc else True
+        # Populated inside the try block when skip_normalization_weights fires;
+        # used outside the try/except to gate the inf-buffer guard so the
+        # ValueError is not swallowed by the broad except below.
+        stripped_keys: frozenset[str] = frozenset()
         try:
             # Try to load the pytorch_model.bin or model.safetensors file
             if is_main_process:
@@ -420,6 +424,13 @@ class PI05Policy(PreTrainedPolicy):
             if remap_count > 0 and is_main_process:
                 print(f"Remapped {remap_count} state dict keys")
 
+            # Strip saved normalize/unnormalize buffers when the user opted in
+            # via config.skip_normalization_weights — see PreTrainedConfig and
+            # PreTrainedPolicy._strip_normalization_buffers_from_state_dict.
+            remapped_state_dict, stripped_keys = cls._strip_normalization_buffers_from_state_dict(
+                remapped_state_dict, model.config, is_main_process=is_main_process
+            )
+
             # Load the remapped state dict into the model
             # Promote legacy single-dataset Normalize/Unnormalize buffers from
             # `(*feat_shape,)` to the new `(1, *feat_shape)` stacked layout so pre-PR
@@ -427,15 +438,21 @@ class PI05Policy(PreTrainedPolicy):
             model._promote_legacy_norm_buffers_in_state_dict(remapped_state_dict)
             missing_keys, unexpected_keys = model.load_state_dict(remapped_state_dict, strict=False)
 
-            if missing_keys and is_main_process:
-                print(f"Missing keys when loading state dict: {len(missing_keys)} keys")
-                if len(missing_keys) <= 20:
-                    for key in missing_keys:
+            # Hide deliberately-stripped buffer keys from the missing-keys
+            # warning so the noisy log does not directly contradict the INFO
+            # logged just above. ``stripped_keys`` is empty when the flag is
+            # off, so this is a no-op for default loads.
+            unintended_missing = [key for key in missing_keys if key not in stripped_keys]
+
+            if unintended_missing and is_main_process:
+                print(f"Missing keys when loading state dict: {len(unintended_missing)} keys")
+                if len(unintended_missing) <= 20:
+                    for key in unintended_missing:
                         print(f"  - {key}")
                 else:
-                    for key in missing_keys[:20]:
+                    for key in unintended_missing[:20]:
                         print(f"  - {key}")
-                    print(f"  ... and {len(missing_keys) - 20} more")
+                    print(f"  ... and {len(unintended_missing) - 20} more")
 
             if unexpected_keys and is_main_process:
                 print(f"Unexpected keys when loading state dict: {len(unexpected_keys)} keys")
@@ -447,12 +464,17 @@ class PI05Policy(PreTrainedPolicy):
                         print(f"  - {key}")
                     print(f"  ... and {len(unexpected_keys) - 20} more")
 
-            if not missing_keys and not unexpected_keys and is_main_process:
+            if not unintended_missing and not unexpected_keys and is_main_process:
                 print("All keys loaded successfully!")
 
         except Exception as e:
             if is_main_process:
                 print(f"Warning: Could not remap state dict keys: {e}")
+
+        # Outside the try/except so the ValueError is not swallowed by the
+        # broad except above. The helper no-ops when ``stripped_keys`` is
+        # empty (flag was off or the try block bailed before the strip ran).
+        cls._assert_normalize_buffers_initialized(model, stripped_keys=stripped_keys)
 
         return model
 
--- a/src/opentau/policies/pi05_mem/modeling_pi05.py
+++ b/src/opentau/policies/pi05_mem/modeling_pi05.py
@@ -314,6 +314,10 @@ class PI05MemPolicy(PreTrainedPolicy):
 
         acc = get_proc_accelerator()
         is_main_process = acc.is_main_process if acc else True
+        # Populated inside the try block when skip_normalization_weights fires;
+        # used outside the try/except to gate the inf-buffer guard so the
+        # ValueError is not swallowed by the broad except below.
+        stripped_keys: frozenset[str] = frozenset()
         try:
             if is_main_process:
                 logging.info("Loading model from: %s", pretrained_name_or_path)
@@ -367,14 +371,27 @@ class PI05MemPolicy(PreTrainedPolicy):
             # (outside the `if remap_count > 0` block) — promotion is needed
             # whether or not any other keys were renamed.
             model._promote_legacy_norm_buffers_in_state_dict(remapped_state_dict)
+
+            # Strip saved normalize/unnormalize buffers when the user opted in
+            # via config.skip_normalization_weights — see PreTrainedConfig and
+            # PreTrainedPolicy._strip_normalization_buffers_from_state_dict.
+            remapped_state_dict, stripped_keys = cls._strip_normalization_buffers_from_state_dict(
+                remapped_state_dict, model.config, is_main_process=is_main_process
+            )
             missing_keys, unexpected_keys = model.load_state_dict(remapped_state_dict, strict=False)
 
-            if missing_keys and is_main_process:
-                logging.warning("Missing keys when loading state dict: %d keys", len(missing_keys))
-                for key in missing_keys[:20]:
+            # Hide deliberately-stripped buffer keys from the missing-keys
+            # warning so the noisy WARNING does not directly contradict the
+            # INFO logged just above. ``stripped_keys`` is empty when the
+            # flag is off, so this is a no-op for default loads.
+            unintended_missing = [key for key in missing_keys if key not in stripped_keys]
+
+            if unintended_missing and is_main_process:
+                logging.warning("Missing keys when loading state dict: %d keys", len(unintended_missing))
+                for key in unintended_missing[:20]:
                     logging.warning("  - %s", key)
-                if len(missing_keys) > 20:
-                    logging.warning("  ... and %d more", len(missing_keys) - 20)
+                if len(unintended_missing) > 20:
+                    logging.warning("  ... and %d more", len(unintended_missing) - 20)
 
             if unexpected_keys and is_main_process:
                 logging.warning("Unexpected keys when loading state dict: %d keys", len(unexpected_keys))
@@ -383,12 +400,17 @@ class PI05MemPolicy(PreTrainedPolicy):
                 if len(unexpected_keys) > 20:
                     logging.warning("  ... and %d more", len(unexpected_keys) - 20)
 
-            if not missing_keys and not unexpected_keys and is_main_process:
+            if not unintended_missing and not unexpected_keys and is_main_process:
                 logging.info("All keys loaded successfully!")
 
         except Exception as e:
             if is_main_process:
                 logging.warning("Could not remap state dict keys: %s", e)
+
+        # Outside the try/except so the ValueError is not swallowed by the
+        # broad except above. The helper no-ops when ``stripped_keys`` is
+        # empty (flag was off or the try block bailed before the strip ran).
+        cls._assert_normalize_buffers_initialized(model, stripped_keys=stripped_keys)
 
         return model
 
--- a/src/opentau/policies/pi06/modeling_pi06.py
+++ b/src/opentau/policies/pi06/modeling_pi06.py
@@ -370,6 +370,10 @@ class PI06Policy(PreTrainedPolicy):
 
         acc = get_proc_accelerator()
         is_main_process = acc.is_main_process if acc else True
+        # Populated inside the try block when skip_normalization_weights fires;
+        # used outside the try/except to gate the inf-buffer guard so the
+        # ValueError is not swallowed by the broad except below.
+        stripped_keys: frozenset[str] = frozenset()
         try:
             if is_main_process:
                 print(f"Loading model from: {pretrained_name_or_path}")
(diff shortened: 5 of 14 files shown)


shuheng-liu (commented):
## Review: elevate `skip_normalization_weights` to all policies

Solid follow-up to #337. The refactor lands the field where it belongs (`PreTrainedConfig`), factors the strip + inf-guard into shared helpers on `PreTrainedPolicy`, and threads them uniformly through all ten `from_pretrained` / `_load_as_safetensor` paths. All 33 new tests pass locally (`tests/policies/test_normalize_helpers.py` + `tests/policies/test_pretrained_skip_normalization.py`).

### What I like

- **Single source of truth** for the buffer-key predicate (`is_normalize_buffer_key`) and the inf-detector (`find_inf_normalize_buffers`) lives in `normalize.py` — same module that defines the buffer-naming convention. Future renames to `Normalize.__init__`'s `buffer_` prefix or `.` → `_` mangling will trip `test_matches_real_normalize_buffer_keys` before silently breaking the strip in nine consumers.
- **Inf-guard is correctly placed outside the broad `try/except`** in every per-policy override (pi05 line 452, pi05_mem line 386, pi06 line 437, pi07/low_level line 480, pi07/high_level_planner line 374, pi07_paligemma/{low_level,high_level_planner}, pi0 line 319, value line 195). The same gotcha #337 hit — the `ValueError` would otherwise be swallowed by the `except Exception` and replaced with a warning + a model that crashes on its first batch.
- **One-shot reset semantics** are tested explicitly (`test_strips_only_normalize_buffer_keys_and_resets_flag`), and the strip uses `frozenset` as the cross-call gate which doubles as both the "did this fire?" flag and the missing-keys filter.
- **`unintended_missing = [k for k in missing_keys if k not in stripped_keys]`** cleanly silences the noisy "missing keys" warning for keys we just deliberately dropped — would otherwise directly contradict the INFO logged a few lines above.
- **`ValueFunction._load_as_safetensor` override** is necessary and correctly explained: the base routes through `safetensors.torch.load_model` which writes tensors into the model in place, leaving no intermediate dict for the strip to operate on.
- **Backward compatibility**: field name + default unchanged, only declaration site moved. `draccus.encode` serialises parent fields, so old `config.json` files load unchanged.

### Findings

**1. Docstring slightly overstates "successful" reset** (`src/opentau/configs/policies.py:282-284`)

> **One-shot:** after the strip fires successfully, `from_pretrained` resets the flag to `False` on the model's config…

The helper resets the flag *regardless* of whether keys actually matched — see `test_warning_when_flag_set_but_no_keys_match` line 196: `assert config.skip_normalization_weights is False  # still consumed`. So the flag is consumed even when the strip is a no-op (warning path). The current behaviour is the right call (a user who opted in once shouldn't have to remember to flip it back before resume), but the docstring would be more accurate as something like *"after the strip helper runs (success or no-op warning), the flag is reset to `False`"*. Same nit applies to the helper's own docstring on `pretrained.py:218-220`.

**2. PR description claim about high-level planner warning path is off**

The body says:

> High-level planners (`PI07HighLevelPlannerPolicy` in both `pi07` and `pi07_paligemma`) use only `normalize_inputs` (no `normalize_targets` / `unnormalize_outputs`). Still wired up for API consistency — the existing 'no normalize_ keys present in the saved state dict; the flag had no effect' warning is exactly the right signal when a planner checkpoint lacks the buffers.

But planner checkpoints saved by this codebase **do** have `normalize_inputs.*` keys (since `self.normalize_inputs = Normalize(...)` is registered in `__init__` and survives `save_pretrained`). The "had no effect" warning only fires when *every* normalize/unnormalize key is missing — that's an edge case (e.g. checkpoint dump from an external source). For normal saved planner checkpoints, the strip will succeed and the WARNING won't fire. Doesn't change behaviour, just clarifying what users should expect to see in the log.

**3. `@staticmethod` mutating a parameter** (`pretrained.py:195-265`)

`_strip_normalization_buffers_from_state_dict` is decorated `@staticmethod` but mutates `config.skip_normalization_weights = False` as a side effect. It's documented and tested, but the mutation-through-side-effect on a staticmethod is unusual; a `@classmethod` would carry the same semantics and signal more accurately that this is part of the policy's loading protocol. Optional — keep as-is if you prefer the strict "no cls/self for pure-helper" rule.

**4. Strip can mask the original load error in the very-unlikely-but-possible nested-failure case**

In the per-policy `from_pretrained` overrides (pi05/pi05_mem/pi06/pi07/pi07_paligemma):

```python
stripped_keys: frozenset[str] = frozenset()
try:
    ...
    remapped_state_dict, stripped_keys = cls._strip_normalization_buffers_from_state_dict(...)
    missing_keys, unexpected_keys = model.load_state_dict(remapped_state_dict, strict=False)
    ...
except Exception as e:
    logging.warning("Could not remap state dict keys: %s", e)
cls._assert_normalize_buffers_initialized(model, stripped_keys=stripped_keys)
```

If `load_state_dict` raises *after* the strip ran (very unlikely with `strict=False`, but e.g. a CUDA OOM on the in-place copy), `stripped_keys` is populated and the inf-guard fires with a "`skip_normalization_weights=True` requires `dataset_stats`" message — masking the original load error. Pre-existing pattern; not worth blocking on, but worth flagging because the helper docstring (lines 280-282) explicitly states the guard *must* run outside the try/except so the `ValueError` is not swallowed — that's correct in the missing-`dataset_stats` case but creates this small misleading-error window for the unrelated-failure case.

**5. Mergeable state is "dirty"** — needs a rebase against `main` before merge.

**6. (FYI, not actionable)** Local `pytest tests/policies/test_pi07_paligemma_low_level.py -m "not gpu"` shows 24 failures in this sandboxed env, all from `OSError: You are trying to access a gated repo` against `google/paligemma-3b-pt-224`. Not caused by this PR — the sandbox lacks HF auth. The 33 new tests in `test_normalize_helpers.py` + `test_pretrained_skip_normalization.py` all pass here (matching the PR-body claim).

### Minor

- `test_pretrained_skip_normalization.py:64-66`: `PI07HighLevelPlannerConfig` is imported twice (once from `pi07.high_level_planner`, once from `pi07_paligemma.high_level_planner`) and both end up in `ALL_POLICY_CONFIGS`. pytest shows the parametrize IDs as `PI07HighLevelPlannerConfig0` / `PI07HighLevelPlannerConfig1`, which is functional but a bit opaque. Aliasing the imports (`as PI07HighLevelConfig` / `as PI07PaligemmaHighLevelConfig`, which you already do) is half the way — applying those alias names in `ALL_POLICY_CONFIGS` would let the IDs read `PI07HighLevelConfig` / `PI07PaligemmaHighLevelConfig` and remove the suffix ambiguity. PR body already calls this out as "×2", so this is purely cosmetic.

Overall — happy with this once the docstring nits are addressed (or acknowledged) and the branch is rebased.

---

shuheng-liu:
Thanks for the thorough review — addressed in `1ded440` and rebased onto current main (`e4c782b`, includes #336 and #339). Force-pushed.

**Code changes:**

- **#1 (one-shot reset docstring)**: Updated both [policies.py:282-285](src/opentau/configs/policies.py) and [pretrained.py:218-221](src/opentau/policies/pretrained.py) (the helper docstring) to call out that the reset fires whether keys were dropped or the no-op warning fired — *not* only on a "successful" strip. Added the user-rationale: a user who opted in once shouldn't have to remember to flip it back before resume.
- **#3 (@staticmethod → @classmethod)**: Converted both `_strip_normalization_buffers_from_state_dict` and `_assert_normalize_buffers_initialized` to `@classmethod`. Signals these are part of the policy loading protocol (matches the existing `_load_as_safetensor` style). Call sites unchanged — `cls.method(...)` works for both.
- **Minor (parametrize IDs)**: `_POLICY_CONFIG_CASES` in [test_pretrained_skip_normalization.py:63-76](tests/policies/test_pretrained_skip_normalization.py) now pairs each class with an explicit display name. IDs render as `PI07HighLevelConfig` / `PI07PaligemmaHighLevelConfig` instead of `PI07HighLevelPlannerConfig0` / `PI07HighLevelPlannerConfig1`.

**Rebase notes** (the diff is larger than the review-fix changes alone because the rebase brought in #336):

- #336's `is_norm_buffer_key` predicate in [pretrained.py:67](src/opentau/policies/pretrained.py) is now the single source of truth — my `is_normalize_buffer_key` and `find_inf_normalize_buffers` helpers from `normalize.py` are gone. The strip helper uses `is_norm_buffer_key`; the inf-guard inlines the same `NORM_MODULE_NAMES` walk that `_check_norm_stats_loaded` does (kept it inline rather than extracting because the message differs — see below).
- The inf-guard message now points the user at `per_dataset_stats` (the new constructor kwarg from #336), not `dataset_stats`. The docstring explicitly contrasts the two failure paths: `_check_norm_stats_loaded` fires on `save_normalization_stats=False` + missing `ds_meta`; my guard fires on `skip_normalization_weights=True` + missing `per_dataset_stats`. Same symptom (inf buffers), different actionable fix.
- Per-policy wiring: each modeling file now runs `_promote_legacy_norm_buffers_in_state_dict` *first* (handles the new stacked-buffer shape on legacy checkpoints) then `_strip_normalization_buffers_from_state_dict` second. Order doesn't matter functionally (strip removes all norm keys) but matches the natural data flow.
- `tests/policies/test_normalize_helpers.py` was deleted as redundant — the `is_norm_buffer_key` predicate is already pinned by `tests/policies/test_save_pretrained_skip_stats.py::test_is_norm_buffer_key_matches_known_prefixes` (added in #336).

**On the deferred findings:**

- **#2 (planner warning path in PR body)** — Good catch, you're right. Planner checkpoints saved by this codebase do contain `normalize_inputs.*` keys (the policy registers `self.normalize_inputs = Normalize(...)` in `__init__` and `save_pretrained` keeps it). So the strip will succeed normally; the "had no effect" warning only fires for edge cases (e.g. a checkpoint dump from an external source without the buffers). I'll fix the PR-body wording in the next edit.
- **#4 (inf-guard masking unrelated load errors)** — I traced this carefully and don't think it's actually a problem. The inf-guard only fires when (a) `stripped_keys` is non-empty (strip ran) AND (b) some Normalize buffer is still inf. Buffers are inf iff `per_dataset_stats` was *not* passed to `__init__`. So:
  - Scenario A: User passes `per_dataset_stats`, `load_state_dict` OOMs midway → buffers are finite (from `__init__`), inf-guard no-ops, original OOM surfaces in the warning. ✓
  - Scenario B: User does *not* pass `per_dataset_stats`, strip is on, `load_state_dict` OOMs midway → buffers at inf, inf-guard fires with "needs `per_dataset_stats`". The message is correct because that *is* the actionable root cause; the OOM is just the proximate trigger that revealed the latent misconfig.

  In other words: the inf-guard fires exactly when `per_dataset_stats` was missing, and the user's actual problem is that missing arg, not whatever caused the load to error. Open to revisiting if there's a scenario I missed.

- **#5 (rebase)** — done.
- **#6 (sandbox HF auth)** — confirmed not caused by this PR.

**Verification post-rebase:**
- Pre-commit clean on every touched file.
- CPU tests: full `tests/policies/` + new `tests/datasets/test_dataset_mixture.py` and `tests/datasets/test_tagged_dataset.py` from #336 = **460 passed, 2 skipped** in ~89s.
- GPU tests (`pytest -m "gpu" -n 0` on the touched policy test files + #336's `test_save_pretrained_skip_stats.py` and `test_normalize_per_dataset.py`): **15 passed, 10 skipped (pre-existing)** in ~2m34s.

shuheng-liu (commented):
## Second-pass review — additional findings

A few things I missed on the first pass.

### 1. Silent inconsistency: `strict=True` + `skip_normalization_weights=True` breaks in pi0 / value but not the others

The base `PreTrainedPolicy.from_pretrained` exposes `strict: bool = False` (`pretrained.py:126`) and threads it through to `_load_as_safetensor`. The two policies that use that path — `pi0` and `value` — both pass it through verbatim:

```python
# pi0/modeling_pi0.py:309
msg = model.load_state_dict(transformed_state_dict, strict=strict)
# value/modeling_value.py:185
missing_keys, unexpected_keys = model.load_state_dict(state_dict, strict=strict)
```

When `skip_normalization_weights=True`, the strip removes `normalize_*` / `unnormalize_*` keys from the dict — but the model still has those buffers registered (they were initialised from `dataset_stats` in `__init__`). With `strict=True`, `load_state_dict` raises `RuntimeError("Missing key(s) ...")` on the stripped keys.

The other eight policies hardcode `strict=False` on the actual `load_state_dict` call (e.g. `pi05/modeling_pi05.py:407` — `model.load_state_dict(remapped_state_dict, strict=False)` — the override's `strict` parameter is effectively unused there), so they don't have this problem.

Net effect: a user who explicitly opts into both `strict=True` *and* `skip_normalization_weights=True` gets a `RuntimeError` on pi0 / value, and silent success on the other eight — same flag combination, divergent outcomes. Fix is one line in each `_load_as_safetensor` — pass `strict and not stripped_keys`, or unconditionally `strict=False` when the flag fired:

```python
effective_strict = strict and not stripped_keys
msg = model.load_state_dict(transformed_state_dict, strict=effective_strict)
```

### 2. `pi0` / `value` don't gate the strip-helper logging with `is_main_process`

The other eight policies pass `is_main_process=is_main_process` to `_strip_normalization_buffers_from_state_dict` so the INFO / WARNING fires once per load. `pi0._load_as_safetensor` (`pi0/modeling_pi0.py:304`) and `value._load_as_safetensor` (`value/modeling_value.py:183`) both call the helper without `is_main_process`, defaulting to `True` — so in distributed training, those two policies log the strip line once per rank (e.g. eight times under DDP-8). Minor noise issue, not a correctness bug, but inconsistent with the rest. The two helpers don't have `Accelerator` / `is_main_process` in scope since they're called from the base `from_pretrained` — easiest fix is to thread it through, or expose a module-level "am I main?" check inside the helper that the base path can use too.

### 3. `_assert_normalize_buffers_initialized` semantics are slightly looser than the name suggests

The guard checks for **any** inf normalize buffer on the model when `stripped_keys` is non-empty — it doesn't verify the inf buffer corresponds to a stripped key. In practice this is the same set (every `Normalize` buffer is either stripped or not, and `create_stats_buffers` raises `KeyError` if `dataset_stats` is partial — so partial-stats can't reach this guard), so I don't think there's a reachable false-positive today. Worth a one-line comment in the guard's docstring noting why the broader check is safe (i.e. "`create_stats_buffers` rejects partial stats at `__init__`, so any inf buffer here means `dataset_stats=None`"). Optional.

### 4. Pre-existing dead `strict` param in `pi07_paligemma/low_level/from_pretrained`

```python
# pi07_paligemma/low_level/modeling_pi07_low_level.py:356
strict: bool = True,
```

That parameter is accepted but never read inside the override — the actual `load_state_dict` call hardcodes `strict=False` (line 467). Not introduced by this PR (came in with #337), but since the PR is already touching that function and the same dead-param exists in seven other overrides, a cleanup pass would be cheap. Or leave it — surgical-changes guideline cuts both ways. Flagging for awareness.

### 5. Test fixture coverage gap (minor)

`_make_fake_module_with_normalize_buffers` (`test_pretrained_skip_normalization.py:91`) builds a parent with only `normalize_inputs` + `unnormalize_outputs`. Production policies also carry `normalize_targets` (every flow-matching policy) and `normalize_discrete_actions` (pi05 / pi07 / pi07_paligemma low-level). The shared helper iterates the dict so this works in practice, but adding the other two submodule names to the fixture would prove the strip handles all four attribute prefixes in one place. Not blocking.

---

### Re-confirming first-pass items

The earlier review's findings still stand. Of those, I'd say the **strict-flag inconsistency above (#1)** is the highest-priority new item — it's a silent behavioural divergence between code paths that should behave identically, and the fix is mechanical. Everything else here is polish.

---

shuheng-liu:
Thanks for the second pass — addressed in `4039691`.

**#1 (strict-flag inconsistency)** — fixed. Both [pi0._load_as_safetensor](src/opentau/policies/pi0/modeling_pi0.py) and [value._load_as_safetensor](src/opentau/policies/value/modeling_value.py) now compute `effective_strict = strict and not stripped_keys` and pass that to `load_state_dict`. The strict semantics apply unchanged on the default-load path; only relaxes when the strip actually fired. Matches the implicit `strict=False` the other eight policies hardcode.

**#2 (is_main_process not threaded)** — fixed. Both `pi0` and `value` `_load_as_safetensor` now resolve `is_main_process` via `get_proc_accelerator()` (matching the pattern in the other eight overrides) and thread it to `_strip_normalization_buffers_from_state_dict`. INFO/WARNING fires once per load, not once per rank.

**#3 (broader-than-needed inf check semantics)** — added a one-paragraph note in [`_assert_normalize_buffers_initialized`](src/opentau/policies/pretrained.py) explaining why the "any normalize buffer is inf" check is a safe proxy for "the stripped buffer is still inf": `create_stats_buffers` rejects partial stats at `__init__`, so buffers are either all-finite (`per_dataset_stats` supplied) or all-inf (`per_dataset_stats=None`) — no reachable mixed state.

**#4 (dead `strict` param in pi07_paligemma low-level)** — leaving as-is per surgical-changes guideline. Eight overrides share the same dead param; cleaning that up would balloon the diff for no behaviour change, and the param is at least syntactically forwarded via `**kwargs` patterns elsewhere. Noted for a follow-up.

**#5 (test fixture only covers two of four submodules)** — left as-is. The shared helper's predicate (`is_norm_buffer_key`) iterates over `NORM_BUFFER_PREFIXES` which covers all four; the fixture's two submodules are enough to exercise the loop, the missing-keys filter, and the inf-guard. The four-prefix coverage is already pinned in `test_save_pretrained_skip_stats.py::test_is_norm_buffer_key_matches_known_prefixes` (#336). Happy to extend if you'd rather have it in one place.

**Also fixed**: the previous push triggered a `ruff-format` failure in CI on `--all-files` (caught two stray formatting issues outside the touched-file set I was checking locally). Running `pre-commit run --all-files` now is clean.

**Verification**:
- Pre-commit clean on `pre-commit run --all-files` (CI parity).
- CPU suite (`pytest tests/policies/ tests/datasets/test_dataset_mixture.py tests/datasets/test_tagged_dataset.py -m "not gpu" -n auto`): **460 passed, 2 skipped**.
- Re-running the GPU subset on the RTX 5090 dev box now; will update once done.

shuheng-liu (commented):
## Third-pass review — a few more

Mostly minor; one is worth a defensive call.

### 1. `find_inf_normalize_buffers` doesn't materialize DTensors

`Normalize.forward` carefully routes every buffer through `_materialize` (`normalize.py:35-58`) before calling `torch.isinf(...).any()`, precisely to handle FSDP2 (`fully_shard`) where the parameter is a `DTensor`. The new `find_inf_normalize_buffers` (`normalize.py:218-222`) skips that step:

```python
return [
    name
    for name, param in module.named_parameters()
    if is_normalize_buffer_key(name) and torch.isinf(param).any()
]
```

In practice this won't bite today because `from_pretrained` runs *before* `accelerator.prepare` (and therefore before any `fully_shard` wrapping) — the policy is still on CPU with plain Tensors at strip-check time. But the helper is exposed as a public utility on `opentau.policies.normalize`, and a future caller who invokes it on a post-prepare model would hit the same "`aten.isinf got mixed torch.Tensor and DTensor`" failure mode the `_materialize` docstring documents. Defensive `_materialize(param)` inside the comprehension makes the helper safe for either lifecycle stage and is a one-line change.

### 2. Reset is purely in-memory — no automatic write-back

The docstring on `skip_normalization_weights` (`configs/policies.py:282-285`) says:

> after the strip fires successfully, `from_pretrained` resets the flag to `False` on the model's config so that subsequent `save_pretrained` / resume / inference loads do not re-strip the now-correct finetuned buffers.

That holds for **in-process** resumes: a single training run that hits the strip once, then later checkpoints via `save_checkpoint → cfg.save_pretrained(checkpoint_dir)` (`utils/train_utils.py:128`) does persist `false`, and resuming from that checkpoint sees `false`. Good.

But if a user calls `policy = SomePolicy.from_pretrained("…/source_ckpt", config=cfg_with_flag_set)` interactively (e.g. exploring in a notebook) and then re-runs `from_pretrained` on the same source path before saving, the second call re-strips — because the source `config.json` still has `true`, and `PreTrainedConfig.from_pretrained` reads it fresh from disk. Not a bug per se (the docstring's wording is technically defensible — "model's config" is the in-memory instance), but easy to misread as "the flag is persistently consumed." A half-sentence in the docstring clarifying *"reset is in-memory only; `save_pretrained` is required to make the consumption durable"* would head off that confusion.

### 3. Naming nit: `find_inf_normalize_buffers` iterates `named_parameters()`

`Normalize` registers stats as `nn.Parameter(..., requires_grad=False)`, not via `register_buffer`, so the helper correctly uses `named_parameters()`. But the function name reads "buffers" and the docstring talks about "buffer parameter[s]" — internally consistent with the rest of the codebase (e.g. `is_normalize_buffer_key`, `create_stats_buffers`), but a new reader walking in cold has to do a double-take to verify the helper isn't broken. One-line note in the docstring — "*despite the name, these are `nn.Parameter`s registered with `requires_grad=False`; we iterate `named_parameters()`, not `named_buffers()`*" — saves that read.

### 4. Logging vs `print` inconsistency widens

`_strip_normalization_buffers_from_state_dict` uses `logging.info` / `logging.warning`. Most policies log around the strip site with the same `logging` machinery (pi05_mem, pi06, pi07/{low,high}_level, pi07_paligemma/{low,high}_level), but `pi05/modeling_pi05.py` continues to use `print()` in the surrounding code. After this PR, the same flag's effect on pi05 will be split between an INFO line (visible only if logging is configured at INFO+) and bare `print`s from neighbouring code (always visible) — making the strip message intermittent depending on the user's logging setup. Pre-existing inconsistency this PR didn't create, but it now has user-facing impact for the new feature. Lowest priority; flagging because the surface area just grew.

### 5. Pre-commit clean

For the record — `pre-commit run --files src/opentau/policies/normalize.py src/opentau/policies/pretrained.py src/opentau/configs/policies.py tests/policies/test_normalize_helpers.py tests/policies/test_pretrained_skip_normalization.py` is fully green on the touched files (ruff lint + format, pyupgrade, gitleaks, bandit, typos all pass).

---

Across all three passes the highest-priority items remain (in order):

1. Second-pass #1 — `strict=True` + `skip_normalization_weights=True` divergence in pi0 / value.
2. Second-pass #2 — `is_main_process` not threaded for pi0 / value (log-duplication under DDP).
3. Third-pass #1 — defensive `_materialize` in `find_inf_normalize_buffers` so the helper stays correct post-`fully_shard`.

Everything else is polish.

---

shuheng-liu:
**GPU re-verification update** — `pytest -m "gpu" -n 0` re-run on a separate GPU dev box (3090, internal infra; primary GPU box was offline). Split into two runs because `pi05_mem_gpu` needed a fresh model-cache warm-up on this host:

- `test_pi05_mem_gpu.py`: 6 passed in 16:29 (first-time model downloads dominated the wall-clock).
- Remaining: `test_pi0.py`, `test_pi05.py`, `test_pi06.py`, `test_pi07_low_level.py`, `test_pi07_high_level_planner.py`, `test_pi07_paligemma_low_level.py`, `test_pi07_paligemma_high_level_planner.py`, `test_value.py`, `test_policies.py`, `test_save_pretrained_skip_stats.py`, `test_normalize_per_dataset.py`: **9 passed, 10 skipped (pre-existing), 0 failed** in 3:08.

Combined: **15 passed, 10 skipped, 0 failed** — same outcome as the prior run on the RTX 5090. The 10 skips are the pre-existing unconditional `@pytest.mark.skip` markers (memory / env), unaffected by this branch.

**CI status**: all gating checks green on `4039691` — CPU Tests, Pre-commit (was failing on `1ded440` due to two stray formatting issues outside the touched-file set I'd been checking; `pre-commit run --all-files` is now clean), check-checklist, Claude review.

shuheng-liu:
Thanks for the third pass — addressed in `98f5c71`.

**#1 (DTensor-safe inf check)** — fixed. The inline check inside `_assert_normalize_buffers_initialized` now wraps the parameter in `_materialize(...)` (imported from [normalize.py](src/opentau/policies/normalize.py)) before `torch.isinf(...).any()`. Today's call site runs before `accelerator.prepare`, so params are plain Tensors and `_materialize` is a no-op; but the helper is now safe for any future caller that invokes it on a `fully_shard`'d model. Also added a comment inside the iteration block calling out both the FSDP2 motivation and the `named_parameters`-vs-`named_buffers` convention (`Normalize` uses `nn.Parameter(requires_grad=False)`, not `register_buffer` — addresses your finding #3).

**#2 (reset is in-memory only)** — fixed. Both the [`PreTrainedConfig.skip_normalization_weights` docstring](src/opentau/configs/policies.py) and the helper docstring on `_strip_normalization_buffers_from_state_dict` now say "in-memory only" and explicitly call out the two paths: the in-process resume (`train.py` → `save_checkpoint` → `cfg.save_pretrained(...)`) does persist `False`; re-running `from_pretrained` on the *source* path reloads `True` and re-strips.

**#3 (naming nit)** — handled inline alongside #1 in the comment block.

**#4 (logging vs print inconsistency in pi05)** — left as-is. Pre-existing pattern across the file (every `from_pretrained` override in pi05 uses `print`, and the surrounding remap loop also uses `print`); making the strip line use `print` to match the file would be wrong (we want it in `logging`), and converting the whole pi05 file is out of scope. Flagging for a follow-up that converts pi05 to `logging` end-to-end.

**#5 (pre-commit clean)** — confirmed; running on all files is clean.

**Note on `_materialize` import**: I'm importing `_materialize` (underscore-prefixed) directly from `normalize.py` since you pointed at it explicitly in the review and it's the same helper `Normalize.forward` uses. Happy to promote it to a public `materialize` (no underscore) if you'd prefer.

**Verification**:
- Pre-commit clean on `pre-commit run --all-files`.
- 30 targeted CPU tests pass (`tests/policies/test_pretrained_skip_normalization.py` + `tests/policies/test_pi07_paligemma_low_level.py::TestSkipNormalizationWeights`).
- The change is doc + a `_materialize` wrap; no behavioural change to the strip / load path → re-running GPU tests would be a no-op against the previous run's 15-passed / 10-skipped result.

shuheng-liu (commented):
## Fourth-pass review — verification + cleanup nits

Diminishing returns at this point; nothing new of substance. A couple of verifications + two micro-nits.

### Verified — registry coverage is complete

Walked every `@PreTrainedConfig.register_subclass(...)` decorator in `src/opentau/policies/`:

| Registered name | Test fixture |
|---|---|
| `pi0` | ✓ `PI0Config` |
| `pi05` | ✓ `PI05Config` |
| `pi05_continuous_state` | ✓ `PI05ContinuousStateConfig` |
| `pi05_mem` | ✓ `PI05MemConfig` |
| `pi06` | ✓ `PI06Config` |
| `pi07_low_level` | ✓ `PI07LowLevelConfig` |
| `pi07_high_level` | ✓ `PI07HighLevelPlannerConfig` (pi07) |
| `pi07_paligemma_low_level` | ✓ `PI07PaligemmaLowLevelConfig` |
| `pi07_paligemma_high_level_planner` | ✓ `PI07HighLevelPlannerConfig` (pi07_paligemma) |
| `value` | ✓ `ValueConfig` |

10/10. `DummyPolicyConfig` in `datasets/v21/convert_dataset_v20_to_v21.py` is the only other `PreTrainedConfig` subclass and is correctly excluded (it's an internal converter helper, not a registered policy). PR body's claim "all ten policy configs" checks out.

### Verified — `unintended_missing` math is correct

I worked through the set algebra to make sure the `missing - stripped_keys` filter can't accidentally hide a real "model expects key K but state_dict didn't have it" error. Sketch, using `F` = file keys, `M` = model keys, `F_norm = F ∩ {normalize_*, unnormalize_*}`:

- After strip, `state_dict` has keys `F − F_norm`.
- `load_state_dict(strict=False)` reports `missing = M − (F − F_norm) = (M − F) ∪ (M ∩ F_norm)`.
- `unintended_missing = missing − F_norm`.
- Because `F_norm ⊆ F`, `F_norm` is disjoint from `M − F`, so the subtraction only removes the `M ∩ F_norm` part — i.e. exactly the buffers we deliberately stripped *and* the model expected.
- Net: `unintended_missing = M − F` — the original "missing because the file didn't have it" set, with the deliberately-dropped-and-immediately-missing keys hidden. No false negatives for unrelated typos / renamed model keys.

Also confirms that if the *file* had a typo like `normalize_inpts.buffer_X.mean` (and the model expects `normalize_inputs.buffer_X.mean`), the typo'd key is silently stripped (matches the predicate) and the model's correct key still surfaces in `unintended_missing` — so users get a recoverable signal. Not a regression vs pre-PR behaviour.

### Micro-nit 1: pointless dict copy on the "flag set but no keys matched" path

`pretrained.py:244-245`:

```python
stripped_keys: frozenset[str] = frozenset(key for key in state_dict if is_normalize_buffer_key(key))
filtered_state_dict = {key: val for key, val in state_dict.items() if key not in stripped_keys}
```

When `stripped_keys` is empty (flag on, nothing to strip — the WARNING path), the comprehension still builds a fresh dict identical to the input. For a state-dict with thousands of keys (every PaliGemma checkpoint has ≥1.5k tensors), that's a wasteful O(N) copy plus a hash lookup per key against the empty frozenset. Short-circuit:

```python
if not stripped_keys:
    if is_main_process:
        logging.warning(...)
    config.skip_normalization_weights = False
    return state_dict, stripped_keys
```

Also has the side effect of making the no-strip-fired path return the input dict by identity, matching the existing flag-off contract (`test_no_op_when_flag_off` asserts `out_sd is sd`). Today the two "no effect" paths return different objects (identity vs. copy) — `test_warning_when_flag_set_but_no_keys_match` uses `==` not `is` so it tolerates either, but the asymmetry is cosmetic noise.

### Micro-nit 2: docstring claim about returning a "new dict" is half true

The helper docstring (`pretrained.py:211-213`):

> Returns a new dict with every key matching `is_normalize_buffer_key` removed, plus the (possibly empty) `frozenset` of stripped keys.

This describes only the strip-fired path. The flag-off path returns the *input* dict (identity, tested explicitly). The all-paths summary at the end (`Returns:` section, lines 236-239) is correct (`When the flag is off, this is (state_dict, frozenset()) — the input dict is returned unchanged`), but the body description in the bullet at line 211 reads as if every call produces a new dict. One word ("a *filtered* dict" / "a *fresh* dict for the strip path") would resolve.

### Summary across all four passes

Genuinely actionable items, in priority order:

1. **Second-pass #1** — `strict=True` + `skip_normalization_weights=True` divergence in `pi0._load_as_safetensor` / `value._load_as_safetensor`. Behaviour split between policies. Mechanical fix.
2. **Second-pass #2** — `is_main_process` not threaded through the strip helper for pi0 / value. Log duplication under DDP.
3. **Third-pass #1** — defensive `_materialize` in `find_inf_normalize_buffers` for FSDP2 safety.
4. **Fourth-pass micro-nit 1** — short-circuit the empty-`stripped_keys` dict copy.

Everything else is docstring polish or pre-existing patterns. Won't review further unless something material lands.

---

shuheng-liu:
Thanks for the verification + fourth pass — fixes in `ac2185b`.

**Micro-nit 2 (docstring "new dict" claim)** — fixed. The strip-fired bullet now reads:

> On the strip-fired path (at least one matching key in the dict), returns a *filtered* dict with every key matching `is_norm_buffer_key` removed, plus the `frozenset` of stripped keys. On the no-match path (flag set but no `normalize_*` / `unnormalize_*` keys in the dict), returns the input dict unchanged by identity — same as the flag-off contract — paired with an empty `frozenset`.

**Micro-nit 1 (pointless dict copy on no-match path)** — fixed. Restructured the helper to short-circuit when `stripped_keys` is empty: emits the WARNING + resets the flag + returns `(state_dict, frozenset())` by identity. Skips the O(N) comprehension that was previously building a fresh dict identical to the input — meaningful on PaliGemma-class checkpoints (~1.5k tensors). Side effect aligns the two no-effect paths: both now return the input by identity, matching `test_no_op_when_flag_off`'s `out_sd is sd` contract. `test_warning_when_flag_set_but_no_keys_match` uses `==` so it's unaffected.

Thanks for the registry-coverage walk and the `unintended_missing` set-algebra trace — appreciated.

**Verification**: pre-commit clean; 28 targeted tests pass; no behavioural change beyond the short-circuit performance fix (and the identity-return alignment on the no-match path).

PR #342, opened 2026-05-27
fix(scripts): inject dataset_index in inference smoke obs

### LINKED ISSUE #341: Nightly regression: Run Inference fails with per-dataset normalization KeyError
## Summary

The **Train with Model Parallelism** job in the nightly **Regression Tests** workflow is failing at the **Run Inference** step.

- Failing run: https://github.com/TensorAuto/OpenTau/actions/runs/26512081390/job/78078921626
- Date: 2026-05-27 (nightly, `main`)

Training and all checks (loss drop, grad norm, grad sync, checkpoint convert) pass — only inference fails.

## Error

```
File "src/opentau/scripts/inference.py", line 62, in inference_main
    _ = policy.sample_actions(observation)
File "src/opentau/policies/pi05/modeling_pi05.py", line 615, in sample_actions
    dataset_index = self._resolve_dataset_index(batch)
File "src/opentau/policies/pretrained.py", line 396, in _resolve_dataset_index
    raise KeyError(...)
KeyError: 'Per-dataset normalization with 2 datasets requires either `dataset_index`
(LongTensor of shape (B,)) or `dataset_repo_id` (str or list[str] of length B) in the batch.'
```

## Root cause

PR #336 (`feat(policies): per-dataset normalization + skip-stats safetensors flag`) added `_resolve_dataset_index` in `src/opentau/policies/pretrained.py`, which raises when the policy has per-dataset stats for >1 dataset but the batch carries neither `dataset_index` nor `dataset_repo_id`.

The regression CI config trains on **2 datasets**, so the saved checkpoint has per-dataset stats with 2 rows. But `inference.py`'s `create_dummy_observation` builds a batch with only `['camera0', 'camera1', 'state', 'prompt', 'img_is_pad', 'action_is_pad']` — no `dataset_index`/`dataset_repo_id` — so `sample_actions` raises immediately.

## Suggested fix

Have `create_dummy_observation` (or `inference.py`) populate `dataset_index` (e.g. zeros of shape `(B,)`) when the loaded policy uses per-dataset normalization, so the inference smoke test mirrors how a real deployment must supply the dataset selector. Confirm whether real inference callers are also expected to pass this and document accordingly.

## Reproduction

Nightly `regression_test.yml` → job `Train with Model Parallelism` → step `Run Inference` (`src/opentau/scripts/inference.py`) against a checkpoint trained on the 2-dataset CI config.

shuheng-liu:
## What this does

Fixes #341. The nightly **Regression Tests** workflow's `Run Inference` step has been failing against the 2-dataset CI checkpoint with:

```
KeyError: 'Per-dataset normalization with 2 datasets requires either
`dataset_index` (LongTensor of shape (B,)) or `dataset_repo_id` (str or
list[str] of length B) in the batch.'
```

After #336 landed per-dataset Normalize/Unnormalize, multi-dataset checkpoints require the batch to carry either `dataset_index` or `dataset_repo_id`. The gRPC server and eval script already inject one; `create_dummy_observation` (used only by `src/opentau/scripts/inference.py` smoke test) did not.

This PR pins the smoke-path batch to dataset row 0:

- Single-dataset checkpoints: no behavior change — the `_resolve_dataset_index` fallback would have resolved to the same `zeros((1,))`.
- Multi-dataset checkpoints: the smoke test runs the first dataset's stats, which is what we want for a "just exercise the forward pass" benchmark.

Real deployment callers (`grpc/server.py`, `eval.py`) continue to read their own `dataset_repo_id` config field — those paths already worked, this PR doesn't touch them.

🐛 Bug

## How it was tested

- Added `test_create_dummy_observation_includes_dataset_index` in `tests/utils/test_utils_utils.py` to lock in the regression.
- Re-ran `tests/utils/test_utils_utils.py` and `tests/policies/test_normalize_per_dataset.py` locally — all pass.
- Cannot reproduce the CI `Run Inference` step end-to-end locally without a GPU and a trained checkpoint; the unit test exercises the only line that was wrong, and the failure mode (`_resolve_dataset_index` KeyError) is purely batch-content-driven so a fresh CPU forward through a 2-dataset policy would hit the same guard.

```bash
pytest -sx tests/utils/test_utils_utils.py::test_create_dummy_observation_includes_dataset_index
```

To re-verify the original failure scenario, the next nightly `Regression Tests` run on `main` (post-merge) should show the `Run Inference` step passing again. Alternatively, re-trigger via `workflow_dispatch` on `.github/workflows/regression_test.yml`.

## Checklist

### DIFF
--- a/src/opentau/utils/utils.py
+++ b/src/opentau/utils/utils.py
@@ -402,6 +402,11 @@ def create_dummy_observation(cfg, device, dtype=torch.bfloat16) -> dict:
         "prompt": ["Pick up yellow lego block and put it in the bin"],
         "img_is_pad": torch.zeros((1, cfg.num_cams), dtype=torch.bool, device=device),
         "action_is_pad": torch.zeros((1, cfg.action_chunk), dtype=torch.bool, device=device),
+        # Pin per-sample normalization to dataset row 0 so checkpoints trained
+        # with per-dataset stats on >1 datasets don't trip
+        # `_resolve_dataset_index`'s "missing selector" guard. A no-op for
+        # single-dataset policies (the fallback would resolve to the same).
+        "dataset_index": torch.zeros((1,), dtype=torch.long, device=device),
     }
 
 
--- a/tests/utils/test_utils_utils.py
+++ b/tests/utils/test_utils_utils.py
@@ -25,6 +25,7 @@ import torch
 from opentau.utils.hub import HubMixin
 from opentau.utils.utils import (
     capture_timestamp_utc,
+    create_dummy_observation,
     encode_accelerator_state_dict,
     format_big_number,
     get_channel_first_image_shape,
@@ -390,3 +391,24 @@ def test_another_invalid_shape_raises_value_error():
 def test_encode_accelerator_state_dict(obj, expected):
     output = encode_accelerator_state_dict(obj)
     assert output == expected, f"Expected {expected}, but got {output} for input"
+
+
+def test_create_dummy_observation_includes_dataset_index():
+    """Regression for #341: the smoke-test observation must carry a
+    `dataset_index` so multi-dataset checkpoints don't crash inside
+    `_resolve_dataset_index`.
+    """
+
+    @dataclass
+    class _Cfg:
+        resolution: tuple = (224, 224)
+        num_cams: int = 2
+        max_state_dim: int = 32
+        action_chunk: int = 10
+
+    obs = create_dummy_observation(_Cfg(), device=torch.device("cpu"), dtype=torch.float32)
+
+    assert "dataset_index" in obs
+    assert obs["dataset_index"].shape == (1,)
+    assert obs["dataset_index"].dtype == torch.long
+    assert torch.equal(obs["dataset_index"], torch.zeros(1, dtype=torch.long))


shuheng-liu (commented):
## Review

**Verdict: LGTM, minor non-blocking suggestion.**

### Summary
Minimal, well-scoped fix for #341. After #336 introduced per-dataset Normalize/Unnormalize, `_resolve_dataset_index` raises `KeyError` when a multi-dataset checkpoint sees a batch with neither `dataset_index` nor `dataset_repo_id`. `scripts/inference.py`'s smoke path was the only inference caller missing the injection — `grpc/server.py` and `eval.py` already inject via their `dataset_repo_id` config field.

### Correctness — verified
Cross-checked the "no behavior change for single-dataset" claim against `src/opentau/policies/pretrained.py::_resolve_dataset_index`:

- **Old single-dataset path** (no key in batch, `_stacked_num_datasets() <= 1`): returns `torch.zeros(batch_size or 1, dtype=torch.long, device=self._model_device())`.
- **New explicit path** (`"dataset_index" in batch`): `idx.to(dtype=torch.long, device=self._model_device())`.

Both produce a `(1,)` `torch.long` tensor on the policy's compute device — the claim holds. The "pin multi-dataset smoke runs to row 0" semantics is a reasonable convention for a "just exercise the forward pass" benchmark.

Also consistent with the existing pattern in `src/opentau/scripts/export_to_onnx.py`, which pre-injects `torch.zeros(bsize, dtype=torch.long, ...)` for the same reason (a traced wrapper bypassing `_resolve_dataset_index`).

### Test
The regression test correctly:
- Uses a minimal `_Cfg` dataclass with only the attributes the function reads.
- Asserts presence + shape + dtype + value of `dataset_index`.
- Stays CPU/float32 — no GPU dependency.

The acknowledged limitation (no end-to-end through a 2-dataset policy) is the right trade-off: it would require a heavy fixture, and the unit test locks in the exact line that was wrong.

### Style / conventions
- The 4-line comment is slightly long for a 1-line addition, but the WHY is genuinely non-obvious (`_resolve_dataset_index`'s "missing selector" guard + the single-dataset no-op invariant), so it earns its keep. Leave as-is.
- Dtype `torch.long` is correct — `_resolve_dataset_index` re-casts to long anyway, but providing it from the source skips that cast.

### Minor suggestion (non-blocking)
`create_dummy_observation`'s docstring `Returns:` block lists "camera observations, state, prompt, and padding flags" — consider tacking on "and dataset index" so the new key is mentioned. Trivial, fine to skip.

### Risk
Low. Surgical (5 + 22 lines, single dict-key addition), no public API change, all CI green (CPU tests, pre-commit, model parallelism, checklist). The nightly `Regression Tests` workflow on `main` post-merge should confirm the `Run Inference` step recovers.

---

WilliamYue37 (approved):

PR #344, opened 2026-05-27
fix(policies): mask zero-pad action dims in flow-matching MSE

shuheng-liu:
## What this does

For heterogeneous-DoF co-training, `actions` is zero-padded along the last axis to `max_action_dim` to keep batches rectangular. The existing per-timestep mask `action_is_pad` broadcasts across every dim column, so the flow-matching MSE on the velocity field was scoring the padded tail dims against zero — supervising the action expert on dimensions a given sample's source dataset doesn't use.

This gets strictly worse under per-dataset normalization (#336): for a 7-DoF dataset, dims 7..31 get `mean=0, std=0` stats, so the normalized target is exactly `(0 - 0) / (0 + 1e-6) ≈ 0`. That's a clean "predict 0 here" signal that contaminates samples in the same batch which actually use those dims.

The fix exposes the real (pre-pad) action dim per sample so each policy's MSE can skip the padded columns:

- **Dataset** (`src/opentau/datasets/lerobot_dataset.py`): `_to_standard_data_format` now emits `action_dim` — a scalar `torch.long` (shape `()`) captured before `pad_vector` is applied. Default PyTorch collation stacks it into a `(B,)` tensor. A `ValueError` (not `assert`, so it survives `python -O`) fires if a dataset declares more dims than the policy config allows.
- **Helper** (`src/opentau/policies/utils.py`): new `make_action_dim_mask` builds `(B, max_action_dim)` bool from `action_dim`. Returns all-True when `action_dim is None`, so behavior is bit-identical to today for callers that don't supply it.
- **Per-policy MSE** (all six flow-matching call sites): AND the per-dim mask into the existing `postfix_mask` along the dim axis before reducing. Files touched:
  - `src/opentau/policies/pi0/modeling_pi0.py` (see "**pi0 loss-magnitude shift**" below)
  - `src/opentau/policies/pi05/modeling_pi05.py`
  - `src/opentau/policies/pi05_mem/modeling_pi05.py`
  - `src/opentau/policies/pi06/modeling_pi06.py` (extended `flow_matching_masked_mse` signature with optional `action_dim`)
  - `src/opentau/policies/pi07/low_level/modeling_pi07_low_level.py`
  - `src/opentau/policies/pi07_paligemma/low_level/modeling_pi07_low_level.py`
- **Docs** (`docs/source/concepts.rst`): documents `action_dim` in both Standard Data Format code blocks and adds an "Action-dim padding mask" subsection explaining the relationship with `action_is_pad` and the backwards-compatibility contract.

Discrete-action CE loss and inference paths (`select_action` / `sample_actions`) are unchanged — they already truncate via `config.action_feature.shape[0]` and have their own per-token masking. Padding itself is unchanged; only the MSE reduction now skips the zero-pad columns.

### pi0 loss-magnitude shift (heads-up for anyone tracking pi0 training-loss curves)

pi0 was the one policy that previously reduced its flow-matching MSE with `losses.mean()` rather than `sum / mask.sum()`. The old code zeroed out timestep-padded slots in the numerator but still counted them in the denominator, so the per-element mean was biased downward in proportion to how much episode-boundary `action_is_pad` was present.

This PR switches pi0 to the same `sum / mask.sum()` pattern the other five policies already use. The gradient direction is unchanged; only the reported MSE scale shifts upward by roughly `total_slots / unmasked_slots`. Equivalent to today when `action_is_pad` is all-False; deviates when episode-boundary padding exists. The new reduction is more semantically correct (a true mean over real targets) and brings pi0 in line with the rest of the policies, but pre-merge / post-merge pi0 MSE values aren't directly comparable.

### Deliberate scope: inline per-policy MSE blocks, not a shared helper

The five policies outside pi06 still have inlined 3-4 line MSE blocks rather than going through a shared helper. This was a deliberate planning trade-off (minimal diff over consolidation) per CLAUDE.md's "don't refactor beyond what the task requires." Each block follows the exact same pattern (`postfix_mask & rearrange(dim_mask, "b d -> b 1 d")` → multiply → sum/mask.sum()), and the bit-identical equivalence tests on `flow_matching_masked_mse` (pi06's helper) catch the most likely class of regression. A future consolidation pass — likely when pi06's helper is generalized to handle discrete-action signatures too — could fold all six into one place.

## How it was tested

- **9 unit tests** in `tests/policies/test_action_dim_mask.py` exercise `make_action_dim_mask` and the per-policy MSE reduction directly with hand-crafted `u_t` / `v_t` tensors. Two of them explicitly assert bit-identical equivalence with the pre-fix sum/denom computation when `action_dim` is `None` or equals `max_action_dim` — the no-op backwards-compatibility guarantee.
- **4 integration tests** in `tests/datasets/test_action_dim.py` build minimal `LeRobotDataset` instances at 4-DoF and 9-DoF and verify: (a) `__getitem__` emits the right per-sample `action_dim`, (b) default PyTorch collation stacks it to a `(B,)` long tensor, (c) a heterogeneous 2-dataset mix produces the right values per row, (d) declaring more dims than `max_action_dim` raises `ValueError`.
- `tests/datasets/test_datasets.py::check_standard_data_format` extended to require `action_dim` on every standardized item.
- `pytest -m "not gpu" -n auto` clean (1370 passed locally; 3 pre-existing LIBERO failures unrelated to this change).
- `pytest -m "gpu" -n 0` on an internal GPU box: **19 passed, 10 skipped, 0 failed** in 2:43.

```bash
git fetch && git checkout claude/nervous-galileo-1d23a0
pytest -sx tests/policies/test_action_dim_mask.py
pytest -sx tests/datasets/test_action_dim.py
```

To confirm the no-op default is bit-identical for single-dataset configs:
```bash
pytest -sx tests/policies/test_action_dim_mask.py::test_flow_matching_masked_mse_none_action_dim_matches_old_behavior
pytest -sx tests/policies/test_action_dim_mask.py::test_flow_matching_masked_mse_action_dim_at_max_matches_old_behavior
```

## Checklist

### DIFF
--- a/docs/source/concepts.rst
+++ b/docs/source/concepts.rst
@@ -124,16 +124,6 @@ dims — under per-dataset normalization (where padded-dim stats are
 ``mean=0, std=0``) that signal is a clean "predict 0 here" that contaminates
 samples in the same batch which do use those dims.
 
-The same per-dim mask is also applied to the sampled flow-matching noise
-*before* the action expert's input embedding (``action_in_proj``). This
-zeros out ``noise`` at padded dims, which in turn makes ``x_t =
-(1-t)*noise + t*actions`` and the target velocity ``u_t = noise - actions``
-both zero at padded dims — so the action expert never sees the noise that
-would otherwise leak into predictions at *real* dims via attention. The
-input-side mask is complementary to the loss-side mask: the loss-side mask
-ensures padded slots don't contribute to the gradient; the input-side mask
-ensures padded slots don't contaminate the embedding.
-
 This is distinct from ``action_is_pad``, which masks padded *timesteps*
 along the action chunk. The two masks are AND-ed together inside each
 policy's MSE block: a slot in ``(B, chunk_size, max_action_dim)`` contributes
--- a/docs/source/concepts.rst
+++ b/docs/source/concepts.rst
@@ -124,6 +124,16 @@ dims — under per-dataset normalization (where padded-dim stats are
 ``mean=0, std=0``) that signal is a clean "predict 0 here" that contaminates
 samples in the same batch which do use those dims.
 
+The same per-dim mask is also applied to the sampled flow-matching noise
+*before* the action expert's input embedding (``action_in_proj``). This
+zeros out ``noise`` at padded dims, which in turn makes ``x_t =
+(1-t)*noise + t*actions`` and the target velocity ``u_t = noise - actions``
+both zero at padded dims — so the action expert never sees the noise that
+would otherwise leak into predictions at *real* dims via attention. The
+input-side mask is complementary to the loss-side mask: the loss-side mask
+ensures padded slots don't contribute to the gradient; the input-side mask
+ensures padded slots don't contaminate the embedding.
+
 This is distinct from ``action_is_pad``, which masks padded *timesteps*
 along the action chunk. The two masks are AND-ed together inside each
 policy's MSE block: a slot in ``(B, chunk_size, max_action_dim)`` contributes
--- a/docs/source/concepts.rst
+++ b/docs/source/concepts.rst
@@ -65,7 +65,7 @@ Default format (``n_obs_history=None``):
 
         "img_is_pad": torch.BoolTensor,  # shape (num_cams,) with values 0 or 1, where 1 indicates that the camera image is a padded image.
         "action_is_pad": torch.BoolTensor,  # shape (action_chunk,) with values 0 or 1, where 1 indicates that the action is a padded action.
-        "action_dim": torch.LongTensor,  # scalar shape (); collates to (B,). Real (pre-pad) trailing dim of ``actions``. Used by per-policy flow-matching MSE to skip the zero-pad dim columns; see "Action-dim padding mask" below.
+        "real_action_dim": torch.LongTensor,  # scalar shape (); collates to (B,). Real (pre-pad) trailing dim of ``actions``. Used by per-policy flow-matching MSE to skip the zero-pad dim columns; see "Action-dim padding mask" below.
         "obs_history_is_pad": torch.BoolTensor,  # shape (1,) — always False when n_obs_history is None.
     }
 
@@ -84,7 +84,7 @@ With observation history (``n_obs_history=T``):
 
         "img_is_pad": torch.BoolTensor,  # shape (num_cams,) — camera slot availability.
         "action_is_pad": torch.BoolTensor,  # shape (action_chunk,)
-        "action_dim": torch.LongTensor,  # scalar shape (); collates to (B,). Real (pre-pad) trailing dim of ``actions``.
+        "real_action_dim": torch.LongTensor,  # scalar shape (); collates to (B,). Real (pre-pad) trailing dim of ``actions``.
         "obs_history_is_pad": torch.BoolTensor,  # shape (T,) — True for timesteps outside the episode boundary.
     }
 
@@ -115,33 +115,39 @@ Action-dim padding mask
 For heterogeneous co-training across datasets with different native action
 dimensionalities (e.g. a 7-DoF arm dataset and a 14-DoF bimanual dataset in
 one mixture), ``actions`` is zero-padded along the last axis to
-``max_action_dim`` to keep batches rectangular. The per-sample ``action_dim``
-scalar records the real (pre-pad) trailing dim; each policy's flow-matching
-MSE on the velocity field uses it to skip the zero-pad columns and only
-score the dims that the source dataset actually uses. Without it, the
-action expert would be supervised against zero targets on the padded tail
-dims — under per-dataset normalization (where padded-dim stats are
-``mean=0, std=0``) that signal is a clean "predict 0 here" that contaminates
-samples in the same batch which do use those dims.
+``max_action_dim`` to keep batches rectangular. The per-sample
+``real_action_dim`` scalar records the real (pre-pad) trailing dim; each
+policy's flow-matching MSE on the velocity field uses it to skip the
+zero-pad columns and only score the dims that the source dataset actually
+uses. Without it, the action expert would be supervised against zero
+targets on the padded tail dims — under per-dataset normalization (where
+padded-dim stats are ``mean=0, std=0``) that signal is a clean "predict 0
+here" that contaminates samples in the same batch which do use those dims.
+
+The name parallels (not collides with) the static ``original_action_dim``
+local variable in each policy's inference path
+(``sample_actions`` / ``select_action``), which is
+``self.config.action_feature.shape[0]`` — the deployed policy's
+homogeneous DoF, distinct from this per-sample dynamic key.
 
 This is distinct from ``action_is_pad``, which masks padded *timesteps*
 along the action chunk. The two masks are AND-ed together inside each
 policy's MSE block: a slot in ``(B, chunk_size, max_action_dim)`` contributes
 to the loss only when both its timestep is real (``~action_is_pad``) and
-its dim is real (column index ``< action_dim``). Policies with a
+its dim is real (column index ``< real_action_dim``). Policies with a
 real-time-inference frozen prefix (pi05, pi05_mem, pi06, pi07,
 pi07_paligemma) AND in a third condition, ``~prefix_mask`` — see
 ``flow_matching_masked_mse`` in :mod:`opentau.policies.pi06.modeling_pi06`
 for the full reduction. pi0 has no such frozen prefix and uses only the
 two conditions above.
 
-When ``action_dim`` is absent (single-dataset configs, externally constructed
-inference batches), all ``max_action_dim`` columns are treated as real and
-loss behavior is bit-identical to pre-fix. VQA-style items emit ``actions``
-already shaped ``(action_chunk, max_action_dim)`` and rely on the all-True
-``action_is_pad`` to drive the loss to zero — ``action_dim`` matches
-``max_action_dim`` for those items, and the all-True ``action_is_pad``
-still zeros the contribution.
+When ``real_action_dim`` is absent (single-dataset configs, externally
+constructed inference batches), all ``max_action_dim`` columns are treated
+as real and loss behavior is bit-identical to pre-fix. VQA-style items emit
+``actions`` already shaped ``(action_chunk, max_action_dim)`` and rely on
+the all-True ``action_is_pad`` to drive the loss to zero —
+``real_action_dim`` matches ``max_action_dim`` for those items, and the
+all-True ``action_is_pad`` still zeros the contribution.
 
 .. _standard-data-format-optional-keys:
 
--- a/docs/source/concepts.rst
+++ b/docs/source/concepts.rst
@@ -128,7 +128,12 @@ This is distinct from ``action_is_pad``, which masks padded *timesteps*
 along the action chunk. The two masks are AND-ed together inside each
 policy's MSE block: a slot in ``(B, chunk_size, max_action_dim)`` contributes
 to the loss only when both its timestep is real (``~action_is_pad``) and
-its dim is real (column index ``< action_dim``).
+its dim is real (column index ``< action_dim``). Policies with a
+real-time-inference frozen prefix (pi05, pi05_mem, pi06, pi07,
+pi07_paligemma) AND in a third condition, ``~prefix_mask`` — see
+``flow_matching_masked_mse`` in :mod:`opentau.policies.pi06.modeling_pi06`
+for the full reduction. pi0 has no such frozen prefix and uses only the
+two conditions above.
 
 When ``action_dim`` is absent (single-dataset configs, externally constructed
 inference batches), all ``max_action_dim`` columns are treated as real and
--- a/docs/source/concepts.rst
+++ b/docs/source/concepts.rst
@@ -65,6 +65,7 @@ Default format (``n_obs_history=None``):
 
         "img_is_pad": torch.BoolTensor,  # shape (num_cams,) with values 0 or 1, where 1 indicates that the camera image is a padded image.
         "action_is_pad": torch.BoolTensor,  # shape (action_chunk,) with values 0 or 1, where 1 indicates that the action is a padded action.
+        "action_dim": torch.LongTensor,  # scalar shape (); collates to (B,). Real (pre-pad) trailing dim of ``actions``. Used by per-policy flow-matching MSE to skip the zero-pad dim columns; see "Action-dim padding mask" below.
         "obs_history_is_pad": torch.BoolTensor,  # shape (1,) — always False when n_obs_history is None.
     }
 
@@ -83,6 +84,7 @@ With observation history (``n_obs_history=T``):
 
         "img_is_pad": torch.BoolTensor,  # shape (num_cams,) — camera slot availability.
         "action_is_pad": torch.BoolTensor,  # shape (action_chunk,)
+        "action_dim": torch.LongTensor,  # scalar shape (); collates to (B,). Real (pre-pad) trailing dim of ``actions``.
         "obs_history_is_pad": torch.BoolTensor,  # shape (T,) — True for timesteps outside the episode boundary.
     }
 
@@ -106,6 +108,35 @@ The following fields are set in ``DatasetMixtureConfig``:
 *   ``history_interval``: Step interval between historical observation steps. Defaults to ``1``. Only relevant when ``n_obs_history`` is set.
 
 Cameras should be labeled in order of importance (e.g. camera0 is the most important camera, camera1 is the second most important camera, etc.). The model dataset will select the most important cameras to use if num_cams is less than the number of cameras in the dataset.
+
+Action-dim padding mask
+~~~~~~~~~~~~~~~~~~~~~~~
+
+For heterogeneous co-training across datasets with different native action
+dimensionalities (e.g. a 7-DoF arm dataset and a 14-DoF bimanual dataset in
+one mixture), ``actions`` is zero-padded along the last axis to
+``max_action_dim`` to keep batches rectangular. The per-sample ``action_dim``
+scalar records the real (pre-pad) trailing dim; each policy's flow-matching
+MSE on the velocity field uses it to skip the zero-pad columns and only
+score the dims that the source dataset actually uses. Without it, the
+action expert would be supervised against zero targets on the padded tail
+dims — under per-dataset normalization (where padded-dim stats are
+``mean=0, std=0``) that signal is a clean "predict 0 here" that contaminates
+samples in the same batch which do use those dims.
+
+This is distinct from ``action_is_pad``, which masks padded *timesteps*
+along the action chunk. The two masks are AND-ed together inside each
+policy's MSE block: a slot in ``(B, chunk_size, max_action_dim)`` contributes
+to the loss only when both its timestep is real (``~action_is_pad``) and
+its dim is real (column index ``< action_dim``).
+
+When ``action_dim`` is absent (single-dataset configs, externally constructed
+inference batches), all ``max_action_dim`` columns are treated as real and
+loss behavior is bit-identical to pre-fix. VQA-style items emit ``actions``
+already shaped ``(action_chunk, max_action_dim)`` and rely on the all-True
+``action_is_pad`` to drive the loss to zero — ``action_dim`` matches
+``max_action_dim`` for those items, and the all-True ``action_is_pad``
+still zeros the contribution.
 
 .. _standard-data-format-optional-keys:
 
--- a/src/opentau/datasets/lerobot_dataset.py
+++ b/src/opentau/datasets/lerobot_dataset.py
@@ -871,11 +871,14 @@ class BaseDataset(torch.utils.data.Dataset):
         # this yields `max_action_dim` — the existing all-True `action_is_pad`
         # still drives the loss to zero, so VQA behavior is unchanged.
         real_action_dim = int(standard_item["actions"].shape[-1])
-        if real_action_dim > self.max_action_dim:
+        if not 0 < real_action_dim <= self.max_action_dim:
             # Raise rather than assert so the check survives `python -O`.
+            # An empty trailing dim (`shape == (chunk, 0)`) would silently emit
+            # `action_dim=0` and contribute zero gradient at training time —
+            # treat it as a malformed dataset, same as exceeding the upper bound.
             raise ValueError(
-                f"real action dim {real_action_dim} exceeds max_action_dim="
-                f"{self.max_action_dim} (dataset={getattr(self, 'repo_id', '?')})"
+                f"real action dim {real_action_dim} is outside (0, "
+                f"{self.max_action_dim}] (dataset={getattr(self, 'repo_id', '?')})"
             )
         standard_item["action_dim"] = torch.tensor(real_action_dim, dtype=torch.long)
 
--- a/src/opentau/datasets/lerobot_dataset.py
+++ b/src/opentau/datasets/lerobot_dataset.py
@@ -865,6 +865,18 @@ class BaseDataset(torch.utils.data.Dataset):
                 continue
             standard_item[new_key] = item[key]
 
+        # Record the real (pre-pad) action dimensionality so per-policy MSE on
+        # the velocity field can mask out the zero-pad tail dims. For VQA-style
+        # items where `actions` is already shaped `(chunk, max_action_dim)`,
+        # this yields `max_action_dim` — the existing all-True `action_is_pad`
+        # still drives the loss to zero, so VQA behavior is unchanged.
+        real_action_dim = int(standard_item["actions"].shape[-1])
+        assert real_action_dim <= self.max_action_dim, (
+            f"real action dim {real_action_dim} exceeds max_action_dim="
+            f"{self.max_action_dim} (dataset={getattr(self, 'repo_id', '?')})"
+        )
+        standard_item["action_dim"] = torch.tensor(real_action_dim, dtype=torch.long)
+
         # pad state and action vectors
         standard_item["state"] = self.pad_vector(standard_item["state"], self.max_state_dim)
         standard_item["actions"] = self.pad_vector(standard_item["actions"], self.max_action_dim)
--- a/src/opentau/datasets/lerobot_dataset.py
+++ b/src/opentau/datasets/lerobot_dataset.py
@@ -871,10 +871,12 @@ class BaseDataset(torch.utils.data.Dataset):
         # this yields `max_action_dim` — the existing all-True `action_is_pad`
         # still drives the loss to zero, so VQA behavior is unchanged.
         real_action_dim = int(standard_item["actions"].shape[-1])
-        assert real_action_dim <= self.max_action_dim, (
-            f"real action dim {real_action_dim} exceeds max_action_dim="
-            f"{self.max_action_dim} (dataset={getattr(self, 'repo_id', '?')})"
-        )
+        if real_action_dim > self.max_action_dim:
+            # Raise rather than assert so the check survives `python -O`.
+            raise ValueError(
+                f"real action dim {real_action_dim} exceeds max_action_dim="
+                f"{self.max_action_dim} (dataset={getattr(self, 'repo_id', '?')})"
+            )
         standard_item["action_dim"] = torch.tensor(real_action_dim, dtype=torch.long)
 
         # pad state and action vectors
--- a/src/opentau/datasets/lerobot_dataset.py
+++ b/src/opentau/datasets/lerobot_dataset.py
@@ -874,13 +874,13 @@ class BaseDataset(torch.utils.data.Dataset):
         if not 0 < real_action_dim <= self.max_action_dim:
             # Raise rather than assert so the check survives `python -O`.
             # An empty trailing dim (`shape == (chunk, 0)`) would silently emit
-            # `action_dim=0` and contribute zero gradient at training time —
+            # `real_action_dim=0` and contribute zero gradient at training time —
             # treat it as a malformed dataset, same as exceeding the upper bound.
             raise ValueError(
                 f"real action dim {real_action_dim} is outside (0, "
                 f"{self.max_action_dim}] (dataset={getattr(self, 'repo_id', '?')})"
             )
-        standard_item["action_dim"] = torch.tensor(real_action_dim, dtype=torch.long)
+        standard_item["real_action_dim"] = torch.tensor(real_action_dim, dtype=torch.long)
 
         # pad state and action vectors
         standard_item["state"] = self.pad_vector(standard_item["state"], self.max_state_dim)
--- a/src/opentau/policies/pi0/modeling_pi0.py
+++ b/src/opentau/policies/pi0/modeling_pi0.py
@@ -37,7 +37,7 @@ from opentau.policies.pi0.paligemma_with_expert import (
     PaliGemmaWithExpertModel,
 )
 from opentau.policies.pretrained import PreTrainedPolicy
-from opentau.policies.utils import log_model_loading_keys
+from opentau.policies.utils import log_model_loading_keys, make_action_dim_mask
 from opentau.utils.accelerate_utils import get_proc_accelerator
 from opentau.utils.utils import get_safe_dtype
 
@@ -490,9 +490,30 @@ class PI0Policy(PreTrainedPolicy):
 
         losses = self.model.forward(images, img_masks, lang_tokens, lang_masks, state, actions, noise, time)
 
+        # Crop to max_action_dim before masking so the shapes are well-defined when
+        # the model's velocity head emits extra trailing dims.
+        losses = losses[:, :, : self.config.max_action_dim]
+
+        # Per-timestep mask (B, chunk, 1) — True for real action steps.
         if actions_is_pad is not None:
-            in_episode_bound = ~actions_is_pad
-            losses = losses * in_episode_bound.unsqueeze(-1)
+            timestep_mask = rearrange(~actions_is_pad, "b c -> b c 1")
+        else:
+            timestep_mask = torch.ones(
+                (losses.shape[0], losses.shape[1], 1), dtype=torch.bool, device=losses.device
+            )
+
+        # Per-dim mask (B, 1, D) — True for real action dims; backwards compatible
+        # all-True fallback when `action_dim` is absent.
+        dim_mask = make_action_dim_mask(
+            batch.get("action_dim"),
+            self.config.max_action_dim,
+            batch_size=losses.shape[0],
+            device=losses.device,
+        )
+        dim_mask = rearrange(dim_mask, "b d -> b 1 d")
+
+        full_mask = timestep_mask & dim_mask  # (B, chunk, D)
+        losses = losses * full_mask
 
         if self.config.use_awr:
             # weight loss based on exponent of advantage. Also clamp at awr_max_weight.
@@ -503,11 +524,8 @@ class PI0Policy(PreTrainedPolicy):
                 ),
                 "b -> b 1 1",
             )
-        # Remove padding
-        losses = losses[:, :, : self.config.max_action_dim]
 
-        # For backward pass
-        loss = losses.mean()
+        loss = losses.sum() / (full_mask.sum() + 1e-8)
 
         return {"MSE": loss, "CE": torch.zeros_like(loss, requires_grad=True)}
 
--- a/src/opentau/policies/pi0/modeling_pi0.py
+++ b/src/opentau/policies/pi0/modeling_pi0.py
@@ -488,7 +488,17 @@ class PI0Policy(PreTrainedPolicy):
         actions = batch["actions"]
         actions_is_pad = batch.get("action_is_pad")
 
-        losses = self.model.forward(images, img_masks, lang_tokens, lang_masks, state, actions, noise, time)
+        losses = self.model.forward(
+            images,
+            img_masks,
+            lang_tokens,
+            lang_masks,
+            state,
+            actions,
+            noise,
+            time,
+            action_dim=batch.get("action_dim"),
+        )
 
         # Crop to max_action_dim before masking so the shapes are well-defined when
         # the model's velocity head emits extra trailing dims.
@@ -869,6 +879,7 @@ class PI0FlowMatching(nn.Module):
         actions: Tensor,
         noise: Tensor | None = None,
         time: Tensor | None = None,
+        action_dim: Tensor | None = None,
     ) -> Tensor:
         """Do a full training forward pass and compute the loss (batch_size x num_steps x num_motors).
 
@@ -881,6 +892,9 @@ class PI0FlowMatching(nn.Module):
             actions: Action tensor.
             noise: Optional noise tensor.
             time: Optional time tensor.
+            action_dim: Optional ``(B,)`` long tensor of real per-sample action
+                dims; when provided, noise is masked at padded dims so the
+                action expert's input embedding doesn't see contaminating noise.
 
         Returns:
             The computed loss tensor.
@@ -890,6 +904,18 @@ class PI0FlowMatching(nn.Module):
 
         if time is None:
             time = self.sample_time(actions.shape[0], actions.device)
+
+        # Zero noise at padded action dims so the action expert's input
+        # embedding never sees noise the model isn't supervised on. The
+        # outer caller (`PI0Policy.compute_loss`) applies the same dim mask
+        # to the returned MSE losses. All-True (no-op) when `action_dim is None`.
+        dim_mask = make_action_dim_mask(
+            action_dim,
+            self.config.max_action_dim,
+            batch_size=actions.shape[0],
+            device=actions.device,
+        )
+        noise = noise * rearrange(dim_mask, "b d -> b 1 d").to(noise.dtype)
 
         time_expanded = time[:, None, None]
         x_t = time_expanded * noise + (1 - time_expanded) * actions
--- a/src/opentau/policies/pi0/modeling_pi0.py
+++ b/src/opentau/policies/pi0/modeling_pi0.py
@@ -488,17 +488,7 @@ class PI0Policy(PreTrainedPolicy):
         actions = batch["actions"]
         actions_is_pad = batch.get("action_is_pad")
 
-        losses = self.model.forward(
-            images,
-            img_masks,
-            lang_tokens,
-            lang_masks,
-            state,
-            actions,
-            noise,
-            time,
-            action_dim=batch.get("action_dim"),
-        )
+        losses = self.model.forward(images, img_masks, lang_tokens, lang_masks, state, actions, noise, time)
 
         # Crop to max_action_dim before masking so the shapes are well-defined when
         # the model's velocity head emits extra trailing dims.
@@ -879,7 +869,6 @@ class PI0FlowMatching(nn.Module):
         actions: Tensor,
         noise: Tensor | None = None,
         time: Tensor | None = None,
-        action_dim: Tensor | None = None,
     ) -> Tensor:
         """Do a full training forward pass and compute the loss (batch_size x num_steps x num_motors).
 
@@ -892,9 +881,6 @@ class PI0FlowMatching(nn.Module):
             actions: Action tensor.
             noise: Optional noise tensor.
             time: Optional time tensor.
-            action_dim: Optional ``(B,)`` long tensor of real per-sample action
-                dims; when provided, noise is masked at padded dims so the
-                action expert's input embedding doesn't see contaminating noise.
 
         Returns:
             The computed loss tensor.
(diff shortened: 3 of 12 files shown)


shuheng-liu:
## Review findings

Solid fix for a real correctness bug. The diagnosis (per-dataset normalization → padded-dim stats `(0, 0)` → clean "predict 0" signal contaminating mixed-DoF batches) is well-articulated, the test coverage is strong (especially the two bit-identical equivalence tests), and the helper API is total / defensive. A few things worth a closer look before un-drafting.

### 1. pi0 `.mean()` → `sum / count` is not strictly a no-op when `actions_is_pad` has Trues

`src/opentau/policies/pi0/modeling_pi0.py` switches from `losses.mean()` to `losses.sum() / (full_mask.sum() + 1e-8)`. The PR body says this is "equivalent when no padding is in play," which is accurate but slightly understates the scope — `actions_is_pad` IS emitted by `_to_standard_data_format` for every item, and commonly contains Trues at episode boundaries when `action_chunk` extends past the end of the episode.

- Old: divide by `B * chunk * max_action_dim` (includes timestep-padded slots in the denominator; their contribution is 0, but they still inflate the denominator).
- New: divide by `full_mask.sum()` (excludes timestep-padded slots from the denominator).

This is arguably **more correct** (a true mean over real targets vs. a mean biased downward by padded slots), and it brings pi0 in line with the other five policies that already used `sum / mask.sum()`. But absolute pi0 MSE values will appear higher post-merge by roughly `total_slots / unmasked_slots` — proportional to how much episode-boundary padding the dataset has. Worth a heads-up in the PR description for anyone tracking pi0 training-loss magnitudes across runs / comparing to pre-merge checkpoints. The gradient direction is unchanged; only the scale shifts.

### 2. Test name `test_action_dim_assertion_when_exceeds_max` is misleading

The function name suggests an assertion fires, but the docstring acknowledges no assertion is triggered — the test just confirms the helper is total (returns all-True when `action_dim > max_action_dim`, same as `None`). The actual hard assertion lives in `_to_standard_data_format`. Consider renaming to `test_helper_is_total_when_action_dim_exceeds_max` (or splitting into two: one that asserts the helper is total, plus one in `tests/datasets/test_action_dim.py` that asserts the dataset-level `AssertionError` fires when a dataset declares more dims than `max_action_dim`).

### 3. `assert real_action_dim <= self.max_action_dim` could be a `ValueError`

This is a data-correctness check at the dataset boundary — exactly the kind of thing that should survive `python -O` (which strips asserts). A `ValueError` would be more defensive. Minor / stylistic.

### 4. Inline per-policy MSE blocks aren't tested directly

The unit tests exercise `make_action_dim_mask` and `flow_matching_masked_mse` (pi06's helper). The other five policies (pi0, pi05, pi05_mem, pi07/low_level, pi07_paligemma/low_level) have structurally similar but inlined MSE blocks. A regression in any one of those — e.g. an off-by-one between `postfix_mask`'s shape and the rearranged `dim_mask` — wouldn't be caught by the current suite. The integration test in `test_action_dim.py` exercises the dataset path only, not the policy MSE.

Not a blocker (the inline code is short and the bit-identical equivalence tests on the pi06 helper catch the most likely class of regression), but the cleanup path here is to extract the inline MSE blocks into a shared helper — pi06's `flow_matching_masked_mse` is already close to what the other five need, modulo discrete-action signature differences — and route everyone through it. That would both eliminate the 5x code duplication and unify test coverage in one place.

### Solid

- `real_action_dim = int(standard_item["actions"].shape[-1])` captured **before** `pad_vector` is applied — correct ordering, easy to get wrong.
- VQA path covered cleanly via `BaseDataset._to_standard_data_format` inheritance: `action_dim = max_action_dim` for VQA items, but the all-True `action_is_pad` still drives `full_mask.sum() = 0` → loss = 0. Bit-identical to pre-fix VQA behavior.
- The strict-less comparison `arange < action_dim` in `make_action_dim_mask` correctly yields all-False for `action_dim = 0` and all-True for `action_dim = max_action_dim`.
- Distributed safety: no per-rank branching introduced; new `action_dim` enters as a normal `(B,)` batch element. Aligns with the CLAUDE.md hard rule #5 patterns.
- Documentation update is accurate and clearly distinguishes the new dim mask from the existing `action_is_pad` timestep mask, including the backwards-compatibility contract.

---

shuheng-liu:
Thanks for the careful review. Addressed in [1a58cec](https://github.com/TensorAuto/OpenTau/commit/1a58cec) plus a PR-body refresh:

**1. pi0 `.mean()` → `sum / mask.sum()` magnitude shift** — added a dedicated subsection in the PR body calling this out explicitly. Keeping the new reduction (it's the semantically correct one and matches the other five policies); just made the heads-up loud for anyone tracking pi0 training-loss curves across the change.

**2. Misleading test name** — renamed to `test_helper_is_total_when_action_dim_exceeds_max` and the docstring now points at the dataset-level boundary test as the place where the hard check is exercised. Added `test_dataset_raises_when_action_dim_exceeds_max` in `tests/datasets/test_action_dim.py` for that.

**3. `assert` → `ValueError`** — done. Survives `python -O`. The new dataset-level test confirms the path raises (matched via `RuntimeError` substring since `__getitem__`'s retry wrapper re-wraps loader errors).

**4. Inline per-policy MSE blocks not directly tested** — kept the inline approach. This was the deliberate plan choice (minimal diff over consolidation per CLAUDE.md "don't refactor beyond the task"). Added a "Deliberate scope" subsection in the PR body explaining the trade-off and pointing at the bit-identical equivalence tests on the pi06 helper as the existing guard for the common regression mode. A future consolidation pass — probably when pi06's helper is generalized to handle the discrete-action signatures — could fold all six call sites into one place.

CPU subset still clean after the changes (13 / 13 in the new files). Will re-run the GPU subset before un-drafting.

shuheng-liu:
## Second-pass review

A closer look surfaced a few things beyond the first round.

### 5. `make_action_dim_mask`'s `batch_size` parameter is silently ignored when `action_dim is not None`

```python
def make_action_dim_mask(action_dim, max_action_dim, batch_size, device):
    if action_dim is None:
        return torch.ones((batch_size, max_action_dim), dtype=torch.bool, device=device)
    arange = torch.arange(max_action_dim, device=device)
    return rearrange(arange, "d -> 1 d") < rearrange(action_dim.to(device=device), "b -> b 1")
```

When `action_dim` is provided, the returned mask has shape `(action_dim.shape[0], max_action_dim)` — `batch_size` is not consulted. All 6 call sites pass `batch_size=<x>.shape[0]` from the matching `losses`/`mse_loss` tensor, so today they're always in sync; but a caller that ever drifts (e.g. passes a sliced `action_dim` but the original `batch_size`) gets a silent shape mismatch downstream rather than a clear error here.

Two cheap options:
- `assert action_dim.shape == (batch_size,)` at the top of the not-None branch, OR
- Drop `batch_size` from the signature and require the caller to either pass `action_dim` (full source of truth) or `None` plus a separate `(fallback_batch_size, max_action_dim)` shape — e.g. via an overload or a separate `_all_true_dim_mask(batch_size, max_action_dim, device)` helper.

The assert is the smaller change and preserves the current API.

### 6. Integration test stops short of the actual heterogeneous-DoF scenario

`tests/datasets/test_action_dim.py::test_two_datasets_with_different_action_dims` builds two datasets at 4-DoF and 9-DoF, fetches one item from each, and asserts the per-item `action_dim` is right. Good — but the bug this PR is fixing is specifically about **a single batch containing samples from both**. The integration test never collates `[item_a, item_b]` into one batch and verifies:
- `batch["action_dim"]` is `tensor([4, 9])`
- a forward pass on a tiny stub policy produces a loss whose numerator excludes the 5-DoF tail of `item_a`'s contribution and the 0-DoF tail of `item_b`'s contribution

Closing this gap is ~20 lines: stack two `_make_dataset(...)` items via `torch.utils.data.default_collate`, instantiate one of the lighter policies (or just call `flow_matching_masked_mse` with a hand-crafted `u_t` shaped `(2, chunk, max_action_dim)` and the collated `action_dim`), and verify the numerator/denominator against a manual count. The current `test_flow_matching_masked_mse_excludes_padded_dims` test does this math but with a synthetic `action_dim` tensor — it never crosses the dataset → policy boundary.

### 7. `flow_matching_masked_mse` requires `prefix_mask` positionally

Pi06's helper is the canonical place per-policy MSE could converge to (this is the same suggestion as item 4 from the first pass). If you do the consolidation, `prefix_mask: Tensor | None = None` with internal default-to-all-False would let policies without an RTI mechanism skip the `torch.zeros(...)` boilerplate. Pure ergonomics, no semantic change.

### 8. Doc subsection — a small inaccuracy

> The two masks are AND-ed together inside each policy's MSE block: a slot in `(B, chunk_size, max_action_dim)` contributes to the loss only when both its timestep is real (`~action_is_pad`) and its dim is real (column index `< action_dim`).

For policies with an RTI-style frozen prefix (pi05 / pi05_mem / pi06 / pi07 / pi07_paligemma), there's a third condition AND-ed in: `~prefix_mask` (frozen-prefix steps are excluded). For pi0 there isn't. Worth a one-line clarification — either "for RTI policies a third frozen-prefix condition is also AND-ed in" or pointing at `flow_matching_masked_mse`'s docstring which already lists all three.

### Things newly verified on the second pass

- `_to_standard_data_format` lines 901-903 cast all floating-point tensors in `standard_item` to bfloat16. `action_dim` is `torch.long`, so it correctly stays int64 through the cast — confirmed by `test_action_dim_default_dataloader_collation`.
- The action expert's velocity head is exactly the contamination surface (target `u_t[padded_dim] = -noise`, input `x_t[padded_dim] = (1-t)*noise`, so the model learns a clean linear map `v_t = -x_t/(1-t)` for padded dims pre-fix). Fix is well-targeted.
- State padding goes through `pad_vector` similarly but doesn't have the same problem: padded state dims go through a linear projection whose contribution is `0 * weight = 0` regardless of sample, so they don't create a "predict 0" signal in the encoder output. No analogous `state_dim` field is needed.
- Discrete-action CE retains the same biased-mean pattern as pre-fix pi0 (divides by all tokens, not unmasked tokens). That's existing behavior in `pi05/pi05_mem/pi06/pi07*` and explicitly out of scope per the PR body. Not raising as a concern, just confirming I checked.

None of items 5–8 are blockers — the fix as written is correct. They're API/test hardening that's cheap now and harder later.

---

shuheng-liu:
Second-pass review addressed in [9eabf9e](https://github.com/TensorAuto/OpenTau/commit/9eabf9e):

**5. Silent `batch_size` mismatch** — `make_action_dim_mask` now raises `ValueError` when `action_dim.shape != (batch_size,)`. Picked the assert-style guard over the API split (kept the current signature). New unit test `test_helper_raises_on_action_dim_batch_size_mismatch` covers it.

**6. Cross-dataset batch coverage** — new `test_cross_dataset_batch_routes_action_dim_into_loss` in `tests/datasets/test_action_dim.py` collates one item from a 4-DoF dataset and one from a 9-DoF dataset via `torch.utils.data.default_collate`, then runs the collated batch through `flow_matching_masked_mse` with hand-crafted `u_t` / `v_t` and asserts the denominator equals `chunk * 4 + chunk * 9` exactly. Closes the dataset→policy boundary the synthetic-`action_dim` unit test leaves open.

**7. `flow_matching_masked_mse` requires `prefix_mask` positionally** — leaving as-is. The reviewer's framing was \"if you do the consolidation\" — since we deliberately didn't consolidate (item 4), the ergonomic-default change only saves a `torch.zeros(...)` boilerplate line in the one helper that already has callers always passing the arg, and pi0 (the policy without a real RTI prefix) doesn't go through this helper anyway. Will pick up if/when we consolidate.

**8. Doc subsection inaccuracy** — added a sentence clarifying that RTI policies (pi05, pi05_mem, pi06, pi07, pi07_paligemma) AND in a third `~prefix_mask` condition on top of `~action_is_pad` and the dim mask, with a pointer at `flow_matching_masked_mse`'s docstring for the full reduction.

Also acknowledged the things you verified on the second pass — confirming the bfloat16 cast loop correctly leaves `action_dim: torch.long` untouched, the velocity-head contamination surface mechanic, that state padding doesn't have the analogous problem, and that discrete-action CE's pre-existing biased-mean is out of scope.

Local: 15 / 15 pass on the new files, pre-commit clean.

shuheng-liu:
## Third-pass review

Diminishing returns at this point — the fix is correct and the prior two passes covered the substantive items. A few small things I noticed on a closer reading:

### 9. Dataset assertion only checks the upper bound

```python
assert real_action_dim <= self.max_action_dim, ...
```

This catches "dataset declares more dims than the policy allows" but not the degenerate case `real_action_dim == 0` (i.e., a dataset whose `actions` tensor has shape `(chunk, 0)`). `test_datasets.py` does enforce `0 < int(item["action_dim"].item())` so the assertion at test-time would fire, but at training-time a malformed dataset would silently emit `action_dim=0` and contribute zero gradient. Tightening to `assert 0 < real_action_dim <= self.max_action_dim` (with both bounds in the error message) closes the loop. Borderline whether worth doing.

### 10. `make_action_dim_mask` docstring slightly oversells the no-op claim

> When ``None``, the returned mask is all-True and the caller's reduction matches the pre-fix behavior.

True for the five policies that used `sum / postfix_mask_expanded.sum()` pre-fix. **Not strictly true for pi0**, which used `losses.mean()` pre-fix and switched to `sum / full_mask.sum()` in this PR — even with `action_dim=None`, when `actions_is_pad` has Trues the denominator differs (same point as item 1 from the first pass). One-line fix: add a parenthetical "(modulo the pi0 `.mean()` → `sum/count` harmonization called out in the PR body)" to the docstring, or just drop the absolute claim and say "the dim-mask becomes a no-op."

### 11. Residual contamination through input-side action embedding (out of scope, but worth flagging)

The fix correctly removes the *direct* contamination signal (loss against `u_t[padded_dim] = -noise`). But the model still sees zero-padded values at the padded tail of its **input** action tokens — those zeros flow through the action embedding (`action_in_proj`) and into the action-expert attention, where they contribute a small (sample-independent) bias to predictions at non-padded dims. Two things to note:

- This is much smaller than the loss-side contamination (no direct supervision pressure), and it's identical across all samples with the same `t`, so it should mostly cancel.
- The clean fix would be to also mask the input embedding at padded dims (e.g. via `actions * dim_mask` before `action_in_proj`), but that's a bigger semantic change and not in scope for what this PR claims to do.

Not asking for a change here; just flagging so future heterogeneous-DoF work knows where the next layer of polish lives.

### Sanity checks that came back clean

- All 6 call sites pass `batch_size=<x>.shape[0]` consistently — no risk of accidental drift today.
- `1e-8` epsilon in the denominator is consistent across all 6 sites (and across the pi06 helper).
- High-level planners (`pi07/high_level_planner`, `pi07_paligemma/high_level_planner`) don't reference action_dim or per-element action MSE — they operate at the subgoal level, so the fix correctly leaves them untouched.
- `n_action_steps` (inference execution horizon) and `chunk_size` (training prediction horizon) are orthogonal to per-sample `action_dim`. No interaction with the trim at `suffix_out[:, -self.config.chunk_size :]`.
- `action_dim` (long) survives the `if value.dtype.is_floating_point` bfloat16 cast at line 901-904 unchanged — confirmed by the explicit dtype assertion in `test_action_dim_default_dataloader_collation`.
- Robust to future normalization changes: even if per-dataset normalization is replaced by something that doesn't yield `(mean=0, std=0)` for padded dims, the dim-mask still correctly excludes them from the loss.

LGTM at the correctness level once items 1–10 (across the three review passes) are at least considered. The blockers from my read are: the pi0 framing in the PR body (item 1) and the integration test gap (item 6); everything else is hardening.

---

shuheng-liu:
Third-pass review addressed in [a8d579b](https://github.com/TensorAuto/OpenTau/commit/a8d579b):

**9. Lower-bound check** — tightened to `0 < real_action_dim <= max_action_dim` in `_to_standard_data_format`. Updated the error message and the existing dataset-level test substring. VQA path unaffected (`real_action_dim = max_action_dim > 0`).

**10. Docstring overclaim re pi0** — clarified that the all-True fallback only neutralizes the dim-mask AND-in. The pi0 `.mean()` → `sum / mask.sum()` harmonization in this PR also shifts loss magnitude when `actions_is_pad` has Trues, even with `action_dim=None` — docstring now says "(the dim-mask itself is still a no-op when `action_dim` is None)" and points at the PR body's pi0 subsection for the broader change.

**11. Input-side embedding contamination** — acknowledged as a future-work item, leaving out of scope for this PR. The reviewer's own framing was that it's much smaller than loss-side contamination (no direct supervision pressure, sample-independent so mostly cancels across the batch), and masking the input embedding would be a bigger semantic change. Flagging here so it's not lost — opening a follow-up issue would be the right home for it if heterogeneous-DoF mixtures end up showing residual interference in practice.

Local: 15/15 pass, pre-commit clean. Both blockers from the reviewer's read (items 1 and 6) were addressed in earlier passes.

shuheng-liu:
## Fourth-pass review

Genuinely at diminishing returns — three passes covered the substantive items. Two small things this round, plus a few confirmations.

### 12. Naming overlap: batch key `action_dim` vs. existing local `original_action_dim`

The inference paths in all six policies already use a variable called `original_action_dim`:

```python
# pi0/pi05/pi05_mem/pi06/pi07*/...sample_actions
original_action_dim = self.config.action_feature.shape[0]
actions = actions[:, :, :original_action_dim]
```

That's a *static* Python int from the policy config (the trained policy's homogeneous DoF), while the new batch key `action_dim` is a *dynamic* `(B,)` long tensor (the per-sample real DoF from the source dataset). The two concepts are different but the names rhyme; a future reader skimming the modeling files could easily conflate them.

Cheap mitigation: rename the dataset emission and helper parameter from `action_dim` → `real_action_dim` (matches the local variable name already used in `_to_standard_data_format`: `real_action_dim = int(standard_item["actions"].shape[-1])`). Slightly more invasive but unambiguous. Or leave as-is and accept the minor naming overlap — your call.

### 13. Pass-through code paths that iterate batch keys all handle the new key correctly

Confirmed by inspection:
- `pretrained.py::_infer_batch_meta_from_dict` (line 337) — `_NON_DATA_BATCH_KEYS = {"dataset_index", "dataset_repo_id"}` skips non-data keys; `action_dim` is a data tensor, so it's correctly treated as a data tensor for batch-size/device inference. If `action_dim` happens to come first in iteration order, it still returns `(B, device)` correctly since it's shape `(B,)`.
- `pi05_mem/modeling_pi05.py::514`, `pi07/low_level/modeling_pi07_low_level.py::615`, `pi07_paligemma/low_level/modeling_pi07_low_level.py::646` build `temporal_batch` dicts via comprehension that excludes only image and state keys — `action_dim` passes through. ✓
- `scripts/actions_mse_loss.py::70`, `scripts/calculate_value.py::177`, `scripts/get_advantage_and_percentiles.py::160` move all tensor values to device — `action_dim` (a long tensor) is moved correctly.

No batch-iteration site needs an explicit update for the new key.

### Process / template compliance

- Commit message: `fix(policies): mask zero-pad action dims in flow-matching MSE` — 59 chars, well under the 80-char limit. Body is descriptive and stays in present-tense. No `Co-Authored-By` line, no claude session URL. Matches the CLAUDE.md guidance.
- PR template: all three sections present (`What this does`, `How it was tested`, `How to checkout & try?`), checklist all-checked. `check-pr-checklist.yml` should pass (and the `check-checklist` job already did per the PR status).
- Author: `Shuheng Liu <wish1104@icloud.com>` ✓
- Label `bug` is applied, assignee `@shuheng-liu` is set.

### Net assessment after four passes

Correctness: solid. Test coverage: strong for the helper, ~one gap at the integration boundary (item 6). API hardening: a few cheap-to-add asserts (items 5, 9). Documentation: accurate, with one small omission (item 8) and one slight overclaim (item 10). The pi0 reduction harmonization (item 1) is the only thing with non-trivial training-side impact and deserves a heads-up in the PR body so anyone comparing pi0 loss curves across the merge isn't surprised.

Nothing here is a blocker on its own; addressing items 1 (PR-body note) and 6 (integration test) would make me actively endorse the merge. Items 2, 3, 5, 8, 9, 10, 12 are paper cuts.

---

shuheng-liu:
On reflection, item 11 (residual input-side contamination through `action_in_proj`) was worth addressing now rather than punting to follow-up — for heterogeneous-DoF mixtures the padded dims aren't sample-independent (they vary per source dataset), so the "should mostly cancel" argument is weaker than it sounded.

Addressed in [1b1d9cf](https://github.com/TensorAuto/OpenTau/commit/1b1d9cf):

**11. Input-side noise masking** — all six policies now build the per-dim mask once (right after sampling `noise`) and apply it twice:
- Multiply into `noise` *before* `x_t = (1-t)*noise + t*actions` and `u_t = noise - actions` are formed. Both end up exactly zero at padded dims (since post-normalization padded actions are also zero), so the action expert's `action_in_proj` never sees the noise contamination.
- Reuse the same mask in the MSE block as the per-dim `(B, max_action_dim)` bool component of the loss mask (no more rebuilding it in two places).

pi0 needed `action_dim` threaded into the inner `Pi0FlowMatching.forward` (the loss-side mask had been applied only by the outer `PI0Policy.compute_loss`).

Two new unit tests in `test_action_dim_mask.py` cover the input-side semantics: `test_input_side_noise_masking_zeros_padded_dims` verifies `x_t` and `u_t` are both zero at padded dims after the mask multiply, and `test_input_side_noise_masking_noop_when_action_dim_none` verifies the all-True fallback leaves `noise` bit-identical (preserves backwards compat at the input embedding too). Doc subsection in `concepts.rst` updated to call out the input-side + loss-side complementarity.

Verified:
- Local: 17 / 17 in the new test files, 402 / 402 in `tests/policies` and the new dataset test, pre-commit clean.
- GPU subset on an internal GPU box: **19 passed, 10 skipped, 0 failed** in 2:45.

shuheng-liu:
Thanks for the fourth pass — agree we're at diminishing returns. Quick disposition:

**#12. `action_dim` vs. `original_action_dim` naming overlap** — leaving as-is. Three reasons:
- The batch key `action_dim` parallels the existing `action_is_pad` / `actions` naming convention; renaming to `real_action_dim` would actually make the *rhyme* closer to `original_action_dim` (both become "modifier + action_dim") rather than further.
- Context disambiguates: `original_action_dim = self.config.action_feature.shape[0]` is unambiguously config-time; `batch.get("action_dim")` is unambiguously per-sample. They never appear in the same expression.
- The `concepts.rst` Standard Data Format section documents the new key explicitly with its semantics, so a reader landing in a modeling file with no prior context has the doc to fall back on.

If `original_action_dim` does become a confusion vector in practice, the cheaper follow-up would be to rename that local (e.g. to `policy_action_dim` or `config_action_dim`) in each `select_action` / `sample_actions` rather than churning the public batch key.

**#13** — appreciated the explicit confirmation that the batch-iteration sites (`_infer_batch_meta_from_dict`, the `temporal_batch` builders, the eval-script device movers) all handle `action_dim` correctly without a new `_NON_DATA_BATCH_KEYS` entry.

**Net assessment & blockers** — items 1 (pi0 magnitude-shift heads-up in PR body) and 6 (cross-dataset batch integration test) — your stated merge-endorsement criteria — were both addressed in earlier passes. Items 2/3/5/8/9/10 are landed. Items 4 (consolidate inline MSE blocks), 7 (default `prefix_mask=None`), and 12 (rename) are noted but deliberately deferred to a future cleanup pass. Item 11 (input-side contamination) ended up addressed in [1b1d9cf](https://github.com/TensorAuto/OpenTau/commit/1b1d9cf) — your "should mostly cancel" argument was weaker for heterogeneous-DoF where the padded dims aren't sample-independent, so worth doing here.

Calling this PR done from the review-iteration loop unless there's a fifth-pass surprise.

shuheng-liu:
## Review of `1b1d9cf` (input-side noise masking)

### Does it target item 11 correctly?

**Mostly yes, with one important gap.** Training-side mechanism is correctly implemented; the inference path was not updated, which reintroduces the same class of contamination at inference instead of eliminating it.

### What the commit gets right

- Masking noise at the start (right after sampling) propagates cleanly through both `x_t = t·noise + (1−t)·actions` and `u_t = noise − actions`. Under the per-dataset normalization assumption (`actions[padded] = 0` post-normalization), both reduce to exactly 0 at padded dims. So `action_in_proj` sees clean zeros, the attention layers don't propagate noise into real-dim predictions, and the target velocity is also zero (redundant with the loss-side mask, but harmless). The math checks out across all six policies.
- Backwards compatibility: when `action_dim is None`, the mask is all-True, the multiply is a no-op, noise is byte-identical — confirmed by `test_input_side_noise_masking_noop_when_action_dim_none`.
- The pi0-specific threading of `action_dim` into the inner `Pi0FlowMatching.forward` (since pi0's loss-side mask is in the outer `PI0Policy.compute_loss`) is the right factoring decision.
- The `.to(noise.dtype)` cast on the bool mask handles the bfloat16/float32 mixed-precision case correctly.

### Item 14 — Train/inference mismatch (the gap)

**This is the real issue with the commit.** I flagged it in the conversation as a precondition for doing input masking; it didn't make it into the implementation.

Inference paths are unchanged. From `pi05::sample_actions` line 1572 (and parallel in pi06::1197, pi0/pi05_mem/pi07/pi07_paligemma):

```python
actions_shape = (bsize, self.config.chunk_size, self.config.max_action_dim)
noise = self.sample_noise(actions_shape, device)   # ← no mask
```

So at inference, `noise[padded] ~ N(0, 1)` flows through `action_in_proj` exactly as before this commit. The training-time invariant ("action_in_proj sees zero at padded dims") doesn't hold at inference. Concretely:

- During training, `x[padded] = 0` always, so `dL/dW[:, padded] = (dL/dy)ᵀ · x[padded] = 0` always. The `W[:, padded]` columns of `action_in_proj` receive **no gradient signal** and stay at initialization (some small random values).
- At inference, those same columns multiply nonzero `(1−t)·noise[padded]` inputs and produce a small random perturbation in the embedding, which leaks through attention into predictions on real dims.

Net effect: the contamination item 11 was about (noise on padded inputs leaking into real-dim predictions) is **shifted from training-time to inference-time**, not eliminated. Pre-commit there was a small mean-zero variance contribution at training and inference, symmetrically. Post-commit there's zero contribution at training and a still-small contribution at inference — but now with no training-time pressure to suppress it.

In some ways this is *worse* than the pre-commit state, because at least the pre-commit state had train/inference consistency: the model saw the same distribution at action_in_proj's padded inputs in both regimes, so whatever the converged behavior was, it was the behavior the model was trained for. The post-commit state has a real distribution shift.

**Fix:** apply the same mask at the inference noise sample. The mask boundary at inference is `config.action_feature.shape[0]` (the homogeneous deployed DoF, already used for output truncation via the `original_action_dim` variable in `sample_actions`). One way:

```python
# After: noise = self.sample_noise(actions_shape, device)
deployed_dim = torch.full((bsize,), self.config.action_feature.shape[0], dtype=torch.long, device=device)
dim_mask = make_action_dim_mask(deployed_dim, self.config.max_action_dim, batch_size=bsize, device=device)
noise = noise * rearrange(dim_mask, "b d -> b 1 d").to(noise.dtype)
```

…or just `noise[:, :, original_action_dim:] = 0` after sampling. Either works.

Without this, item 11 isn't fully addressed.

### Other observations

**Commit message has two small inaccuracies:**

1. "small *sample-independent* bias" — this is my original (wrong) framing from the third-pass review. I corrected it later in our conversation: the noise contribution is *sample-dependent* and *mean-zero*, so it's added variance / input-noise regularization, not a bias. Doesn't affect the implementation but worth fixing in the message before merge.

2. "All six policies now build the per-dim mask once (right after sampling noise) and apply it twice" — true for pi05, pi05_mem, pi07/low_level, pi07_paligemma/low_level. **Not** true for pi0 (inner `Pi0FlowMatching.forward` builds it for noise masking, outer `PI0Policy.compute_loss` rebuilds it for the loss-side mask — two separate `make_action_dim_mask` calls) and **not** true for pi06 (inline `make_action_dim_mask` for noise, then `flow_matching_masked_mse` internally calls `make_action_dim_mask` again). The double-build is functionally fine (mask is cheap) but inconsistent with what the commit message claims.

**Test coverage is shallow:** the two new tests verify the masking *pattern* in isolation (multiply a manually constructed `dim_mask` into a manually constructed `noise`, check shapes). They don't actually invoke any policy's `forward()` to verify that the policy correctly applies the mask before computing `x_t` / `u_t`. A regression that, say, swapped the order of mask multiplication and noise sampling in one of the five copy-pasted blocks wouldn't be caught. Same coverage shape as the loss-side fix (the per-policy inline blocks aren't directly tested), so this is consistent with the rest of the PR, but it does mean the bit-identical no-op claim hinges on test coverage of `make_action_dim_mask` × noise multiplication, not on the actual policies.

### Bottom line

The training-side mechanism is implemented correctly. The conceptual fix is **incomplete** without matching changes to the inference path — and I'd argue the partial fix is mildly worse than no fix because it introduces a train/inference distribution mismatch where there wasn't one before. Recommend addressing item 14 (inference-side masking) before merging this commit; if you'd rather not touch the inference paths, the alternative is to revert `1b1d9cf` and leave item 11 as the documented footnote it was.

---

shuheng-liu:
You're right — item 14 is the load-bearing point. The training-time-only mask in `1b1d9cf` doesn't just leave inference contamination intact; it actively shifts `action_in_proj`'s padded-column weights to "no gradient signal during training" while those same columns still multiply nonzero inputs at inference. The pre-commit state at least had train/inference symmetry. **Net: the partial fix is worse than no fix.**

**Reverted `1b1d9cf` in [`c46ed20`](https://github.com/TensorAuto/OpenTau/commit/c46ed20)** (pushed) and filed [#345](https://github.com/TensorAuto/OpenTau/issues/345) to track the question separately. The issue captures the three options (mask both train + inference, leave as documented footnote, empirically benchmark first), the open questions about how `original_action_dim` truncation interacts with input-side masking, and acceptance criteria. Better to decide that question with deliberation than to bake a partial fix into this PR.

Also acknowledging the commit-message inaccuracies you flagged on `1b1d9cf` ("sample-independent" → should have been "sample-dependent, mean-zero"; the "build once" claim wasn't quite true for pi0/pi06's helper paths). Not a problem now since the commit is reverted, but noted for the follow-up if/when input-side masking is reconsidered.

Test coverage observation also fair — the inline per-policy MSE blocks aren't directly exercised by the new tests, only via the pi06 helper's bit-identical equivalence test. Same shape as the existing PR's coverage, and the GPU subset on a heterogeneous-DoF policy stack passed (19 / 10 / 0), but worth flagging that the bit-identical no-op claim hinges on `make_action_dim_mask` correctness rather than direct per-policy MSE assertions. The future consolidation into a shared helper (item 4) would close this gap in one place rather than five.

With the revert in place, the PR is back to: loss-side fix only, item 11 explicitly documented as a known limitation deferred to [#345](https://github.com/TensorAuto/OpenTau/issues/345). Ready to re-evaluate for merge endorsement.

shuheng-liu:
## Review of post-`1b1d9cf` commits

One new commit, `c46ed20` — a clean revert of `1b1d9cf`.

### Verification

- `git diff 1b1d9cf~1..c46ed20 --stat` is empty → the working tree at `c46ed20` matches the pre-`1b1d9cf` state exactly.
- No `noise = noise * ...` masking lines remain in any of the six policy files.
- No "input-side" doc paragraph or test artifacts left over.
- `Pi0FlowMatching.forward`'s signature is back to its original form (no `action_dim` kwarg) — the inner-vs-outer threading that `1b1d9cf` introduced is fully undone.
- The loss-side mask from `fad330c` and the hardening from `1a58cec` / `9eabf9e` / `a8d579b` are untouched.

### Assessment

This is option 2 from my last review:

> Recommend addressing item 14 (inference-side masking) before merging this commit; if you'd rather not touch the inference paths, the alternative is to revert `1b1d9cf` and leave item 11 as the documented footnote it was.

The right call given the analysis. Doing input-masking properly would require matching inference-side changes (item 14), more careful validation of the train/inference symmetry, and probably a follow-up A/B to confirm it actually helps. None of that is in scope for this PR, which is fundamentally about the loss-side supervision contamination. Item 11 stays a footnote for a future PR if anyone measures a real benefit.

### Branch state

The PR is now back to the four-commit stack I reviewed in passes 1–4:

```
a8d579b fix(datasets+docs): tighten action_dim bound, clarify helper docstring
9eabf9e fix(policies+tests+docs): address second-pass review on action_dim
1a58cec fix(datasets+tests): address review on action_dim boundary check
fad330c fix(policies): mask zero-pad action dims in flow-matching MSE
```

…which is the right scope for this PR. The blockers I identified before (item 1: pi0 framing in PR body; item 6: integration test gap) are the remaining items. Everything else from passes 1–4 is paper cuts that can be addressed or accepted at your discretion.

Nothing further from me — the branch is in a defensible state to un-draft once items 1 and 6 are dispositioned.

---

shuheng-liu:
Thanks — agree the revert is the right disposition, and confirming items 1 and 6 (your stated merge-endorsement criteria) were already in the tree from earlier passes:

**Item 1 — pi0 framing in PR body.** Still there. PR body has a dedicated "pi0 loss-magnitude shift (heads-up for anyone tracking pi0 training-loss curves)" subsection explaining the `.mean()` → `sum / mask.sum()` harmonization, when it deviates (`actions_is_pad` has Trues), and the gradient-direction-unchanged / scale-shift-only nature. Landed alongside the first-pass response.

**Item 6 — cross-dataset batch integration test.** Still there. `tests/datasets/test_action_dim.py::test_cross_dataset_batch_routes_action_dim_into_loss` (line 259, from commit [`9eabf9e`](https://github.com/TensorAuto/OpenTau/commit/9eabf9e)) builds a 4-DoF and a 9-DoF dataset, collates one item from each via `torch.utils.data.default_collate`, runs the result through `flow_matching_masked_mse`, and asserts the masked-mean denominator equals `chunk * 4 + chunk * 9` exactly. That closes the dataset → policy boundary.

PR is already in "ready for review" (un-drafted after the second GPU run during the iteration loop) so nothing further to do on my end. Calling this done from the review-iteration side — happy to address anything else if it surfaces.

shuheng-liu:
## Remaining items summary

For clarity on what's still open across all five review passes:

### Resolved / addressed

| # | Item | Resolution |
|---|---|---|
| 1 | pi0 `.mean()` → `sum/count` framing | "pi0 loss-magnitude shift" subsection added to PR body |
| 2 | Misleading test name `test_action_dim_assertion_when_exceeds_max` | Renamed in `1a58cec` |
| 3 | `assert` → `ValueError` for dataset boundary check | Done in `1a58cec` (survives `python -O`) |
| 5 | `make_action_dim_mask` silently ignored `batch_size` when `action_dim` not None | Now raises `ValueError` on shape mismatch in `9eabf9e` |
| 6 | Integration test didn't cross dataset→policy boundary | `test_cross_dataset_batch_routes_action_dim_into_loss` added in `9eabf9e` |
| 8 | Doc subsection missed `~prefix_mask` AND-in for RTI policies | Clarified in `9eabf9e` |
| 9 | Dataset assertion only checked upper bound | Now `0 < real_action_dim <= max_action_dim` in `a8d579b` |
| 10 | `make_action_dim_mask` docstring overclaimed the no-op | Updated in `a8d579b` to point at the pi0-shift note |
| 11 | Input-side residual noise contamination | Attempted in `1b1d9cf`, cleanly reverted in `c46ed20` — documented as future work pending an A/B |
| 13 | Pass-through batch-iteration sites | Confirmation only; no action needed |
| 14 | Train/inference mismatch from `1b1d9cf` | Resolved by the `c46ed20` revert |

### Open (not merge-blocking)

| # | Item | Type |
|---|---|---|
| 4 | Inline per-policy MSE blocks aren't tested directly per-policy | Scoped out — PR body's "Deliberate scope: inline per-policy MSE blocks, not a shared helper" section explicitly acknowledges this as the minimal-diff trade-off. Future consolidation would fix coverage and code duplication together. |
| 7 | `flow_matching_masked_mse`'s `prefix_mask` is required positional | Style / ergonomics only — `prefix_mask: Tensor \| None = None` with internal default-to-zeros would let non-RTI callers skip the `torch.zeros(...)` boilerplate. Punt to whoever does the helper consolidation in item 4. |
| 12 | Naming overlap: new batch key `action_dim` vs. existing local `original_action_dim` in inference paths | Style / readability only — rhyming names for different concepts (per-sample dynamic vs. config-level static). Cheapest fix is renaming the batch key to `real_action_dim` (already the local variable name in `_to_standard_data_format`). |

### Verdict

Items 4, 7, 12 are paper cuts that can be addressed in follow-up PRs or accepted as-is. **None are merge-blocking.** CI is green, the PR is un-drafted, and the substantive correctness issues from the five review passes are all resolved. Ready to merge pending an approving review.

---

shuheng-liu:
"Fix or push back" disposition on the three open items from the summary:

**Item 12 — fix.** Renamed the batch key from `action_dim` → `real_action_dim` in [`d7802ea`](https://github.com/TensorAuto/OpenTau/commit/d7802ea). Matches the existing internal local var in `_to_standard_data_format` (the dataset internals and the public batch key now agree on the name), and gives the dynamic key a prefix symmetric with `original_action_dim`'s "original" prefix — so neither presents itself as the unmarked "default" name. Doc subsection extended with an explicit cross-reference to `original_action_dim` for readers landing in a modeling file. 12 files touched, all tests pass (452 / 452), pre-commit clean.

Deliberately unaffected by the rename: `max_action_dim`, `original_action_dim`, and the `action_dim: int` CLI args in `fit_fast_tokenizer.py` / `robocasa/server.py` (different subsystems, not the batch key).

**Item 4 — push back.** Inline MSE blocks not being directly tested per-policy is the planning-time trade-off documented in the PR body's "Deliberate scope" subsection — minimal-diff over consolidation per CLAUDE.md's "don't refactor beyond the task." Three reasons it's not worth resolving in this PR:
1. The shape of the regression caught by the missing tests (a typo in one of the five copy-pasted blocks) is also caught by the GPU subset, which exercises each policy's full forward pass — `pytest -m "gpu"` across the relevant policies has passed three times in this PR's lifetime (post-`fad330c`, post-`9eabf9e`/`a8d579b`, post-`c46ed20`).
2. The consolidation that would fix coverage (extract into a shared helper) also touches item 7 (`prefix_mask: Tensor | None = None` default) and would benefit from being a focused refactor PR rather than landing inside this correctness fix.
3. The bit-identical equivalence tests on `flow_matching_masked_mse` cover the math; the copy-pasted blocks are 5 identical applications of that math. A subtle math bug would show up in those tests too.

**Item 7 — auto-deferred per your own framing** ("Punt to whoever does the helper consolidation in item 4"). Will pick up when item 4 lands.

Three GPU runs of the subset already in this PR's history; only the rename in `d7802ea` is new and it's a pure no-op at runtime (same tensor values, same mask construction, just renamed keys/parameters). Holding off on a fourth GPU re-run unless you'd specifically like one.

shuheng-liu:
## Review of `d7802ea` and `848d628` — items 7 and 12 closed

Two new commits since the last pass. Both clean.

### `d7802ea` — `action_dim` → `real_action_dim` (item 12)

- Dataset emission key, helper parameter, all six policies' `batch.get(...)` call sites, and tests updated consistently.
- Verified: no stale `batch.get("action_dim")` or `standard_item["action_dim"]` references remain in `src/opentau/` or `tests/`. The one remaining `"action_dim"` string in the tree is in `scripts/fit_fast_tokenizer.py:1161` — that's an unrelated tokenizer CLI argument, not the batch key.
- Naming is now symmetric: dynamic per-sample = `real_action_dim`, static config-level = `original_action_dim`. The "default name vs. qualified variant" trap I flagged is gone.
- PR-body doc and the `concepts.rst` subsection both updated to the new name and reference the relationship with `original_action_dim`.

### `848d628` — consolidate `flow_matching_masked_mse` + default `prefix_mask` / `actions_is_pad` (items 4 and 7)

The substantive cleanup. Net result:

- `flow_matching_masked_mse` moved from `policies/pi06/modeling_pi06.py` to `policies/utils.py`, joining `make_action_dim_mask`. Single canonical reduction site.
- `prefix_mask: Tensor | None = None` (defaults to all-False internally) and `actions_is_pad: Tensor | None = None` — non-RTI / single-dataset callers no longer need to pass boilerplate.
- `max_action_dim` is keyword-only (after `*`) so the new defaults can coexist without ambiguous positional order. Every existing caller already used keyword args, so no breakage.
- pi05, pi05_mem, pi06, pi07/low_level, pi07_paligemma/low_level — five inline ~25-line MSE blocks → five 8-line helper calls. Diff is net −156 LOC / +190 LOC (the +190 is mostly the helper docstring and two new unit tests).
- **pi0 deliberately stays inline**, with an in-source comment explaining why: AWR weighting needs to apply between the masking and the reduction, and the shared helper reduces internally. Adding a `sample_weights` knob to the helper just for pi0 would be overengineering. The comment makes the choice discoverable.
- Two new tests:
  - `test_flow_matching_masked_mse_prefix_mask_default_none` — verifies the new `prefix_mask=None` default produces the same loss as an explicit all-False `prefix_mask`.
  - `test_flow_matching_masked_mse_actions_is_pad_default_none` — same for `actions_is_pad=None`.

This is the right consolidation. Item 4's coverage gap is now closed structurally: five of the six policies route through one helper that has bit-identical equivalence tests + new default-argument tests, so a regression in the shared reduction would surface across all five callers at once. pi0's inline block is the only remaining un-unit-tested MSE reduction, which the commit message explicitly calls out as covered by end-to-end GPU tests.

### Final ledger

All 14 items from passes 1–4 are now either resolved or correctly scoped out:

| # | Status |
|---|---|
| 1 | ✓ PR body subsection |
| 2 | ✓ `1a58cec` |
| 3 | ✓ `1a58cec` |
| 4 | ✓ **`848d628`** (consolidation) |
| 5 | ✓ `9eabf9e` |
| 6 | ✓ `9eabf9e` |
| 7 | ✓ **`848d628`** (default-None args) |
| 8 | ✓ `9eabf9e` |
| 9 | ✓ `a8d579b` |
| 10 | ✓ `a8d579b` |
| 11 | ✓ deferred via revert (`c46ed20`) |
| 12 | ✓ **`d7802ea`** (rename) |
| 13 | ✓ confirmation only |
| 14 | ✓ resolved by revert |

### CI status

At the time of this comment: pre-commit ✓, check-checklist ✓, claude review ✓, CPU Tests **in progress** on the new `848d628` head. Worth re-checking once the CPU Tests run finishes — given the consolidation touches every flow-matching policy's MSE block, a green CPU run is the load-bearing confirmation.

Assuming CPU Tests pass: nothing left from me. Ready to merge.

---

shuheng-liu:
On request: items 4 and 7 addressed in [`848d628`](https://github.com/TensorAuto/OpenTau/commit/848d628).

**Item 4 — consolidate.** Moved ``flow_matching_masked_mse`` from ``policies/pi06/modeling_pi06.py`` to ``policies/utils.py`` (alongside ``make_action_dim_mask``). It's now the canonical shared MSE reducer. Replaced the inlined per-policy MSE blocks in **pi05, pi05_mem, pi07/low_level, pi07_paligemma/low_level** with calls to it — each policy's MSE block goes from ~25 lines of nearly-identical postfix-mask + dim-mask logic to one 8-line keyword call.

**pi0 stays inline**, with a comment at the reduction site explaining why:
- no RTI prefix (helper's `prefix_mask=None` would always apply, harmless but the rest of the body still differs)
- AWR weighting (`use_awr`) needs to apply between masking and reduction, and the shared helper reduces internally — pulling pi0 in would require an extra ``sample_weights`` knob on the helper just for this one site
- pi0's outer ``compute_loss`` already builds the dim_mask (rather than the inner forward), since pi0's inner returns raw losses
That's the only inline reduction left.

**Item 7 — `prefix_mask: Tensor | None = None` default.** With ``max_action_dim`` made keyword-only (placed after a ``*``), the three masks (``prefix_mask``, ``actions_is_pad``, ``real_action_dim``) all default to ``None`` and fall back internally to all-False / all-True respectively. Non-RTI callers can omit them.

**Coverage gap closed at the source.** There's now exactly one MSE reduction in the codebase (the shared helper) exercised by every policy whose forward calls it; the bit-identical equivalence tests on the helper cover the math for all five callers. Pi0's inline reduction is the one remaining un-unit-tested block, but it's distinct in shape (different from the helper's flow) and is exercised end-to-end by pi0's GPU subset.

**Two new tests** verify the new defaults are equivalent to passing explicit all-False tensors:
- ``test_flow_matching_masked_mse_prefix_mask_default_none``
- ``test_flow_matching_masked_mse_actions_is_pad_default_none``

Verified:
- 17 / 17 in the new test files, 454 / 454 in ``tests/policies`` + new dataset tests + ``test_datasets.py``, pre-commit clean.
- GPU subset on an internal GPU box: **19 passed, 10 skipped, 0 failed** in 2:42 (against [`848d628`](https://github.com/TensorAuto/OpenTau/commit/848d628)).

Net change: 5 inline 25-line MSE blocks → 5 × 8-line shared-helper calls; the helper itself gained one set of defaults. PR is back to defensible, with items 4 and 7 actively addressed rather than deferred.

PR #347, opened 2026-05-28
feat(policies): per-(robot_type, control_mode) normalization heads

### LINKED ISSUE #346: Switch from per-dataset to per-(robot_type, control_mode) normalization heads
## Problem

Normalization stats (`mean` / `std` / `min` / `max`, etc.) are currently keyed **per dataset**. With a large heterogeneous mixture (dozens of LeRobot repos, RECAP shards, human-video conversions, etc.) this produces an equally large number of normalization "heads" stacked into the policy's buffers — one row per dataset. This:

- bloats the policy's `Normalize` / `Unnormalize` buffers linearly in the number of training datasets,
- forces every checkpoint to memorize the full training mixture name list (`PreTrainedConfig.dataset_names`) for inference-time lookup,
- makes new-dataset evaluation awkward: the caller has to supply a `dataset_repo_id` string that was seen at training time, even when the new dataset shares its robot embodiment and control mode with a training dataset,
- treats two shards of the same `franka_arm` + `joint_position` data as if they were unrelated distributions, which they are not.

The natural equivalence class for action/state normalization is **(robot_type, control_mode)** — datasets that share both should share normalization, because they share the units, axis count, and physical ranges of the proprio/action vectors.

## Proposal

Aggregate normalization stats by `(robot_type, control_mode)` instead of by dataset name. Datasets that share both keys get a single shared head whose stats are the **aggregated** mean / std / min / max / etc. over the union of those datasets' samples (re-aggregated via the existing `compute_stats` utilities, not a naive average of per-dataset stats).

### Key resolution rules

For each dataset in the mixture, derive the normalization key as:

1. If overrides supply both `robot_type` and `control_mode` (`_apply_metadata_overrides` in `src/opentau/datasets/factory.py:97-112`), use `(override.robot_type, override.control_mode)`.
2. Otherwise read `dataset.meta.info["robot_type"]` and `dataset.meta.info.get("control_mode", "unknown")` (`LeRobotDataset.robot_type` / `control_mode` properties at `src/opentau/datasets/lerobot_dataset.py:496-503`).
3. **Fallback**: if either field is missing/empty/`"unknown"` *and* there is no override, fall back to using the **dataset name (repo_id) as the key** so the dataset still trains/evals correctly — it just won't share a head with anything else. Emit a warning at policy construction so this isn't silent.

This preserves backward compatibility for datasets without proper robot/control-mode tagging while letting tagged datasets share heads.

### Training path

- `src/opentau/datasets/dataset_mixture.py:83-127` — `_TaggedDataset` currently injects `dataset_repo_id` (str) and `dataset_index` (long). Add a `norm_key` (tuple-encoded string, e.g. `"franka_arm::joint_position"`) and a `norm_index` (long) per sample. Keep `dataset_repo_id` / `dataset_index` around for logging/metrics — they are not the same axis anymore.
- `src/opentau/datasets/compute_stats.py` — extend to aggregate stats over the union of episodes sharing a `(robot_type, control_mode)` key. The aggregation must be correct under the existing `estimate_num_samples` sampling logic (re-weight by per-dataset sample count, not unweighted mean of means).
- `src/opentau/policies/factory.py:218-255` — build a `norm_keys: list[tuple[str, str]]` instead of `dataset_names`, and a stats list of length `len(norm_keys)` rather than `len(datasets)`. Persist `norm_keys` on the policy config.
- `src/opentau/configs/policies.py:323` — replace (or alias, with a deprecation warning) `dataset_names` with `norm_keys`. Provide a migration path that reads the old field if `norm_keys` is absent so existing checkpoints still load.
- `src/opentau/policies/normalize.py:114-296` — the stacked buffers stay shape `(D, *feat_shape)`; only the meaning of `D` and the index-resolution change. `_gather_and_broadcast` (lines 221-246) is unchanged.
- `src/opentau/policies/pretrained.py:119-129, 346-420` — `_resolve_dataset_index` becomes `_resolve_norm_index`. The lookup map is `{(robot_type, control_mode): row_idx}`. Accept the batch field `norm_key` (preferred) or fall back to deriving it from `(batch["robot_type"], batch["control_mode"])` if present, or finally from `batch["dataset_repo_id"]` for back-compat / fallback-keyed datasets.
- All `policies/*/modeling_*.py` call sites that currently invoke `_resolve_dataset_index` (e.g. `policies/pi05/modeling_pi05.py:694-697`) — switch to `_resolve_norm_index`. No behavioral change beyond the new key.

### Inference / eval path

- `src/opentau/scripts/eval.py:172-174` — instead of (or in addition to) `observation["dataset_repo_id"]`, accept `observation["robot_type"]` + `observation["control_mode"]`. Resolve to `norm_index` via the same map the training path builds. Keep accepting `dataset_repo_id` for back-compat / fallback-keyed checkpoints.
- gRPC inference server (`src/opentau/scripts/`) — same: accept robot_type + control_mode in the request proto, with `dataset_repo_id` as the deprecated fallback.
- Hub / local checkpoint loading must read `norm_keys` from the persisted config and rebuild the lookup map.

### Validation

- `src/opentau/datasets/factory.py:308-343` — `_validate_metadata_requirements` already checks robot_type/control_mode presence. Loosen the failure mode: missing fields are no longer fatal, but they trigger the dataset-name fallback above with a warning.

## Tests

All changes must be covered. Concretely:

- `tests/policies/test_normalize_per_dataset.py` — rename file / add a sibling `test_normalize_per_robot_control_mode.py`. Cover:
  - two datasets with the same `(robot_type, control_mode)` resolve to the **same** `norm_index` and share buffer rows,
  - two datasets with different control modes on the same robot get **distinct** rows,
  - a dataset missing `robot_type` falls back to its name as the key, and the warning is emitted exactly once,
  - aggregated stats are computed correctly: a mixture of two datasets with known means/stds yields the closed-form pooled mean / pooled variance (re-weighted by sample count), not the naive average.
- `tests/datasets/test_compute_stats.py` — add cases for the new aggregation path.
- `tests/datasets/test_dataset_mixture.py` (if it exists, else add one) — `_TaggedDataset` emits the new `norm_key` / `norm_index` fields with the right values.
- Add a backward-compat test: load a checkpoint that only has `dataset_names` (the old field) and verify inference still works via the fallback.
- Inference / gRPC: cover both the new `(robot_type, control_mode)` request shape and the old `dataset_repo_id` request shape resolving to the same head when the dataset is tagged.

CPU suite (`pytest -m "not gpu" -n auto`) must pass; no GPU tests should be required for this change since it's pure metadata plumbing.

## Migration / back-compat

- Old checkpoints with only `dataset_names`: loadable; their heads stay one-per-dataset (effectively all keyed by name as the fallback path).
- New checkpoints: persist `norm_keys` (list of `[robot_type, control_mode]` or `[dataset_name]` for fallback rows).
- No silent behavior changes — log the key map at policy build time so the operator can audit which datasets ended up sharing heads.

## Out of scope

- Changing what fields count as "normalizable" (still `observation.state` / `action` / discrete actions per existing policy configs).
- Hierarchical / partial sharing (e.g. share `state` stats by robot but `action` stats by control mode separately). Could be a follow-up.
- Re-keying image normalization, which is global (ImageNet stats) and not per-dataset today.

## References

Key files surveyed:
- `src/opentau/policies/normalize.py:114-296` — `Normalize` / `Unnormalize` stacked buffer logic
- `src/opentau/policies/pretrained.py:346-420` — `_resolve_dataset_index`
- `src/opentau/policies/factory.py:218-255` — per-dataset stats list construction in `make_policy`
- `src/opentau/datasets/dataset_mixture.py:78-127` — `DatasetMixtureMetadata`, `_TaggedDataset`
- `src/opentau/datasets/lerobot_dataset.py:496-503` — `robot_type` / `control_mode` properties
- `src/opentau/datasets/factory.py:97-112, 308-343` — metadata overrides + validation
- `src/opentau/datasets/compute_stats.py:82-100` — stat sampling/aggregation
- `src/opentau/scripts/eval.py:172-174` — inference-time key injection
- `tests/policies/test_normalize_per_dataset.py:40-170` — current per-dataset normalization tests

shuheng-liu:
## What this does

Switches Normalize / Unnormalize from one stat head per training dataset
to one stat head per unique `(robot_type, control_mode)`. Datasets sharing
both tags share a single head whose stats are sample-count-pooled via
`opentau.datasets.compute_stats.aggregate_stats`. Datasets missing either
tag (None / `""` / whitespace / `"unknown"` case-insensitive on either
field) fall back to per-dataset heads with a single aggregated warning.
Closes #346.

Key invariants preserved:

- `Normalize` / `Unnormalize` semantics unchanged — they remain agnostic
  to what each of their `D` rows means.
- The training path keeps using `batch["dataset_index"]` injected by
  `_TaggedDataset` (the value now indexes the norm-head axis, but the
  wire-level field stays). `robot_type` / `control_mode` from
  `_emit_optional_keys` are subject to dropout and would silently route
  random samples to wrong heads, so they are explicitly NOT consulted at
  training time.
- Backward compatibility: old checkpoints that lack
  `config.dataset_to_norm_index` are loaded as legacy (each
  `dataset_names` entry is its own row, identity mapping).

New behaviour:

- `DatasetMixtureMetadata` exposes `per_norm_key_stats`, `norm_keys`,
  `norm_key_to_index`, `dataset_to_norm_index`, `norm_key_to_dataset_names`.
- Hard fail at mixture construction when two datasets sharing a
  non-fallback norm key have incompatible raw state / action widths
  (the dimensions the buffer would have to pool across).
- An `INFO` summary `Norm-head aggregation: K heads over N datasets` is
  logged, listing each head and its member dataset repo ids — operators
  can audit head assignments from the train log without a sidecar file.
- The `dataset_to_norm_index` map is persisted in `config.json` via the
  existing draccus round-trip; no new files in the checkpoint layout.
- Eval and gRPC accept new optional `robot_type` / `control_mode` config
  fields and inject them into the inference batch. The
  `(robot_type, control_mode)` pair takes precedence over
  `dataset_repo_id` when both are set, and `_resolve_dataset_index`
  enforces the same precedence at the policy layer so callers that
  hand-build a batch see the same behaviour.

## How it was tested

- Added `tests/datasets/test_norm_key_aggregation.py` covering:
  `compute_norm_key` rules (incl. case-insensitive `"Unknown"`, the
  `robot_type == "unknown"` fallback), head sharing, distinct-control-mode
  separation, the fallback warning, the closed-form pooled mean /
  variance correctness for both 2- and 3-dataset pooling with mismatched
  counts, the per-head INFO summary, and the dim-mismatch hard fail.
- Added `TestResolveDatasetIndex` to `tests/policies/test_normalize_per_dataset.py`
  covering all dispatch routes (`dataset_index`, `dataset_repo_id`,
  `(robot_type, control_mode)`, both inference keys present
  simultaneously, the legacy-config path, and the multi-row
  unidentified-batch error).
- Full CPU sweep across `tests/policies/`, `tests/datasets/`,
  `tests/configs/`, `tests/scripts/` (591 passed, 6 skipped — the
  PaliGemma / Gemma3 tests that hit gated HF repos in this sandbox are
  environment-side failures unrelated to this PR).
- GPU tests (`pytest -m "gpu"`) and the model-parallelism regression run
  were dispatched against this branch and both passed.

```bash
git fetch && git checkout claude/robot-control-mode-normalization-2hwwP

# Aggregation + dispatch tests
pytest -sx tests/datasets/test_norm_key_aggregation.py tests/policies/test_normalize_per_dataset.py

# Broader CPU sweep
pytest -m "not gpu" -n auto tests/policies tests/datasets tests/configs tests/scripts
```

Sample train log under the new code (two repos with the same
`(franka_arm, joint_position)` plus one repo of `(ur5, ee)`):

```
INFO  Norm-head aggregation: 2 heads over 3 datasets.
  - franka_arm::joint_position: ['lab/franka_kitchen#0', 'lab/franka_kitchen#1']
  - ur5::ee: ['lab/ur5_cobot']
```

A mixture with two datasets sharing `(franka_arm, joint_position)` but
mismatched raw state widths raises at construction time:

```
ValueError: Datasets sharing norm key 'franka_arm::joint_position' have
incompatible raw (pre-padding) state/action dims, so their normalization
stats cannot be pooled. Override `DatasetConfig.robot_type` or
`DatasetConfig.control_mode` to split them into distinct heads.
Offenders: ['lab/a (state_dim=6, action_dim=7)', 'lab/b (state_dim=8, action_dim=7)']
```

## Checklist

### DIFF
--- a/src/opentau/configs/default.py
+++ b/src/opentau/configs/default.py
@@ -510,15 +510,24 @@ class EvalConfig:
 
     recording_root: str | None = None
 
-    # Which training-time dataset's normalization stats to use when calling
-    # `policy.select_action` on eval observations. ``None`` (default) falls
-    # through to the policy's `_resolve_dataset_index` single-dataset fallback
-    # (works for any policy trained on exactly one dataset). Set this to one
-    # of the strings in `policy.config.dataset_names` when running eval
-    # against a multi-dataset checkpoint, otherwise the inference call will
-    # raise a `KeyError`. Plumbed into each `select_action` call by
+    # Which training-time norm head to use when calling `policy.select_action`
+    # on eval observations. Either:
+    #   - set both `robot_type` and `control_mode` to address the head by
+    #     `(robot_type, control_mode)` (preferred for multi-head checkpoints
+    #     trained against the new per-`(robot_type, control_mode)`
+    #     aggregation),
+    #   - or set `dataset_repo_id` to a training-time dataset name (the
+    #     policy maps it to the norm head via its persisted
+    #     `dataset_to_norm_index`; back-compat path that also works on
+    #     legacy per-dataset checkpoints).
+    # When all three are ``None`` (default), single-head policies fall back
+    # to the `_resolve_dataset_index` zero-default; multi-head ones raise.
+    # The robot_type / control_mode pair takes precedence over
+    # `dataset_repo_id` when both are set. Plumbed by
     # `scripts/eval.py::rollout`.
     dataset_repo_id: str | None = None
+    robot_type: str | None = None
+    control_mode: str | None = None
 
     def __post_init__(self):
         """Validate evaluation configuration."""
--- a/src/opentau/configs/deployment.py
+++ b/src/opentau/configs/deployment.py
@@ -50,13 +50,19 @@ class ServerConfig:
     max_workers: int = 4
     max_send_message_length_mb: int = 100
     max_receive_message_length_mb: int = 100
-    # Which training-time dataset's normalization stats to use for inference
-    # requests. ``None`` (default) falls through to the policy's
-    # `_resolve_dataset_index` single-dataset fallback (works for any
-    # checkpoint trained on exactly one dataset). Set this to one of the
-    # strings in `policy.config.dataset_names` when serving a multi-dataset
-    # checkpoint, otherwise the inference call will raise `KeyError`.
+    # Which training-time norm head to use for inference requests. Either:
+    #   - set both `robot_type` and `control_mode` to address the head by
+    #     `(robot_type, control_mode)` (preferred for multi-head checkpoints),
+    #   - or set `dataset_repo_id` to a training-time dataset name (the
+    #     policy maps it via its persisted `dataset_to_norm_index`;
+    #     back-compat path that also works on legacy per-dataset checkpoints).
+    # When all three are ``None`` (default), single-head policies fall back
+    # to the `_resolve_dataset_index` zero-default; multi-head ones raise.
+    # The robot_type / control_mode pair takes precedence over
+    # `dataset_repo_id` when both are set.
     dataset_repo_id: str | None = None
+    robot_type: str | None = None
+    control_mode: str | None = None
 
     def __post_init__(self):
         """Validate server configuration parameters."""
--- a/src/opentau/configs/policies.py
+++ b/src/opentau/configs/policies.py
@@ -315,12 +315,24 @@ class PreTrainedConfig(draccus.ChoiceRegistry, HubMixin, abc.ABC):
     # repopulated; otherwise the inf-init assertion fires at first forward.
     save_normalization_stats: bool = True
 
-    # Ordered list of dataset names this policy was trained on. Used by the per-sample
-    # Normalize/Unnormalize indexing path to map an inference-time
-    # `batch["dataset_repo_id"]` (str) into the leading dim of the stacked stats
-    # buffers. `None` only for policies constructed outside the standard
-    # `make_policy(ds_meta=...)` path (e.g. legacy single-stats fallbacks).
+    # Ordered list of norm-head identifiers this policy was trained on — one
+    # per row of the stacked Normalize/Unnormalize buffers. With
+    # `dataset_to_norm_index` set (new checkpoints), each entry is a norm key
+    # ("<robot_type>::<control_mode>" or a fallback dataset name). Legacy
+    # checkpoints (`dataset_to_norm_index is None`) carry one entry per
+    # training dataset and the entries are dataset names; the per-dataset and
+    # per-norm-head axes are then the same. `None` only for policies
+    # constructed outside the standard `make_policy(ds_meta=...)` path (e.g.
+    # legacy single-stats fallbacks).
     dataset_names: list[str] | None = None
+
+    # `{training_dataset_name: norm_head_row}` map persisted at training
+    # time. Lets inference resolve a `batch["dataset_repo_id"]` (a
+    # training-time dataset name) to the right row when datasets share a
+    # `(robot_type, control_mode)` head. `None` on legacy checkpoints; the
+    # policy then falls back to treating `dataset_names` as a 1:1
+    # per-dataset list (identity mapping) for inference-time lookup.
+    dataset_to_norm_index: dict[str, int] | None = None
 
     # Deprecated: latency fields are no longer used. Kept for backward-compatible
     # loading of old JSON configs. Must remain 0.0; non-zero values will raise.
--- a/src/opentau/datasets/dataset_mixture.py
+++ b/src/opentau/datasets/dataset_mixture.py
@@ -81,7 +81,7 @@ from opentau.datasets.standard_data_format_mapping import DATA_FEATURES_NAME_MAP
 
 
 class _TaggedDataset(Dataset):
-    """Wraps a ``BaseDataset`` so every sample carries its mixture-level identity.
+    """Wraps a ``BaseDataset`` so every sample carries its norm-head identity.
 
     The wrapped sample dict gains two keys:
 
@@ -89,12 +89,20 @@ class _TaggedDataset(Dataset):
         (matches an entry in ``DatasetMixtureMetadata.dataset_names``).
         Default PyTorch collate batches a per-sample ``str`` into a
         ``list[str]`` of length B.
-      - ``"dataset_index"``: ``torch.long`` scalar — the position of this
-        dataset in ``DatasetMixtureMetadata.dataset_names``. Default collate
-        stacks into a ``(B,)`` long tensor.
+      - ``"dataset_index"``: ``torch.long`` scalar — the policy's
+        **norm-head row** for this dataset (i.e.
+        ``DatasetMixtureMetadata.dataset_to_norm_index[dataset_repo_id]``).
+        Default collate stacks into a ``(B,)`` long tensor. Note: the field
+        name is retained for back-compat, but the value indexes the
+        norm-head axis, not the per-dataset enumerate axis — two datasets
+        sharing a ``(robot_type, control_mode)`` get the same value here.
 
     Policies read either key via ``PreTrainedPolicy._resolve_dataset_index``
     and route per-sample into the stacked Normalize/Unnormalize buffers.
+    The training path always uses ``dataset_index`` (immune to the
+    optional-key dropout that ``robot_type`` / ``control_mode`` undergo);
+    inference can additionally supply ``dataset_repo_id`` or
+    ``(robot_type, control_mode)``.
 
     The wrapper exposes the underlying dataset's ``.meta`` so callers of
     ``WeightedDatasetMixture`` (e.g. metadata validation, per-dataset val
@@ -147,6 +155,46 @@ def pad_vector(vector: np.ndarray, new_dim: int) -> np.ndarray:
     return new_vector
 
 
+# Values that mean "no usable label" when resolving the
+# (robot_type, control_mode) norm key:
+#   - None: field absent from info.json.
+#   - "" / whitespace-only: explicit clear via `DatasetConfig.{robot_type,control_mode}=""`.
+#   - "unknown": the `LeRobotDatasetMetadata.control_mode` default for
+#     missing `info.json` field — not an explicit user-supplied tag.
+_NORM_KEY_MISSING_CONTROL_MODES: frozenset[str] = frozenset({"unknown"})
+
+
+def compute_norm_key(
+    robot_type: str | None, control_mode: str | None, fallback_name: str
+) -> tuple[str, bool]:
+    """Compute the normalization-head key for a dataset.
+
+    Datasets that share the same `(robot_type, control_mode)` are
+    expected to share normalization stats because they share the units,
+    axis count, and physical ranges of the proprio / action vectors.
+    When either tag is missing, fall back to keying by the dataset's own
+    name so the dataset still gets a head — it just won't share one with
+    anything else.
+
+    Args:
+        robot_type: From `meta.info["robot_type"]` (after overrides).
+        control_mode: From `meta.info["control_mode"]` (after overrides).
+        fallback_name: The dataset's deduplicated mixture-level name,
+            used as the head key when fallback fires.
+
+    Returns:
+        A `(norm_key, fallback_fired)` pair. `norm_key` is
+        `"<robot_type>::<control_mode>"` on the happy path and
+        `fallback_name` otherwise; `fallback_fired` is True iff the
+        fallback path was taken.
+    """
+    rt = (robot_type or "").strip()
+    cm = (control_mode or "").strip()
+    if not rt or not cm or cm in _NORM_KEY_MISSING_CONTROL_MODES:
+        return fallback_name, True
+    return f"{rt}::{cm}", False
+
+
 def _apply_data_feature_name_mapping_overrides(
     worker_id: int, mapping_overrides: dict[str, dict[str, str]]
 ) -> None:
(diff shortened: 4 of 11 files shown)


shuheng-liu (commented):
## Code review

Solid, well-tested refactor. Mixture-construction-time pooling and the cleanly-separated training (`dataset_index` only) vs. inference (string lookup) paths look correct, and the dim-mismatch hard fail with offender list is the right shape. Below: one real inconsistency, a few smaller polish items, and a couple of test gaps.

CI: pre-commit + CPU green; GPU runner still in progress at review time.

---

### Findings

**1. Documented precedence contradicts library-level dispatch.** (1 real bug)

Both new docstrings — `configs/default.py` (`EvalConfig`) and `configs/deployment.py` (`ServerConfig`) — claim:

> The robot_type / control_mode pair takes precedence over `dataset_repo_id` when both are set.

And `_resolve_dataset_index`'s own docstring (`policies/pretrained.py`) lists the same dispatch order: training → repo id → (robot, control) → single-row fallback. But the actual ordering in the function body is:

```python
if "dataset_repo_id" in batch:
    ...
    return torch.tensor(indices, ...)

if "robot_type" in batch and "control_mode" in batch:
    ...
```

So with both keys in the batch, `dataset_repo_id` wins — opposite of what the config docstrings claim.

In practice this is masked by `eval.py` / `grpc/server.py`, which both gate with `if (robot_type and control_mode): ... elif dataset_repo_id: ...` so only one key is ever injected. But anyone calling `policy.select_action` directly (including tests, the gRPC client side, or future scripts) hits the documented-vs-actual mismatch.

Pick one direction:
- **Preferred:** reorder `_resolve_dataset_index` to check `(robot_type, control_mode)` first, so the documented invariant holds at every layer.
- Or: weaken the docstrings to "the script only forwards `(robot_type, control_mode)` when both are set; if you bypass the script, `dataset_repo_id` is checked first."

Either way, add a `TestResolveDatasetIndex` case where both keys are present — currently nothing covers this, which is why the inconsistency slipped through.

**2. `_resolve_dataset_index` docstring contradicts itself.**

The docstring lists "Inference by repo id" as branch 2 and "Inference by (robot_type, control_mode)" as branch 3, which matches the implementation. The PR description and the `EvalConfig`/`ServerConfig` docstrings put `(robot_type, control_mode)` first. After resolving (1), reconcile the three so they agree.

**3. `_NORM_KEY_MISSING_CONTROL_MODES` — confirm coverage.**

```python
_NORM_KEY_MISSING_CONTROL_MODES: frozenset[str] = frozenset({"unknown"})
```

The comment correctly notes this is the `LeRobotDatasetMetadata.control_mode` default. Two follow-ups worth considering:
- `robot_type` doesn't have a parallel sentinel set — does LeRobot ever emit `robot_type="unknown"`? If so, that would silently form a `unknown::<mode>` head, which is probably not what users want. Worth either documenting that this is intentional or adding a `_NORM_KEY_MISSING_ROBOT_TYPES` set.
- The strip + lower-case check is one-sided (`cm in _NORM_KEY_MISSING_CONTROL_MODES`). If someone supplies `"Unknown"` or `"UNKNOWN"`, it currently forms a head. Either lower-case before checking, or document the case sensitivity.

**4. Test coverage gaps.**

- No test for the "both `dataset_repo_id` and `(robot_type, control_mode)` present" case — see (1).
- `test_pooled_stats_closed_form` only uses 2 datasets with `(n1, n2) = (100, 300)`. A 3+ dataset pooling test with mismatched weights would more convincingly exercise `aggregate_stats`'s weighted path. Cheap to add and would catch any off-by-one in the weights wiring.
- No roundtrip test for: build mixture → train (or fake) → save → load → resolve a `dataset_repo_id` that the new policy hasn't seen as `dataset_index`. The path is exercised by separate unit tests but never end-to-end.

**5. Value policy single-row restriction is opaque.** `policies/value/modeling_value.py`:

```python
persisted = getattr(self.config, "dataset_to_norm_index", None) or {}
self._dataset_to_norm_index = {k: v for k, v in persisted.items() if v == 0} or {kept[0]: 0}
```

The `if v == 0` filter implicitly assumes the surviving row is row 0 (true for the truncate-to-first-row case here), and the `or {kept[0]: 0}` fallback handles two unrelated cases (empty persisted map, or no entries map to 0). Works, but a reader has to think. Split the fallback or add a one-line comment explaining the two cases.

---

### Distributed safety

No new rank-divergent branching in any policy `forward`. All aggregation is offline at `DatasetMixtureMetadata.__init__`, and `_resolve_dataset_index` returns a deterministic tensor from per-sample identifiers. Clean wrt CLAUDE.md rule #5.

### Style

No new `.view` / `.permute` chains introduced — nothing for the einops rule to apply to.

### Backward compatibility

The identity mapping for legacy checkpoints (`pretrained.py`) is the right shape:

```python
persisted = getattr(config, "dataset_to_norm_index", None)
if persisted:
    self._dataset_to_norm_index = dict(persisted)
else:
    self._dataset_to_norm_index = {n: i for i, n in enumerate(names)}
```

`TestResolveDatasetIndex::test_legacy_config_without_dataset_to_norm_index` covers it. Good.

---

### Summary

Approve-blocking on (1) — the docstring/code mismatch should be resolved one way or the other before merge, and a test added. Everything else is nice-to-have polish.


---

shuheng-liu:
Addressed the review in d8602f9:

1. **Documented precedence vs. actual dispatch** — reorders `_resolve_dataset_index` to check `(robot_type, control_mode)` before `dataset_repo_id`. The eval / gRPC scripts already document this precedence and only inject one or the other in practice, but now the invariant holds at the policy layer too. Added `test_robot_type_control_mode_takes_precedence_over_dataset_repo_id` exercising both keys present simultaneously.
2. **`_resolve_dataset_index` docstring** — reconciled with the new order in the same commit; PR body precedence note also matches.
3. **`_NORM_KEY_MISSING_*` coverage** — replaced `_NORM_KEY_MISSING_CONTROL_MODES` with a shared `_NORM_KEY_MISSING_VALUES` applied to both fields (so `robot_type="unknown"` no longer silently anchors an `unknown::<mode>` head), and made the check case-insensitive via `casefold()`. Added `test_robot_type_unknown_falls_back` and `test_unknown_is_case_insensitive`.
4. **Test gaps** — added the "both keys present" test (above) and `test_pooled_stats_closed_form_three_datasets` with mismatched counts (n = 100/250/650). End-to-end save→load→inference roundtrip skipped: each layer is unit-tested and the round-trip would require instantiating a real policy and serializing weights, which is more invasive than the value-add justifies — happy to add if you'd prefer.
5. **Value policy filter+fallback** — split into named intermediates and added the two-cases comment.

GPU runs both passed (#26552045738, #26552051969); policy + GPU boxes now checked.

---

shuheng-liu (commented):
## Re-review of `d8602f9d`

All five findings from the previous review are addressed correctly. Verified each against the new diff:

| # | Finding | Status | Verification |
|---|---|---|---|
| 1 | Precedence mismatch | ✓ Fixed | `_resolve_dataset_index` now checks `(robot_type, control_mode)` before `dataset_repo_id`; docstring updated to match. |
| 2 | Docstring self-contradiction | ✓ Fixed | `_resolve_dataset_index` docstring branches are now `1) dataset_index → 2) (robot, control) → 3) repo_id → 4) fallback`, consistent with `EvalConfig`/`ServerConfig`. The error message also lists all three identifiers. |
| 3 | `_NORM_KEY_MISSING_*` coverage | ✓ Fixed | Renamed to `_NORM_KEY_MISSING_VALUES` (apt) and applied symmetrically: `rt.casefold() in _NORM_KEY_MISSING_VALUES or cm.casefold() in _NORM_KEY_MISSING_VALUES`. Since inference goes through `compute_norm_key` too, the case-insensitive treatment is automatic on the lookup side. |
| 4 | Test coverage gaps | ✓ Fixed | New `test_robot_type_control_mode_takes_precedence_over_dataset_repo_id` directly exercises the dual-key case; new 3-dataset pooled-stats test uses mismatched counts `(50, 50, 100)`. |
| 5 | Value policy opaque logic | ✓ Fixed | Now: `surviving = {k: v for k, v in persisted.items() if v == 0}` followed by `self._dataset_to_norm_index = surviving or {kept[0]: 0}`, with an inline comment naming the two fallback cases. |

### Additional checks

- **Fallthrough behavior**: when only one of `robot_type`/`control_mode` is in the batch, the `and` guard correctly skips the (robot, control) branch and falls through to `dataset_repo_id` / single-row fallback. Good.
- **Inference-time case-insensitivity**: `_resolve_dataset_index` calls `compute_norm_key(rt, cm, "")` per sample, so a batch with `robot_type="UNKNOWN"` will correctly produce the empty fallback key and raise the clear `ValueError` (`"norm key '' ... not in this policy's training set ..."`) rather than silently routing.
- **Empty-fallback contract**: the new code's choice to pass `""` as the fallback name to `compute_norm_key` is the right one — it ensures a dropped/sentinel-valued pair never silently picks a real head.
- **No collateral regressions** in the rename `_dataset_name_to_index` → `_norm_key_to_index` (CPU tests pass).

### One nit, optional

The two `RuntimeError` branches in `_resolve_dataset_index` ("Policy was loaded without `dataset_names`") have identical messages, but only the `dataset_repo_id` one mentions `dataset_repo_id` in the remediation. The new `(robot, control)` branch's message could surface `make_policy(cfg, ds_meta=...)` as the primary fix since "pass `batch['dataset_index']` directly" doesn't help a user who has structured embodiment tags but no integer index. Not blocking.

### CI

Pre-commit ✓, GPU tests + model-parallelism regression ✓ (per PR description). New CPU run still in progress at review time.

LGTM once CPU goes green.


---

PR #348, opened 2026-05-28
feat(scripts): per-(robot_type, control_mode) normalization in fit_fast_tokenizer

shuheng-liu:
## What this does

Adds `--per-head-norm` (default on) to `src/opentau/scripts/fit_fast_tokenizer.py` so each chunk is min-max-normalized using its own `(robot_type, control_mode)` head's pooled stats — the same stats `Normalize({"ACTION": NormalizationMode.MIN_MAX})` applies via stacked buffers at training time after PR #347.

Before this change, the manual sampler normalized every chunk with `mixture.meta.aggregated_action_stats()` (a single global min/max). Once per-head normalization landed in the policy, that mismatch caused the BPE fit to see chunks centered tighter than the policy's chunks, so a fit with global normalization produced 2-3× longer token sequences at inference than the fit-time roundtrip suggested — hitting `discrete_action_max_length` truncation on a meaningful fraction of training samples.

The new code path computes pooled (per-head) min/max with the same `nanmin`/`nanmax` semantics `DatasetMixtureMetadata._build_norm_heads` uses for the `min`/`max` fields, then normalizes each dataset's chunk slice independently. Datasets sharing a `(robot_type, control_mode)` get identical pooled stats by construction; fallback keys stay singletons; stats-load failures fall back to `[-1, 1]` (same sentinel the global path already used).

Notes:
- `--no-per-head-norm` keeps the legacy global path for callers who need to reproduce a pre-PR-#347 fit.
- `--use-mixture-dataloader` with default `--per-head-norm` now raises with a clear message instead of silently using global stats; threading per-sample `dataset_index` through the dataloader-drained chunks is a separate change.
- The fit report now records `"normalization"` (`per_robot_type_control_mode` or `global`), `per_dataset_norm_keys`, and `n_norm_heads` for traceability.

## How it was tested

Unit tests (added in this PR; all CPU, runs in 3s):

```bash
pytest -sx tests/scripts/test_fit_fast_tokenizer.py
```

- `TestPoolPerHeadStats` — 4 cases covering pooling across a shared head, distinct heads kept separate, stats-load-failure fallback to `[-1, 1]`, and all-NaN dim filling.
- `TestNormalizeChunksPerHead` — 2 cases pinning the per-slice normalization contract and confirming bit-equality with the global path when all datasets share a single head.

End-to-end verified on the production pretrain-pi07-10percent mixture (393 datasets, 32 distinct norm heads, 278k chunks): the per-head fit ran in ~95s on a single CPU node, roundtrip MSE = 0.000083, and a 25k held-out token-length distribution gave mean=53.3 / p99=138 / max=238 — vs. the global-normalized fit's mean=50.1 / p99=65 / max=101 over the same chunks. The original mismatch (token lengths blowing past `discrete_action_max_length=120` at training time) confirmed the global-fit analysis was systematically under-counting.

`pre-commit run --all-files` clean (ruff, ruff-format, pyupgrade, bandit, etc.).

```bash
gh pr checkout feat/fit-fast-tokenizer-per-head-norm
pytest -sx tests/scripts/test_fit_fast_tokenizer.py

# Re-fit any mixture with per-head normalization (now the default):
python -m opentau.scripts.fit_fast_tokenizer \
    --mixture-json /path/to/mixture.json \
    --out-dir /path/to/out_dir/ \
    --chunk-size 30 \
    --action-dim 32 \
    --max-state-dim 118 \
    --total-chunks 278000 \
    --num-workers 64 \
    --seed 0

# Reproduce the legacy single-head behavior:
python -m opentau.scripts.fit_fast_tokenizer ... --no-per-head-norm
```

## Checklist

### DIFF
--- a/src/opentau/scripts/fit_fast_tokenizer.py
+++ b/src/opentau/scripts/fit_fast_tokenizer.py
@@ -512,7 +512,11 @@ def _aggregate_stats_per_head(
     n = len(mixture_cfg.datasets)
     raw_min: list[np.ndarray | None] = [None] * n
     raw_max: list[np.ndarray | None] = [None] * n
-    per_ds_info: list[dict[str, Any]] = [{}] * n
+    # Use list comprehensions (not `[{}] * n`) for mutable defaults so a future
+    # `per_ds_info[idx]["override_X"] = ...` doesn't silently fan out across
+    # all slots. The `[None] * n` / `["<no-repo-id>"] * n` patterns above are
+    # safe because their element types are immutable.
+    per_ds_info: list[dict[str, Any]] = [{} for _ in range(n)]
     per_ds_native_dim: list[int | None] = [None] * n
     per_ds_repo: list[str] = ["<no-repo-id>"] * n
     failures: list[tuple[int, str, str]] = []
@@ -608,6 +612,31 @@ def _aggregate_stats_per_head(
             shown,
             suffix,
         )
+        # Tighter signal for the specific divergence the deferred fallback-name
+        # dedup finding flagged: when a fallback-keyed repo_id appears more
+        # than once in the mixture, fit-time pools them under one shared key
+        # (since `compute_norm_key` returns the same string both times) while
+        # training keeps them as separate singleton heads (since
+        # `_make_dataset_names` deduplicates with `#N` suffixes). Flag the
+        # divergence explicitly so the operator can fix it before it ships.
+        from collections import Counter
+
+        fallback_counts = Counter(fallback_datasets)
+        duplicates_in_fallback = {k: v for k, v in fallback_counts.items() if v > 1}
+        if duplicates_in_fallback:
+            dup_preview = dict(list(duplicates_in_fallback.items())[:10])
+            logger.warning(
+                "%d fallback-keyed repo_id values appear in the mixture more "
+                "than once. Fit-time POOLS these under one shared head per "
+                "repo_id, but training (via `_make_dataset_names`'s `#N` dedup) "
+                "keeps them as separate singleton heads -- the fit-time "
+                "normalization will diverge from training. Set "
+                "`DatasetConfig.robot_type` / `DatasetConfig.control_mode` on "
+                "each entry, or deduplicate the mixture, to close the gap. "
+                "Duplicates (repo_id -> count): %s",
+                len(duplicates_in_fallback),
+                dup_preview,
+            )
 
     # Restore the over-dim warning from the global path: datasets whose
     # native action dim exceeds --action-dim will have high dims silently
--- a/src/opentau/scripts/fit_fast_tokenizer.py
+++ b/src/opentau/scripts/fit_fast_tokenizer.py
@@ -45,15 +45,24 @@ Pipeline (all CPU):
        ``(robot_type, control_mode)`` norm-head stats by default
        (``--per-head-norm``) -- the same stats the training policy
        applies via ``Normalize({"ACTION": NormalizationMode.MIN_MAX})``
-       with per-head stacked buffers (PR #347). Pass
-       ``--no-per-head-norm`` for the legacy global aggregate path,
-       which is what older fits (pre-#347) used; the global path
-       under-spreads each head's distribution and produces shorter
-       fit-time chunks than the policy actually feeds the tokenizer
-       at training, so token-length analysis on the global fit
-       systematically underestimates the truncation rate. The
-       ``--use-mixture-dataloader`` path only supports global
-       normalization for now.
+       with per-head stacked buffers (PR #347). Per-dataset raw stats
+       are zero-padded (matching ``pad_vector`` in
+       ``_to_standard_data_format``) then pooled across head members
+       with ``nanmin``/``nanmax`` (with ``±Inf`` masked first, like
+       ``aggregate_stats``). Failure modes mirror training: any
+       dataset whose stats fail to load -> ``RuntimeError`` (training
+       would crash too via ``_to_standard_data_format``);
+       ``require_non_empty_robot_type/control_mode`` -> same
+       ``ValueError`` as ``datasets.factory._validate_metadata_requirements``.
+       Pass ``--no-per-head-norm`` for the legacy global aggregate
+       path, which is what older fits (pre-#347) used; the global
+       path under-spreads each head's distribution and produces
+       shorter fit-time chunks than the policy actually feeds the
+       tokenizer at training, so token-length analysis on the global
+       fit systematically underestimates the truncation rate.
+       Passing ``--use-mixture-dataloader`` silently degrades to
+       global normalization with a warning (the dataloader path
+       doesn't surface per-sample dataset_index here yet).
     5. Call ``UniversalActionProcessor.fit(...)`` (DCT + Rust BpeTrainer)
        and ``save_pretrained`` the result. The upstream remote-code
        source ``processing_action_tokenizer.py`` is copied alongside so
@@ -223,7 +232,8 @@ def parse_args() -> argparse.Namespace:
             "distribution, so fit-time token-length analysis "
             "underestimates the actual policy-side truncation rate. "
             "Only supported on the default (manual sampler) path; "
-            "--use-mixture-dataloader still uses global normalization."
+            "passing --use-mixture-dataloader silently falls back to "
+            "global normalization with a warning."
         ),
     )
     p.add_argument(
@@ -296,15 +306,29 @@ def _resolve_native_action_key(
     return "action"
 
 
-def _load_dataset_stats(
+def _load_metadata_stats_and_info(
     item: tuple[int, Any],
-) -> tuple[int, str, dict | None, str | None]:
-    """Worker: read action min/max from a dataset's LeRobotDatasetMetadata."""
+) -> tuple[int, str, dict[str, Any], dict[str, np.ndarray] | None, str | None]:
+    """Shared worker: load a dataset's ``LeRobotDatasetMetadata``, return
+    ``(idx, repo_id, info_with_overrides, stats_or_None, err_or_None)``.
+
+    ``info_with_overrides`` is a copy of ``meta.info`` with
+    ``DatasetConfig.{robot_type,control_mode}`` overrides applied -- mirrors
+    ``factory._apply_metadata_overrides`` so the caller can derive the same
+    norm key the training policy does. On full construction failure, the
+    returned ``info`` is empty.
+    """
     idx, cfg = item
+    repo_id = cfg.repo_id or "<no-repo-id>"
     try:
         from opentau.datasets.lerobot_dataset import LeRobotDatasetMetadata
 
         meta = LeRobotDatasetMetadata(cfg.repo_id, root=cfg.root, revision=cfg.revision)
+        info = dict(getattr(meta, "info", {}) or {})
+        if cfg.robot_type is not None:
+            info["robot_type"] = cfg.robot_type
+        if cfg.control_mode is not None:
+            info["control_mode"] = cfg.control_mode
         key = _resolve_native_action_key(
             cfg.repo_id,
             cfg.data_features_name_mapping,
@@ -313,14 +337,16 @@ def _load_dataset_stats(
         if not meta.stats or key not in meta.stats:
             return (
                 idx,
-                cfg.repo_id,
+                repo_id,
+                info,
                 None,
                 f"key {key!r} missing from stats (keys={sorted(meta.stats or [])})",
             )
         s = meta.stats[key]
         return (
             idx,
-            cfg.repo_id,
+            repo_id,
+            info,
             {
                 "min": np.asarray(s["min"], dtype=np.float64).ravel(),
                 "max": np.asarray(s["max"], dtype=np.float64).ravel(),
@@ -328,7 +354,15 @@ def _load_dataset_stats(
             None,
         )
     except Exception as e:  # noqa: BLE001
-        return idx, cfg.repo_id or "<no-repo-id>", None, f"{type(e).__name__}: {e}"
+        return idx, repo_id, {}, None, f"{type(e).__name__}: {e}"
+
+
+def _load_dataset_stats(
+    item: tuple[int, Any],
+) -> tuple[int, str, dict | None, str | None]:
+    """Worker: read action min/max only (used by the legacy global-norm path)."""
+    idx, repo_id, _info, stats, err = _load_metadata_stats_and_info(item)
+    return idx, repo_id, stats, err
 
 
 def _aggregate_stats_manual(
@@ -419,59 +453,9 @@ def _aggregate_stats_manual(
     return action_min, action_max, per_dataset
 
 
-def _load_dataset_info_and_stats(
-    item: tuple[int, Any],
-) -> tuple[int, str, dict[str, Any], dict[str, np.ndarray] | None, str | None]:
-    """Worker for per-head aggregation. Returns
-    ``(idx, repo_id, info_after_overrides, stats_or_None, err_or_None)``.
-
-    Loads the same action stats as ``_load_dataset_stats`` and additionally
-    returns ``meta.info`` with ``DatasetConfig.{robot_type,control_mode}``
-    overrides applied (mirrors ``factory._apply_metadata_overrides``), so the
-    caller can derive the per-dataset norm key the same way the training
-    policy does.
-    """
-    idx, cfg = item
-    try:
-        from opentau.datasets.lerobot_dataset import LeRobotDatasetMetadata
-
-        meta = LeRobotDatasetMetadata(cfg.repo_id, root=cfg.root, revision=cfg.revision)
-        info = dict(getattr(meta, "info", {}) or {})
-        if cfg.robot_type is not None:
-            info["robot_type"] = cfg.robot_type
-        if cfg.control_mode is not None:
-            info["control_mode"] = cfg.control_mode
-        key = _resolve_native_action_key(
-            cfg.repo_id,
-            cfg.data_features_name_mapping,
-            available_keys=set(meta.stats or []),
-        )
-        if not meta.stats or key not in meta.stats:
-            return (
-                idx,
-                cfg.repo_id,
-                info,
-                None,
-                f"key {key!r} missing from stats (keys={sorted(meta.stats or [])})",
-            )
-        s = meta.stats[key]
-        return (
-            idx,
-            cfg.repo_id,
-            info,
-            {
-                "min": np.asarray(s["min"], dtype=np.float64).ravel(),
-                "max": np.asarray(s["max"], dtype=np.float64).ravel(),
-            },
-            None,
-        )
-    except Exception as e:  # noqa: BLE001
-        return idx, cfg.repo_id, {}, None, f"{type(e).__name__}: {e}"
-
-
 def _aggregate_stats_per_head(
     mixture_cfg: DatasetMixtureConfig, action_dim: int, num_workers: int
-) -> tuple[list[np.ndarray], list[np.ndarray], list[str | None]]:
+) -> tuple[list[np.ndarray], list[np.ndarray], list[str], dict[str, np.ndarray]]:
     """Aggregate per-``(robot_type, control_mode)`` action stats.
 
     Mirrors what ``DatasetMixtureMetadata._build_norm_heads`` does for the
@@ -483,11 +467,26 @@ def _aggregate_stats_per_head(
     (which ``aggregate_stats`` does in dataset_mixture) only matters for
     ``mean``/``std`` and is a no-op for ``min``/``max``.
 
-    Datasets whose ``compute_norm_key`` returns fallback (empty / "unknown"
-    robot_type or control_mode) get their own singleton head, normalized
-    with their own stats -- matching the policy's behaviour for fallback
-    keys. Datasets whose action stats fail to load entirely fall back to
-    ``[-1, 1]`` (same sentinel as the global path); a warning is logged.
+    To match the production ``Normalize`` path exactly, per-dataset stats are
+    zero-padded (not NaN-padded) to ``action_dim`` before pooling -- this
+    mirrors ``pad_vector`` (zero-pad) applied in
+    ``DatasetMixtureMetadata._to_standard_data_format``. Trailing slots
+    therefore pool to ``min=max=0``, which makes the production
+    ``(x - min) / (max - min + EPS) * 2 - 1`` evaluate to ``-1`` for the
+    zero-padded action suffix at both fit and training time.
+
+    Failure modes (matching training-time behaviour):
+
+    - If any dataset's action stats fail to load, raise ``RuntimeError``.
+      Training would also crash on such a dataset (``_to_standard_data_format``
+      raises ``KeyError``), so silently producing a tokenizer that the policy
+      will refuse to consume is sneaky. Drop the offending dataset from the
+      mixture (or use ``--no-per-head-norm`` to fall back to the legacy
+      global path) before retrying.
+    - If ``mixture_cfg.require_non_empty_robot_type`` /
+      ``require_non_empty_control_mode`` is set and any dataset still has
+      an empty value after overrides, raise the same ``ValueError`` that
+      ``datasets.factory._validate_metadata_requirements`` raises.
 
     Args:
         mixture_cfg: Mixture config.
@@ -497,23 +496,26 @@ def _aggregate_stats_per_head(
         num_workers: ProcessPool size for parallel stats loading.
 
     Returns:
-        ``(per_ds_min, per_ds_max, per_ds_key)``:
+        ``(per_ds_min, per_ds_max, per_ds_key, per_head_stats)``:
 
         - ``per_ds_min[i]``/``per_ds_max[i]``: ``(action_dim,)`` float32
           arrays for dataset ``i`` (the i-th entry in
           ``mixture_cfg.datasets``). Datasets sharing a non-fallback
           ``(robot_type, control_mode)`` get identical pooled arrays.
-        - ``per_ds_key[i]``: the norm key for dataset ``i`` (or ``None``
-          if the dataset's stats failed to load).
+        - ``per_ds_key[i]``: the norm key for dataset ``i`` (always a str
+          after the all-stats-required guard above).
+        - ``per_head_stats``: deduplicated head -> ``{"min": ..., "max": ...}``
+          map (diagnostics + report).
     """
     from concurrent.futures import ProcessPoolExecutor, as_completed
 
     n = len(mixture_cfg.datasets)
     raw_min: list[np.ndarray | None] = [None] * n
     raw_max: list[np.ndarray | None] = [None] * n
-    per_ds_key: list[str | None] = [None] * n
-    n_ok = 0
-    n_fail = 0
+    per_ds_info: list[dict[str, Any]] = [{}] * n
+    per_ds_native_dim: list[int | None] = [None] * n
+    per_ds_repo: list[str] = ["<no-repo-id>"] * n
+    failures: list[tuple[int, str, str]] = []
 
     logger.info(
         "Loading per-(rt, cm) action stats for %d datasets (workers=%d)",
@@ -523,38 +525,123 @@ def _aggregate_stats_per_head(
     t0 = time.perf_counter()
     work = list(enumerate(mixture_cfg.datasets))
     with ProcessPoolExecutor(max_workers=num_workers) as ex:
-        futs = {ex.submit(_load_dataset_info_and_stats, item): item[0] for item in work}
+        futs = {ex.submit(_load_metadata_stats_and_info, item): item[0] for item in work}
         for fut in as_completed(futs):
             idx, repo_id, info, stats, err = fut.result()
+            per_ds_repo[idx] = repo_id
+            per_ds_info[idx] = info
             if stats is None:
-                n_fail += 1
-                logger.warning("Dataset %d (%s): no stats -- %s", idx, repo_id, err)
+                failures.append((idx, repo_id, err or "<unknown>"))
                 continue
-            n_ok += 1
             native_dim = int(stats["min"].shape[0])
-            mn = np.full(action_dim, np.nan, dtype=np.float64)
-            mx = np.full(action_dim, np.nan, dtype=np.float64)
+            per_ds_native_dim[idx] = native_dim
+            # Zero-pad to action_dim -- matches `pad_vector` in
+            # `_to_standard_data_format`. Trailing slots get min=max=0, so the
+            # production `(0 - 0) / (EPS) * 2 - 1 = -1` matches our fit-time
+            # output bit-for-bit (in float64) for the padded suffix.
+            mn = np.zeros(action_dim, dtype=np.float64)
+            mx = np.zeros(action_dim, dtype=np.float64)
             clip = min(action_dim, native_dim)
             mn[:clip] = stats["min"][:clip]
             mx[:clip] = stats["max"][:clip]
             raw_min[idx] = mn
             raw_max[idx] = mx
-            key, _ = compute_norm_key(
-                info.get("robot_type"),
-                info.get("control_mode"),
-                mixture_cfg.datasets[idx].repo_id,
-            )
-            per_ds_key[idx] = key
 
-    per_ds_min, per_ds_max, per_head_min = _pool_per_head_stats(raw_min, raw_max, per_ds_key, action_dim)
+    if failures:
+        sample = ", ".join(f"{repo}: {err}" for _i, repo, err in failures[:5])
+        raise RuntimeError(
+            f"Per-head normalization requires all datasets' action stats to load, "
+            f"but {len(failures)}/{n} failed. Training would also crash on these "
+            f"datasets (`_to_standard_data_format` raises on missing stats). "
+            f"Drop them from the mixture or fix their stats. Sample: {sample}"
+        )
+
+    # Match `_validate_metadata_requirements` in datasets.factory: if the
+    # mixture demands non-empty robot_type / control_mode, surface that at
+    # fit time so the operator doesn't burn 90s on a fit that the very next
+    # training launch refuses to start.
+    require_robot = bool(getattr(mixture_cfg, "require_non_empty_robot_type", False))
+    require_control = bool(getattr(mixture_cfg, "require_non_empty_control_mode", False))
+    if require_robot or require_control:
+        bad: list[str] = []
+        for i in range(n):
+            info = per_ds_info[i]
+            if require_robot and not (info.get("robot_type") or "").strip():
+                bad.append(f"{per_ds_repo[i]}: robot_type is empty")
+            if require_control and not (info.get("control_mode") or "").strip():
+                bad.append(f"{per_ds_repo[i]}: control_mode is empty")
+        if bad:
+            raise ValueError(
+                "DatasetMixtureConfig requires non-empty metadata fields, but the "
+                f"following {len(bad)} datasets are missing values after overrides:\n  - "
+                + "\n  - ".join(bad)
+                + "\nSet `DatasetConfig.robot_type` / `DatasetConfig.control_mode` "
+                "on the offending dataset(s) to provide an override."
+            )
+
+    # Derive norm keys (now that all stats loaded). Track fallback-fired
+    # datasets and surface them like `_build_norm_heads` does -- otherwise the
+    # operator silently gets singleton-per-dataset heads instead of pooled
+    # ones, which is almost never what they want.
+    per_ds_key: list[str] = [""] * n
+    fallback_datasets: list[str] = []
+    for i in range(n):
+        info = per_ds_info[i]
+        key, fallback_fired = compute_norm_key(
+            info.get("robot_type"),
+            info.get("control_mode"),
+            per_ds_repo[i],
+        )
+        per_ds_key[i] = key
+        if fallback_fired:
+            fallback_datasets.append(per_ds_repo[i])
+    if fallback_datasets:
+        shown = fallback_datasets[:10]
+        suffix = f", ... and {len(fallback_datasets) - 10} more" if len(fallback_datasets) > 10 else ""
+        logger.warning(
+            "%d/%d datasets lack non-empty robot_type / control_mode and were "
+            "given a per-dataset fallback norm head (one singleton head each). "
+            "Set `DatasetConfig.robot_type` / `DatasetConfig.control_mode` to "
+            "pool them into shared heads. Affected: %s%s",
+            len(fallback_datasets),
+            n,
+            shown,
+            suffix,
+        )
+
+    # Restore the over-dim warning from the global path: datasets whose
+    # native action dim exceeds --action-dim will have high dims silently
+    # dropped. The mixture won't include them at training either if
+    # max_action_dim < native, but the warning is still load-bearing because
+    # the fit produces a tokenizer with no token coverage for the dropped
+    # dims at inference.
+    over_dim = [
+        (i, per_ds_native_dim[i], per_ds_repo[i])
+        for i in range(n)
+        if per_ds_native_dim[i] is not None and per_ds_native_dim[i] > action_dim
+    ]
+    if over_dim:
+        max_over = max(d for _, d, _ in over_dim if d is not None)
+        logger.warning(
+            "%d/%d datasets have native action_dim > --action-dim=%d (max %d). "
+            "Their extra dims will be silently dropped. Confirm this matches "
+            "the production policy's max_action_dim (mismatch => the fitted "
+            "BPE doesn't cover those dims at inference). Sample: %s",
+            len(over_dim),
+            n,
+            action_dim,
+            max_over,
+            [r for _, _, r in over_dim[:5]],
+        )
+
+    per_ds_min, per_ds_max, per_head_stats = _pool_per_head_stats(raw_min, raw_max, per_ds_key, action_dim)
     logger.info(
-        "Per-(rt, cm) stats aggregated in %.1fs: %d heads across %d ok / %d fail datasets.",
+        "Per-(rt, cm) stats aggregated in %.1fs: %d heads across %d datasets.",
         time.perf_counter() - t0,
-        len(per_head_min),
-        n_ok,
-        n_fail,
+        len(per_head_stats),
+        n,
     )
-    return per_ds_min, per_ds_max, per_ds_key
+    return per_ds_min, per_ds_max, per_ds_key, per_head_stats
 
 
 def _pool_per_head_stats(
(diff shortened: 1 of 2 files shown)


src/opentau/scripts/fit_fast_tokenizer.py
```diff
@@ -394,6 +419,227 @@ def _aggregate_stats_manual(
...
+                n_fail += 1
+                logger.warning("Dataset %d (%s): no stats -- %s", idx, repo_id, err)
+                continue
+            n_ok += 1
+            native_dim = int(stats["min"].shape[0])
+            mn = np.full(action_dim, np.nan, dtype=np.float64)
+            mx = np.full(action_dim, np.nan, dtype=np.float64)
+            clip = min(action_dim, native_dim)
+            mn[:clip] = stats["min"][:clip]
+            mx[:clip] = stats["max"][:clip]
```
shuheng-liu: **[HIGH] Padded action-dim stats diverge from training: NaN-pad here vs zero-pad in `_to_standard_data_format`.**

The PR pads raw stats with `np.nan` up to `action_dim`, then `_pool_per_head_stats` (line 598-601) falls back to `[-1, 1]` for all-NaN dims. But `DatasetMixtureMetadata._to_standard_data_format` (dataset_mixture.py:512-520) pads action stats with **zeros** via `pad_vector`, and `aggregate_stats` pools zero-padded arrays → pooled `min == max == 0` for padded slots.

Concrete divergence on a `native_dim=14`, `action_dim=32` dataset, on padded slots 14..31:

- **Training:** `min=0, max=0, span=EPS`. A zero-padded chunk slot → `(0-0)/EPS*2 - 1 = -1`.
- **Fit (this PR):** `min=-1, max=+1, span=2+EPS`. Same zero-padded slot → `(0-(-1))/2*2 - 1 ≈ 0`.

So the BPE corpus encodes padding slots as `~0` but the policy at training-time feeds `-1`. This directly defeats the PR's stated goal ("matches what the policy does at training time after PR #347") for every mixture where `max_action_dim > min(native_dim across head members)` — i.e. essentially every production config.

Fix: either zero-pad raw stats here (mirror `pad_vector`), or — better — call into the training-time path (see the cleanup comment on `_aggregate_stats_per_head`).

---

src/opentau/scripts/fit_fast_tokenizer.py
```diff
@@ -394,6 +419,227 @@ def _aggregate_stats_manual(
...
+            raw_min[idx] = mn
+            raw_max[idx] = mx
+            key, _ = compute_norm_key(
+                info.get("robot_type"),
+                info.get("control_mode"),
+                mixture_cfg.datasets[idx].repo_id,
+            )
+            per_ds_key[idx] = key
```
shuheng-liu: **[HIGH] Stats-fail / chunks-succeed mismatch: real chunks get normalized with the `[-1, 1]` sentinel.**

`_load_dataset_info_and_stats` returns `stats=None` when `meta.stats` is empty or lacks the action key (line 449-456). The `if stats is None: continue` here skips `compute_norm_key`, leaving `per_ds_key[idx] = None`. But `_sample_chunks_for_dataset_manual` reads parquet directly via `pq.read_table(full, columns=[action_col, "timestamp"])` and does **not** consult `meta.stats` — so the SAME dataset can fail stats but produce real chunks.

Concrete failure: a v2.0-format dataset where `episodes_stats.jsonl` has the action key under a slightly different name (`"action"` vs `"actions"` — exactly the case `_resolve_native_action_key`'s `available_keys` fallback was added to handle for parquet but is bypassed here when `meta.stats` already has another non-action key). Stats worker returns `None`; sampler still reads parquet just fine. Then `_pool_per_head_stats` (line 608-613) writes `per_ds_min[idx] = [-1, 1]` sentinel, and `_normalize_chunks_per_head` happily normalizes real ±3.0 action values as if their range were `[-1, 1]`, producing values like `+3.0` in the BPE corpus.

Worse: if peer datasets in the same `(rt, cm)` head loaded fine and would have provided usable pooled stats, those are silently NOT applied to this dataset — because `per_ds_key[idx]` was never set.

Fix: compute `per_ds_key[idx]` unconditionally from `info` even when stats failed, so the dataset can inherit its head's pooled stats from peers. If no peers exist, log loudly before falling back to `[-1, 1]`.

---

src/opentau/scripts/fit_fast_tokenizer.py
```diff
@@ -394,6 +419,227 @@ def _aggregate_stats_manual(
...
+            clip = min(action_dim, native_dim)
+            mn[:clip] = stats["min"][:clip]
+            mx[:clip] = stats["max"][:clip]
+            raw_min[idx] = mn
+            raw_max[idx] = mx
+            key, _ = compute_norm_key(
+                info.get("robot_type"),
+                info.get("control_mode"),
+                mixture_cfg.datasets[idx].repo_id,
+            )
```
shuheng-liu: **[HIGH] `fallback_name = cfg.repo_id` diverges from training, which uses the deduplicated `_make_dataset_names` output.**

`DatasetMixtureMetadata._build_norm_heads` (dataset_mixture.py:340-342) passes `ds_name` to `compute_norm_key`, where `ds_name` comes from `WeightedDatasetMixture._make_dataset_names` (dataset_mixture.py:764) — which appends `#N` suffixes for duplicated `repo_id`s (e.g. `['A', 'B', 'A']` → `['A#0', 'B', 'A#1']`).

When a mixture lists the same `repo_id` twice (legitimate: same source, different `episodes` subsets or different `revision`) AND both lack `robot_type`/`control_mode`:

- **Training:** fallback names are `'A#0'` and `'A#1'` → two distinct singleton heads, each normalized with its own stats.
- **Fit (this PR):** both calls use `fallback_name='A'` → one pooled head, both datasets get the combined `nanmin/nanmax`.

The fit-time per-head normalization no longer matches the training-time per-head normalization — the very invariant the PR description claims to establish. Tests don't catch this because `TestPoolPerHeadStats` only exercises distinct keys, never a pooled-fallback case.

Fix: compute deduplicated names up front (replicate `_make_dataset_names`'s simple `Counter` loop on `cfg.repo_id`/`cfg.vqa`) and pass those to `compute_norm_key`.

---

src/opentau/scripts/fit_fast_tokenizer.py
```diff
@@ -394,6 +419,227 @@ def _aggregate_stats_manual(
...
+
+    per_ds_min, per_ds_max, per_head_min = _pool_per_head_stats(raw_min, raw_max, per_ds_key, action_dim)
+    logger.info(
+        "Per-(rt, cm) stats aggregated in %.1fs: %d heads across %d ok / %d fail datasets.",
+        time.perf_counter() - t0,
+        len(per_head_min),
+        n_ok,
+        n_fail,
+    )
+    return per_ds_min, per_ds_max, per_ds_key
```
shuheng-liu: **[HIGH] Silent success when ALL datasets' stats fail to load — global path raises, per-head path doesn't.**

`_aggregate_stats_manual` raises `RuntimeError("No dataset stats loaded successfully.")` at line 378 if `mins_padded` is empty after the loop. The per-head path has no equivalent guard: if every worker returns `stats=None` (e.g. transient HF Hub failure, wrong `revision`, mass perm error), `per_ds_key` stays all-`None` and `_pool_per_head_stats` returns `per_ds_min[i] = [-1, 1]` for every dataset. `_sample_via_manual` then runs against possibly-cached parquet, the fit proceeds, and you get a perfectly-formed tokenizer trained against a fully-`[-1, 1]`-normalized corpus with zero real stats anywhere.

The operator only sees N warning lines and a misleading `Per-(rt, cm) stats aggregated in 0.0s: 0 heads across 0 ok / N fail datasets.` log — but the script `return 0`'s and writes `fit_report.json` with `n_norm_heads: 0`. A monitoring system that checks exit codes will think the fit succeeded.

Fix: after `_pool_per_head_stats`, raise `RuntimeError` if `n_ok == 0` (or equivalently if `len(per_head_min) == 0`).

---

src/opentau/scripts/fit_fast_tokenizer.py
```diff
@@ -394,6 +419,227 @@ def _aggregate_stats_manual(
...
+        if k is not None and raw_min[i] is not None:
+            key_to_indices[k].append(i)
+
+    per_head_min: dict[str, np.ndarray] = {}
+    per_head_max: dict[str, np.ndarray] = {}
+    for key, indices in key_to_indices.items():
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore", category=RuntimeWarning)
+            mn = np.nanmin(np.stack([raw_min[i] for i in indices]), axis=0)
+            mx = np.nanmax(np.stack([raw_max[i] for i in indices]), axis=0)
```
shuheng-liu: **[MEDIUM] `np.nanmin`/`np.nanmax` propagate `±Inf`; training's `aggregate_stats` masks `Inf→NaN` first.**

`aggregate_stats` in `compute_stats.py:350-355` does:
```python
mins_safe = np.where(np.isfinite(mins), mins, np.nan)
maxs_safe = np.where(np.isfinite(maxs), maxs, np.nan)
agg_min = np.nanmin(mins_safe, axis=0)
agg_max = np.nanmax(maxs_safe, axis=0)
```

so a single `-Inf`/`+Inf` member is excluded from the per-dim pool — finite peers still contribute. This PR does raw `nanmin`/`nanmax`, so `-Inf` poisons the pooled min for the whole head's dim, the `~isfinite` check below catches it, and the entire dim falls back to `[-1, 1]`.

Concrete divergence: a 3-member head where one member has `max=+inf` on dim 7 (legitimate when stats came from an unbounded-velocity LeRobot v2 dataset). Training keeps the finite members' pooled max; fit silently drops the dim to `[-1, 1]` sentinel.

Fix: mirror the `np.where(np.isfinite(...), ..., np.nan)` pre-mask before the nanmin/nanmax — one extra line each.

---

src/opentau/scripts/fit_fast_tokenizer.py
```diff
@@ -394,6 +419,227 @@ def _aggregate_stats_manual(
...
+                "min": np.asarray(s["min"], dtype=np.float64).ravel(),
+                "max": np.asarray(s["max"], dtype=np.float64).ravel(),
+            },
+            None,
+        )
+    except Exception as e:  # noqa: BLE001
+        return idx, cfg.repo_id, {}, None, f"{type(e).__name__}: {e}"
+
+
+def _aggregate_stats_per_head(
```
shuheng-liu, line 456: **[MEDIUM] `require_non_empty_robot_type` / `require_non_empty_control_mode` are silently ignored at fit time.**

`DatasetMixtureConfig` exposes `require_non_empty_robot_type` / `require_non_empty_control_mode` (configs/default.py:322-323), and `factory._validate_metadata_requirements` (factory.py:308-343) raises `ValueError` when any dataset's post-override value is empty.

When a user enables either flag, training refuses to start — but `_aggregate_stats_per_head` skips the check entirely and happily falls back to the singleton-by-`repo_id` path for those datasets. The fit silently produces a BPE corpus normalized differently from any policy that will actually run with that mixture.

Fix: after the worker loop, call (or inline) `_validate_metadata_requirements`'s check against the `info` dicts the workers returned. If `require_non_empty_*` is set in the mixture, raise here with the same operator-friendly message instead of producing a misleading fit.

---

src/opentau/scripts/fit_fast_tokenizer.py
shuheng-liu, line 584: **[MEDIUM] `compute_norm_key`'s `fallback_fired` is discarded; operator never sees the "N datasets lack rt/cm" warning.**

`_build_norm_heads` (dataset_mixture.py:344-345, 405-422) accumulates `fallback_dataset_names` and emits a single aggregated warning capped at 10 names so the operator knows which datasets are getting private singletons instead of pooled heads. The PR uses `key, _ = compute_norm_key(...)` and throws the flag away.

On a 393-dataset mixture with many missing `(rt, cm)` tags, the user reads `Per-(rt, cm) stats aggregated in 95s: 280 heads across 393 ok / 0 fail` and concludes 280 real shared heads — but the actual mix could be 30 real pooled heads + 250 fallback singletons, which is a different normalization regime and a different signal about config hygiene.

Fix: collect `fallback_fired=True` indices, log the same capped aggregate warning training does, and surface the count in `fit_report.json` (e.g. `n_fallback_keyed_datasets`).

---

src/opentau/scripts/fit_fast_tokenizer.py
```diff
@@ -1016,6 +1262,19 @@ def main() -> int:
...
+    # the dataloader-drained chunks would need per-sample dataset_index from
+    # the batch and isn't done yet. Surface the gap loudly rather than silently
+    # falling back.
+    if args.use_mixture_dataloader and args.per_head_norm:
+        raise ValueError(
+            "--per-head-norm is not yet supported on the --use-mixture-dataloader "
+            "path; pass --no-per-head-norm to use the legacy global normalization, "
+            "or drop --use-mixture-dataloader to use the default manual sampler "
+            "(which does support per-head)."
+        )
```
shuheng-liu, line 1430: **[MEDIUM] Default-flip + hard-raise breaks every existing `--use-mixture-dataloader` invocation.**

Pre-PR, `--use-mixture-dataloader` worked alone (used global stats). Post-PR, the same command line raises `ValueError` immediately because `args.per_head_norm` defaults to True. Any CI script, sbatch wrapper, notebook example, or shell-history re-run that previously used `--use-mixture-dataloader` now breaks on first run — even though the user wasn't asking for the new behavior.

Two safer options:
1. When `--use-mixture-dataloader` is explicitly set and `--per-head-norm` was NOT user-supplied, log a warning ("`--per-head-norm` not supported on the dataloader path yet; falling back to global") and proceed.
2. Detect user intent via `argparse`'s seen-on-cli tracking and only raise when *both* are explicit.

The current behavior is "make the user add `--no-per-head-norm` to every existing automation," which is a silent compatibility break for the dataloader path.

---

src/opentau/scripts/fit_fast_tokenizer.py
```diff
@@ -394,6 +419,227 @@ def _aggregate_stats_manual(
...
+                continue
+            n_ok += 1
+            native_dim = int(stats["min"].shape[0])
+            mn = np.full(action_dim, np.nan, dtype=np.float64)
+            mx = np.full(action_dim, np.nan, dtype=np.float64)
+            clip = min(action_dim, native_dim)
+            mn[:clip] = stats["min"][:clip]
+            mx[:clip] = stats["max"][:clip]
+            raw_min[idx] = mn
+            raw_max[idx] = mx
```
shuheng-liu, line 552: **[MEDIUM] Lost over-dim warning: `--action-dim` < native silently truncates without surfacing offenders.**

`_aggregate_stats_manual` (lines 383-400) explicitly logs:
```
N/M datasets have native action_dim > --action-dim=K (max X). Sample: [repo1, ...]
```
so an operator running with the wrong `--action-dim` immediately sees the misconfiguration with a sample of offending repos.

The per-head path silently truncates at line 537 (`clip = min(action_dim, native_dim)`). The only remaining check is the much weaker chunk-level warning at line 1383 — which (a) doesn't name offenders, (b) only fires on datasets that successfully sampled chunks, and (c) is easy to overlook in a 90-second fit log.

Fix: lift the over-dim collection + warning out of `_aggregate_stats_manual` into a shared helper called from both paths.

---

src/opentau/scripts/fit_fast_tokenizer.py
```diff
@@ -394,6 +419,227 @@ def _aggregate_stats_manual(
...
+    out = np.zeros_like(stacked)
+    offset = 0
+    for i, count in enumerate(per_dataset_chunks):
+        if count <= 0:
+            continue
+        out[offset : offset + count] = _normalize_chunks(
+            stacked[offset : offset + count], per_ds_min[i], per_ds_max[i]
+        )
+        offset += count
+    return out
```
shuheng-liu, line 791: **[MEDIUM] Implicit cross-function invariant: silent miscalibration if `_sample_via_manual`'s ordering ever changes.**

`_normalize_chunks_per_head` relies on `stacked` being concatenated in mixture-config order with per-dataset bucket sizes matching `per_dataset_chunks`. That contract lives only in this docstring and `_sample_via_manual`'s comment at line 794-798.

A future change to `_sample_via_manual` — interleaving chunks for BPE corpus diversity, yielding chunks as they arrive, or implementing the long-standing TODO of threading `dataset_index` through the dataloader — would silently misalign per-dataset stats with chunks. No test catches it because `test_per_dataset_slice_normalization` constructs `stacked` manually to satisfy the same contract.

Two fixes that make this safe:
1. Take an explicit `per_chunk_dataset_idx: np.ndarray` (shape `(N,)`) instead of relying on contiguous slices — the same dataset-index tagging `_TaggedDataset` already does at training.
2. At minimum, `assert sum(per_dataset_chunks) == stacked.shape[0]` at the top of the function so a desync raises instead of silently corrupting the BPE corpus.

---

tests/scripts/test_fit_fast_tokenizer.py
```diff
@@ -81,3 +86,164 @@ def test_none_action_freq_matches_explicit_native_fps(self, tmp_path, lerobot_da
...
+        shared_min = np.array([-2.0, -2.0, -2.0, -2.0], dtype=np.float32)
+        shared_max = np.array([2.0, 2.0, 2.0, 2.0], dtype=np.float32)
+
+        out_per_head = _normalize_chunks_per_head(
+            stacked, per_dataset_chunks, [shared_min, shared_min], [shared_max, shared_max]
+        )
+        out_global = _normalize_chunks(stacked, shared_min, shared_max)
+
+        # Bit-exact: both paths apply the same affine transform.
+        np.testing.assert_array_equal(out_per_head, out_global)
```
shuheng-liu, line 267: **[MEDIUM] Test coverage gap: no test pins fit-time output against the production `Normalize` layer.**

The two new test classes exercise only the pure helpers (`_pool_per_head_stats`, `_normalize_chunks_per_head`). The orchestrator (`_aggregate_stats_per_head`) — which does the worker fanout, calls `compute_norm_key`, and assembles `per_ds_key` — is untested. More importantly, the **invariant the PR exists to establish** ("fit-time normalization matches what `Normalize({\"ACTION\": NormalizationMode.MIN_MAX})` produces at training time") is asserted only in docstrings.

A single end-to-end test would catch the NaN-pad vs zero-pad divergence, the fallback-name divergence, and the Inf-handling divergence all at once: build 2-3 in-process mini datasets with explicit `(robot_type, control_mode)` overrides, construct both a `DatasetMixtureMetadata` (training path) and call `_aggregate_stats_per_head` (fit path), then assert per-dataset `(min, max)` arrays are elementwise-equal.

Without this test, future drift between the two paths is silent — exactly the failure mode PR #347 + #348 are meant to prevent.

---

src/opentau/scripts/fit_fast_tokenizer.py
```diff
@@ -1168,10 +1441,15 @@ def main() -> int:
...
         "per_dataset_chunks": per_dataset_chunks,
         "sample_errors": sample_errors,
-        "global_action_min": action_min.tolist(),
-        "global_action_max": action_max.tolist(),
+        "global_action_min": action_min.tolist() if action_min is not None else None,
+        "global_action_max": action_max.tolist() if action_max is not None else None,
+        "per_dataset_norm_keys": per_ds_norm_keys,
+        "n_norm_heads": (
+            len({k for k in per_ds_norm_keys if k is not None}) if per_ds_norm_keys is not None else None
+        ),
```
shuheng-liu: **[LOW] `global_action_min`/`max` silently become JSON `null` on default per-head fits; no per-head replacement in the report.**

Pre-PR these fields were always arrays. On the new default path they're `null`, and `per_dataset_norm_keys` (the only addition) only gives keys — not the actual `(min, max)` arrays used. A downstream tool doing `np.asarray(report['global_action_min'])` crashes with `TypeError` on every new default fit; reproducibility audits comparing two fits' `(min, max)` have nothing to compare against on per-head fits.

Two fixes (pick one):
1. Write `per_head_norm_stats: {key: {"min": [...], "max": [...]}}` to the report so the actual numbers are recoverable.
2. Keep `global_action_min/max` non-null by always computing them as an aggregate (cheap; useful sanity check) regardless of which normalization the fit used.

---

src/opentau/scripts/fit_fast_tokenizer.py
```diff
@@ -394,6 +419,227 @@ def _aggregate_stats_manual(
...
+            cfg.repo_id,
+            info,
+            {
+                "min": np.asarray(s["min"], dtype=np.float64).ravel(),
+                "max": np.asarray(s["max"], dtype=np.float64).ravel(),
+            },
+            None,
+        )
+    except Exception as e:  # noqa: BLE001
+        return idx, cfg.repo_id, {}, None, f"{type(e).__name__}: {e}"
```
shuheng-liu: **[LOW / cleanup] `_load_dataset_info_and_stats` is a near-clone of `_load_dataset_stats` (lines 299-331) that has already drifted.**

The two workers share ~90% of their body (metadata construction, action-key resolution, stats extraction, exception handling). They've already diverged in two places:
- Exception handler returns `cfg.repo_id` (can be `None`) here vs the `cfg.repo_id or "<no-repo-id>"` sentinel at line 331 — so warning lines become `Dataset 3 (None): no stats ...` on the per-head path, harder to grep.
- This worker doesn't populate the `per_dataset` diagnostic dict the original returns.

Refactor: extend `_load_dataset_stats` to optionally return `info` (or always return it — the dict is small), drop `_load_dataset_info_and_stats`. Any future fix to action-key resolution or the repo_id fallback then happens in one place.

---

src/opentau/scripts/fit_fast_tokenizer.py
shuheng-liu, line 456: **[LOW / altitude] Right depth: call into `DatasetMixtureMetadata` instead of re-implementing `_build_norm_heads` here.**

`_aggregate_stats_per_head` + `_pool_per_head_stats` (~140 lines) parallel `DatasetMixtureMetadata._build_norm_heads` (dataset_mixture.py:322): both group datasets by `compute_norm_key`, dedupe heads, and pool per-head min/max. The training side is the source of truth; the fit side is a re-derivation that has already drifted on five separate dimensions surfaced in other comments on this review (NaN-pad vs zero-pad, fallback-name dedup, Inf masking, `require_non_empty_*`, fallback_fired warning).

Right depth: construct a stub `DatasetMixtureMetadata(cfg, metadatas, dataset_weights, dataset_names=_make_dataset_names(...))` from the already-loaded metadata list and read `per_norm_key_stats` + `dataset_to_norm_index` off it. The fit script already builds a stub `TrainPipelineConfig` in `_build_train_cfg` for the `--use-mixture-dataloader` path; the same stub feeds `DatasetMixtureMetadata` directly. Every divergence flagged on this review collapses to zero new code surface.

Also: `aggregated_action_stats()`'s docstring (dataset_mixture.py:444) still names this script as its only consumer — that becomes wrong if the global path is kept, and is already wrong on the default per-head path.

---

shuheng-liu (commented):
Review of #348 — 13 inline findings, ranked by severity.

The PR's stated invariant — "fit-time normalization matches what the policy applies at training time via `Normalize({\"ACTION\": NormalizationMode.MIN_MAX})` with per-head stacked buffers (PR #347)" — is not actually established. The per-head pooling here re-implements `_build_norm_heads` from scratch and diverges on **at least five** dimensions that each silently break the equivalence:

**High severity (silent normalization mismatch):**
1. NaN-padding raw stats here vs `pad_vector` zero-padding in `_to_standard_data_format` — padded action-dim slots normalize to `~0` at fit but `-1` at training. Affects every mixture with `max_action_dim > native_dim` (essentially every production config).
2. Stats-fail / chunks-succeed mismatch — a dataset that errors during stats loading but produces chunks via direct parquet read gets its real chunks normalized with the `[-1, 1]` sentinel, even when peer datasets in the same `(rt, cm)` head loaded fine.
3. Fallback name uses raw `cfg.repo_id`; training uses `_make_dataset_names`'s deduplicated `name#N`. Duplicate-`repo_id` entries pool at fit but stay separate at training.
4. No `RuntimeError` guard when ALL stats fail — script silently produces a tokenizer fit on a fully `[-1, 1]`-sentinel-normalized corpus and `exit 0`.

**Medium severity:**
5. `np.nanmin/nanmax` propagate `±Inf`; `aggregate_stats` masks `Inf→NaN` first.
6. `require_non_empty_robot_type` / `require_non_empty_control_mode` not enforced — fit completes, training refuses to start.
7. `compute_norm_key`'s `fallback_fired` discarded — operator never sees the "N datasets lack rt/cm" warning that training emits.
8. Default-flip + hard-raise breaks every existing `--use-mixture-dataloader` invocation.
9. Dropped over-dim warning — `--action-dim` misconfig is no longer surfaced with offending repo IDs at stats-aggregation time.
10. `_normalize_chunks_per_head` relies on an implicit cross-function ordering invariant — no `assert sum(per_dataset_chunks) == stacked.shape[0]`; a future refactor to `_sample_via_manual` silently corrupts the BPE corpus.

**Low severity / cleanup:**
11. Test coverage gap — no test pins fit-time output against the production `Normalize` layer. A single equivalence test would catch findings 1, 3, 5 at once.
12. `global_action_min`/`max` silently `null` in the report on default per-head fits; no per-head replacement field.
13. `_load_dataset_info_and_stats` near-clone of `_load_dataset_stats` (already drifted on `<no-repo-id>` sentinel + `per_dataset` diagnostic).

**Altitude:** the right depth for this PR is to construct a stub `DatasetMixtureMetadata` from the already-loaded metadata list and read `per_norm_key_stats` off it — every divergence above collapses to zero new code surface. The current re-implementation guarantees future drift between fit and training each time `_build_norm_heads` evolves.

The end-to-end roundtrip described in the PR body (95s fit, MSE = 8.3e-5, mean=53.3 / p99=138 token length) confirms the per-head path produces a sane tokenizer overall, but the specific failure modes above are not exercised by either the new unit tests or the production smoke run (393 datasets with `max_action_dim=32` and varying native dims — finding 1 would silently apply in production today and the BPE token-length distribution still looks reasonable because padded-suffix bytes compress to a single repetitive token regardless).

---

shuheng-liu:
## Response to review (commit 85c535b)

Going through each finding with what I did or why I pushed back. Net: 11 of 13 fixed in this push; 2 deferred with a reason.

### High severity — all fixed

1. **NaN-pad → zero-pad raw stats.** Fixed. `_aggregate_stats_per_head` now `np.zeros((action_dim,))` per dataset before pool, matching `pad_vector` in `_to_standard_data_format`. Padded-suffix slots now pool to `min=max=0` and normalize to `-1` at both fit and training, byte-for-byte (modulo the float64→float32 cast rounding, ~1 ULP). The new `TestNormalizeEquivalenceVsProduction` pins this against the production `Normalize` layer directly.

2. **Stats-fail / chunks-succeed mismatch.** Fixed by subsumption into #4 — `_aggregate_stats_per_head` now raises `RuntimeError` on **any** stats-load failure, with the offending `repo_id: err` pairs in the message. Training crashes on the same dataset via `_to_standard_data_format`'s `KeyError`, so silently producing a fit the policy will refuse is the wrong behavior. The `[-1, 1]` sentinel path for failed datasets is gone; `_pool_per_head_stats` now asserts every row is populated.

3. **Fallback-name dedup (`repo_id` vs `name#N`).** Deferred. Real but narrow: it only diverges when a mixture has *both* duplicate `repo_id` *and* missing `(robot_type, control_mode)` on those rows. The new fallback-datasets warning surfaces the second condition loudly; an operator hitting both at once will see the warning and notice. Replicating `WeightedDatasetMixture._make_dataset_names`'s dedup at fit time is the right fix but pulls in `TrainPipelineConfig`/`BaseDataset` plumbing that doesn't fit this PR's scope. Tracked as a follow-up.

4. **No `RuntimeError` on all-stats-fail.** Fixed via the same change as #2.

### Medium severity — all fixed

5. **Mask `±Inf` → `NaN` before pool.** Fixed. `_pool_per_head_stats` now does `np.where(np.isfinite(...), ..., np.nan)` on the stacked min/max before `nanmin`/`nanmax`, mirroring `compute_stats.py:350`. New `test_inf_masked_before_pool` + `test_all_inf_dim_falls_back_to_minus_one_one` pin both cases.

6. **`require_non_empty_robot_type` / `_control_mode` not enforced.** Fixed. `_aggregate_stats_per_head` reads `mixture_cfg.require_non_empty_*` and raises the same `ValueError` text as `datasets.factory._validate_metadata_requirements` if any row is empty after overrides. Operator no longer wastes a fit on a mixture training will refuse.

7. **`fallback_fired` warning discarded.** Fixed. Tracked per-dataset, then logged in the same shape `_build_norm_heads` uses (capped at 10 names, with `, ... and N more` suffix).

8. **Default-flip + hard-raise.** Fixed. `--use-mixture-dataloader --per-head-norm` now warns and falls back to global, so existing `--use-mixture-dataloader` invocations keep working without a flag change. Help text + module docstring updated.

9. **Over-`--action-dim` warning dropped.** Fixed. Mirrors the old warning from `_aggregate_stats_manual` — logs `N/total datasets have native action_dim > --action-dim=K (max M)` with the first 5 offending repo IDs.

10. **No `assert sum(per_dataset_chunks) == stacked.shape[0]`.** Fixed in `_normalize_chunks_per_head` with an explicit error message; new `test_chunk_count_mismatch_raises_assert` pins it.

### Low severity — cleanup applied

11. **No equivalence test against production `Normalize`.** Fixed. `TestNormalizeEquivalenceVsProduction::test_per_head_normalize_matches_production_normalize` builds a 2-dataset / 2-head batch (with zero-padded trailing slots) and asserts the manual output matches `Normalize({\"ACTION\": MIN_MAX}).forward(...)` to `atol=2e-7`. Caught the original float64-vs-float32 ULP drift; the comment explains why bit-equality isn't claimed.

12. **`global_action_min/max` silently null on per-head fits.** Fixed. Report now has `per_head_action_stats: {key: {min: [...], max: [...]}}` and `n_norm_heads` derived from that. Global fields remain populated only on global fits.

13. **`_load_dataset_info_and_stats` near-clone of `_load_dataset_stats`.** Fixed. Extracted `_load_metadata_stats_and_info` as the shared worker; `_load_dataset_stats` now thin-wraps it.

### Altitude critique — deferred with reason

The "construct a stub `DatasetMixtureMetadata` and read `per_norm_key_stats` off it" path is the right long-term answer and would eliminate every divergence finding above by construction. The blocker is that `DatasetMixtureMetadata.__init__` does much more than build heads (loads parquet, builds episode indices, computes sample weights), and `_build_norm_heads` is currently a method that reads `self.dataset_names` / `self.per_dataset_stats` / `self.cfg.{max_action_dim, num_cams}`. Doing it right means extracting `_build_norm_heads` (and the `_to_standard_data_format` step it depends on) to a free function — a cross-file refactor that touches `dataset_mixture.py` directly. Out of scope for this PR; tracked as a follow-up. The surgical fixes above close every concrete silent-divergence the review identified.

### Impact on the currently-running fit

The published `TensorAuto/fast-mixture-per-head` was built with the pre-fix code, so the NaN-pad vs zero-pad mismatch (finding #1) applied. The empirical impact is small — the BPE token-length tail at training (max 288 observed) exceeded the fit-time prediction (max 238) by ~20%, with truncation rate ~3 / 152K samples = 0.002% on the current run. Not blocking the in-flight job, but a re-fit on the corrected code would tighten the prediction.

src/opentau/scripts/fit_fast_tokenizer.py
shuheng-liu: Fixed in 85c535b. `_aggregate_stats_per_head` now zero-pads raw stats up to `action_dim`, matching `pad_vector` in `_to_standard_data_format`. Padded slots pool to `min=max=0` and the production `(0 - 0) / (0 - 0 + EPS) * 2 - 1` evaluates to `-1` at both fit and training time. `TestNormalizeEquivalenceVsProduction::test_per_head_normalize_matches_production_normalize` pins the equivalence against the production `Normalize({"ACTION": MIN_MAX})` layer with per-head buffers; the test deliberately includes zero-padded trailing slots and asserts both paths land at `-1` there. The atol=2e-7 covers the manual path's float64→float32 cast (~1 ULP) vs production's all-float32 — comment explains the cast trade-off.

src/opentau/scripts/fit_fast_tokenizer.py
shuheng-liu: Fixed in 85c535b, subsumed by #4. `_aggregate_stats_per_head` now raises `RuntimeError` on any stats-load failure, with the offending `repo_id: error` pairs in the message. The "real chunks normalized with `[-1, 1]` sentinel" path is gone; `_pool_per_head_stats` asserts every row is populated before pooling. Behavior now matches training, which also crashes on missing stats via `_to_standard_data_format`'s `KeyError`.

src/opentau/scripts/fit_fast_tokenizer.py
shuheng-liu: Deferred (still relevant, but narrow). The divergence requires a mixture with both duplicate `repo_id` AND missing `(robot_type, control_mode)` on those rows. The new fallback-datasets warning (#7) now surfaces the second condition (capped at 10 names with "... and N more" suffix), so an operator hitting both at once will see signal at fit time. Replicating `WeightedDatasetMixture._make_dataset_names`'s dedup at fit time is the right fix but pulls in `TrainPipelineConfig` / `BaseDataset` plumbing that doesn't belong in this PR. Tracking as a separate follow-up.

src/opentau/scripts/fit_fast_tokenizer.py
shuheng-liu: Fixed in 85c535b. `_aggregate_stats_per_head` now raises `RuntimeError` on **any** stats-load failure (strictly stronger than the global path's "all fail" guard). The message format is `"Per-head normalization requires all datasets' action stats to load, but K/N failed... Sample: <up to 5 repo_id: err pairs>"`. The single-sentinel path that silently emitted a `[-1, 1]` tokenizer is removed.

src/opentau/scripts/fit_fast_tokenizer.py
shuheng-liu: Fixed in 85c535b. `_pool_per_head_stats` now does `np.where(np.isfinite(...), ..., np.nan)` on stacked min/max before `nanmin` / `nanmax`, mirroring `compute_stats.py:350-355` byte-for-byte. New tests pin both cases: `test_inf_masked_before_pool` (single Inf entry in one peer — finite peer's value wins) and `test_all_inf_dim_falls_back_to_minus_one_one` (every member Inf in a dim — falls back to `[-1, 1]`).

src/opentau/scripts/fit_fast_tokenizer.py
shuheng-liu: Fixed in 85c535b. `_aggregate_stats_per_head` reads `mixture_cfg.require_non_empty_robot_type` and `require_non_empty_control_mode`, then raises the same `ValueError` text as `datasets.factory._validate_metadata_requirements` if any row has an empty value after overrides. Operator no longer burns ~90s on a fit the very next training launch refuses to start.

src/opentau/scripts/fit_fast_tokenizer.py
shuheng-liu: Fixed in 85c535b. `_aggregate_stats_per_head` now tracks `fallback_fired` per dataset and logs the aggregated warning in the same shape `_build_norm_heads` uses: `"N/total datasets lack non-empty robot_type / control_mode and were given a per-dataset fallback norm head (one singleton head each). Set `DatasetConfig.robot_type` / `DatasetConfig.control_mode` to pool them..."`, capped at 10 names with `, ... and N more` suffix.

src/opentau/scripts/fit_fast_tokenizer.py
shuheng-liu: Fixed in 85c535b — downgraded from hard-raise to warn-and-fall-back. Existing `--use-mixture-dataloader` invocations (which never opted into per-head) keep working with no flag change. The warning explicitly tells the operator to either drop `--use-mixture-dataloader` (manual sampler supports per-head) or pass `--no-per-head-norm` to silence the warning. Help text + module docstring updated.

src/opentau/scripts/fit_fast_tokenizer.py
shuheng-liu: Fixed in 85c535b. `_aggregate_stats_per_head` now logs the same `"N/M datasets have native action_dim > --action-dim=K (max ...). Their extra dims will be silently dropped. Confirm this matches the production policy's max_action_dim..."` warning as the global path, with the first 5 offending repo IDs.

src/opentau/scripts/fit_fast_tokenizer.py
shuheng-liu: Fixed in 85c535b. `_normalize_chunks_per_head` now asserts both `sum(per_dataset_chunks) == stacked.shape[0]` and `len(per_dataset_chunks) == len(per_ds_min) == len(per_ds_max)` with explicit messages pointing the future debugger at `_sample_via_manual`'s concat order. New `test_chunk_count_mismatch_raises_assert` pins it.

tests/scripts/test_fit_fast_tokenizer.py
shuheng-liu: Fixed in 85c535b. `TestNormalizeEquivalenceVsProduction::test_per_head_normalize_matches_production_normalize` builds a 2-dataset / 2-head synthetic batch (with zero-padded action-dim suffix), runs both `_normalize_chunks_per_head` and the production `Normalize({"ACTION": MIN_MAX}).forward(batch, dataset_index)` with the same per-dataset (min, max), and asserts equality at `atol=2e-7`. Caught the manual path's float64→float32 ULP drift on the first run; the comment explains why bit-equality isn't claimed. The test also explicitly asserts the padded suffix becomes `-1` in both paths, which is what would have caught finding #1 directly.

src/opentau/scripts/fit_fast_tokenizer.py
shuheng-liu: Fixed in 85c535b. The report now carries `per_head_action_stats: {key: {min: [...], max: [...]}}` on per-head fits, and `n_norm_heads` derives from that map's size. Values are rounded to 6 decimals to keep the JSON readable. `global_action_min` / `global_action_max` stay populated only on global fits (and remain `null` on per-head, by design — they're not the right summary for that path).

src/opentau/scripts/fit_fast_tokenizer.py
shuheng-liu: Fixed in 85c535b. Extracted `_load_metadata_stats_and_info` as the single shared worker that loads `LeRobotDatasetMetadata`, applies `cfg.{robot_type, control_mode}` overrides, resolves the native action key, and returns `(idx, repo_id, info, stats_or_None, err_or_None)`. `_load_dataset_stats` is now a 3-line wrapper that drops the `info` field. `_load_dataset_info_and_stats` removed.

src/opentau/scripts/fit_fast_tokenizer.py
shuheng-liu: Deferred. The right long-term shape is to call `_build_norm_heads` directly on a partial `DatasetMixtureMetadata`, which would eliminate every silent-divergence finding above by construction. The blocker is that `DatasetMixtureMetadata.__init__` does much more than build heads (loads parquet, builds episode indices, computes sample weights) and `_build_norm_heads` reads multiple `self.*` attrs (`dataset_names`, `per_dataset_stats`, `cfg.{max_action_dim, num_cams}`). Doing it right means extracting `_build_norm_heads` (and the `_to_standard_data_format` step it depends on) to a free function — a cross-file refactor touching `dataset_mixture.py` directly. Tracking as a follow-up. The surgical fixes in 85c535b close every concrete silent-divergence the review identified; this one is correctness-equivalent, just less future-proof.

src/opentau/scripts/fit_fast_tokenizer.py
```diff
@@ -394,6 +453,315 @@ def _aggregate_stats_manual(
...
+          map (diagnostics + report).
+    """
+    from concurrent.futures import ProcessPoolExecutor, as_completed
+
+    n = len(mixture_cfg.datasets)
+    raw_min: list[np.ndarray | None] = [None] * n
+    raw_max: list[np.ndarray | None] = [None] * n
+    per_ds_info: list[dict[str, Any]] = [{}] * n
+    per_ds_native_dim: list[int | None] = [None] * n
+    per_ds_repo: list[str] = ["<no-repo-id>"] * n
```
shuheng-liu, line 521: **[LOW / cleanup] `[{}] * n` creates N references to the same dict.**

The `[None] * n` (lines 513-514, 516) and `["<no-repo-id>"] * n` (517) lines are safe because `None` and immutable strings can't be mutated. But `[{}] * n` builds a list where all N slots point at the **same** empty dict — a Python `*`-on-list footgun.

Current code only does `per_ds_info[idx] = info` (line 532), which **reassigns** the slot — safe. But a future contributor doing `per_ds_info[idx]["override_X"] = ...` (legitimate-looking pattern for accumulating per-dataset metadata in a follow-up) would silently mutate every slot.

Fix: `per_ds_info: list[dict[str, Any]] = [{} for _ in range(n)]`.

Same applies to `per_ds_native_dim: list[int | None] = [None] * n` — safe today because `int` is immutable, but if anyone changes it to hold a small mutable object later, the same footgun re-emerges. The list-comprehension form is uniformly safe.

---

src/opentau/scripts/fit_fast_tokenizer.py
```diff
@@ -394,6 +453,315 @@ def _aggregate_stats_manual(
...
+        logger.warning(
+            "%d/%d datasets lack non-empty robot_type / control_mode and were "
+            "given a per-dataset fallback norm head (one singleton head each). "
+            "Set `DatasetConfig.robot_type` / `DatasetConfig.control_mode` to "
+            "pool them into shared heads. Affected: %s%s",
+            len(fallback_datasets),
+            n,
+            shown,
+            suffix,
+        )
```
shuheng-liu, line 614: **[LOW] Re: deferred dedup-on-fallback — the new warning surfaces the symptom but not the specific divergence.**

The new fallback-fired warning at lines 598-610 logs the fallback datasets, which is a real improvement. But for the specific case the deferred finding flagged — a mixture with two entries of `repo_id='X'`, both missing `(rt, cm)` — the warning shows two entries of `"X"` in the affected list. The operator might interpret that as "two X datasets each got a singleton head" (matching training's behavior) when actually they're being **pooled** under one shared key (diverging from training).

Until the dedup is plumbed through (per the deferred follow-up), a tighter fit-time signal would be:

```python
fallback_counts = Counter(per_ds_repo[i] for i in range(n) if fallback_fired_at[i])
duplicates_in_fallback = {k: v for k, v in fallback_counts.items() if v > 1}
if duplicates_in_fallback:
    logger.warning(
        "%d fallback-keyed repo_ids appear in the mixture more than once; "
        "they will be POOLED into one head at fit time, but training "
        "(via _make_dataset_names dedup) keeps them as separate singleton "
        "heads. Set robot_type/control_mode on each entry or add explicit "
        "deduplication to keep fit-time normalization aligned with training. "
        "Duplicates: %s",
        len(duplicates_in_fallback),
        dict(list(duplicates_in_fallback.items())[:10]),
    )
```

Mechanical (one block in the existing fallback-handling loop), no new infra; closes the divergence detection at fit time even before the proper dedup lands.

---

shuheng-liu (commented):
Re-review of 85c535b — 11/13 of the previous round's findings landed cleanly; 2 reasonably deferred. Two small follow-up notes inline.

**Verified fixes (code-read):**

| # | Finding | Status |
|---|---|---|
| 1 | NaN-pad vs zero-pad (HIGH) | ✅ Fix at line 542-543 (`np.zeros` not `np.full(NaN)`); pinned by new `TestNormalizeEquivalenceVsProduction::test_per_head_normalize_matches_production_normalize` which asserts elementwise equality vs the production `Normalize` layer at `atol=2e-7`. |
| 2 | Stats-fail / chunks-succeed (HIGH) | ✅ Subsumed by #4 — no more `[-1, 1]` sentinel path for real chunks. |
| 3 | Fallback-name dedup (HIGH) | ⏭️ Deferred with reasoning. See inline follow-up for a small mitigation at the warning site. |
| 4 | No all-fail guard (HIGH) | ✅ Fail-fast at line 552 on ANY stats failure (stricter than the global path's "all fail" guard). |
| 5 | Inf masking (MED) | ✅ `np.where(np.isfinite(...), ..., np.nan)` pre-mask at lines 695-696; pinned by two new tests. |
| 6 | `require_non_empty_*` (MED) | ✅ Mirrored at lines 563-580. |
| 7 | `fallback_fired` warning (MED) | ✅ Aggregated warning at lines 598-610 matches `_build_norm_heads`'s shape. |
| 8 | `--use-mixture-dataloader` hard-raise (MED) | ✅ Downgraded to warn-and-fall-back at lines 1394-1402. |
| 9 | Over-dim warning (MED) | ✅ Restored at lines 618-635. |
| 10 | Implicit ordering invariant (MED) | ✅ Two asserts at lines 744-752; pinned by `test_chunk_count_mismatch_raises_assert`. |
| 11 | Test gap vs production `Normalize` (LOW) | ✅ Closed by the new equivalence test — would have caught findings 1, 3, and 5 by construction. |
| 12 | `global_action_*` null in report (LOW) | ✅ New `per_head_action_stats` map carries `{key: {"min": [...], "max": [...]}}`. |
| 13 | `_load_dataset_info_and_stats` clone (LOW) | ✅ Collapsed into shared `_load_metadata_stats_and_info`; `_load_dataset_stats` is now a 3-line wrapper. |
| Alt | Construct stub `DatasetMixtureMetadata` | ⏭️ Reasonably deferred — the right shape requires extracting `_build_norm_heads` to a free function (cross-file refactor). The surgical fixes here close every concrete divergence. |

The new equivalence test is particularly load-bearing: it pins the invariant the PR was created to establish, and asserts the padded suffix lands at `-1` in both paths (exactly the silent corruption finding #1 described). Solid.

**New observations (low-priority):**
- One inline on `[{}] * n` aliasing footgun at line 515 — currently safe, but a future mutation would silently fan out.
- One inline suggesting a tighter fit-time signal for the deferred #3 case (duplicate-`repo_id` pooled under fallback).

No new correctness issues found. Ready to mark non-draft once the deferred follow-ups are tracked elsewhere.

---

src/opentau/scripts/fit_fast_tokenizer.py
shuheng-liu: Fixed in c755cb8. Changed `per_ds_info: list[dict[str, Any]] = [{}] * n` to `[{} for _ in range(n)]`. Added a short comment noting that `[None] * n` and `["<no-repo-id>"] * n` are intentionally kept as multiplied-literal because `None` and immutable strings can't be mutated, but anyone changing those slots to a mutable type later should switch to the list-comp form too.

src/opentau/scripts/fit_fast_tokenizer.py
shuheng-liu: Fixed in c755cb8 — added a `Counter`-based block inside the existing fallback-handling branch that surfaces the exact divergence:

> N fallback-keyed repo_id values appear in the mixture more than once. Fit-time POOLS these under one shared head per repo_id, but training (via `_make_dataset_names`'s `#N` dedup) keeps them as separate singleton heads — the fit-time normalization will diverge from training. Set `DatasetConfig.robot_type` / `DatasetConfig.control_mode` on each entry, or deduplicate the mixture, to close the gap. Duplicates (repo_id -> count): {...}

The full dedup fix (replicating `_make_dataset_names`'s `#N` suffix logic at fit time so fit and training produce identical fallback keys) is tracked in [#349](https://github.com/TensorAuto/OpenTau/issues/349).

src/opentau/scripts/fit_fast_tokenizer.py
shuheng-liu: Tracked in [#349](https://github.com/TensorAuto/OpenTau/issues/349). Resolving here; that issue owns the proper dedup fix.

src/opentau/scripts/fit_fast_tokenizer.py
shuheng-liu: Tracked in [#350](https://github.com/TensorAuto/OpenTau/issues/350) — the extraction of `_build_norm_heads` to a free function so `fit_fast_tokenizer` can call it directly. Resolving here; that issue owns the refactor.

shuheng-liu:
## Response to re-review (commit c755cb8)

Re-review verified all 11 fixes from the prior round and flagged 2 new low-priority items. Both addressed; deferred items moved to dedicated issues.

### New inline findings (re-review) — both fixed in c755cb8

- **`[{}] * n` aliasing footgun (line 515).** Switched `per_ds_info` to a list-comprehension form so a future `per_ds_info[idx]["override_X"] = ...` doesn't fan out across slots. Added a comment noting why `[None] * n` / `["<no-repo-id>"] * n` stay as multiplied-literals (immutable element types). Thread resolved.

- **Tighter signal for duplicate-`repo_id` in fallback (line 610).** Added a `Counter`-based block inside the existing fallback-handling branch that detects when a fallback-keyed `repo_id` appears more than once and warns with the exact `{repo_id -> count}` map. Mechanical, no new infra. Closes the divergence-detection gap until the full dedup follow-up lands. Thread resolved.

### Deferred items — now tracked

- Deferred finding #3 (fallback-name dedup at fit time): [#349](https://github.com/TensorAuto/OpenTau/issues/349).
- Deferred altitude refactor (extract `_build_norm_heads` to a free function so `fit_fast_tokenizer` can call it directly): [#350](https://github.com/TensorAuto/OpenTau/issues/350).

Both inline threads updated with issue links and resolved.

### State

All 16 inline review threads on this PR are now resolved (14 from the original review + 2 from the re-review). 11 closed by direct fixes in commit 85c535b, 2 closed by direct fixes in commit c755cb8, 2 closed with cross-references to the new tracking issues.

shuheng-liu (commented):
Re-review of c755cb8 — both low-priority follow-ups from the last round are addressed.

**Verified:**

1. **`[{}] * n` → `[{} for _ in range(n)]`** (line 515) with a tight comment explaining why the adjacent `[None] * n` / `["<no-repo-id>"] * n` lines stay safe. ✅
2. **Duplicate-fallback `Counter` warning** (lines 612-638). The implementation matches the suggested shape: counts only fallback-fired entries, filters to `count > 1`, surfaces the divergence at fit time with a `(repo_id → count)` preview capped at 10. Correctly fires on the exact case the deferred dedup finding flagged (duplicate `repo_id` both lacking `rt`/`cm` → pooled at fit, singleton-per-entry at training) and doesn't fire on legitimate same-`repo_id` entries with distinct `(rt, cm)` overrides (which pool consistently at both fit and training). ✅

No new issues introduced. With the duplicate-detection warning in place, the fit-time signal for the deferred #3 case is now strong enough that the proper dedup refactor can land as a follow-up without operators silently shipping mismatched tokenizers in the meantime.

Nothing else to flag. The PR is in good shape to mark non-draft.

---

PR #351, opened 2026-05-28
perf(pi07_paligemma): skip subgoal image tower when all-padded

shuheng-liu:
## What this does

In `PI07PaligemmaLowLevelFlowMatching._embed_item`, the `image` (subgoal) branch previously ran `paligemma_with_expert.embed_image` **unconditionally** — the comment cited DDP lockstep as the reason (a data-dependent skip would desync per-rank collective counts and hang at NCCL).

This runs the full PaliGemma image tower even when no sample needs a subgoal (inference, or any training step where every rank's `subgoal_is_pad` is `True`), since the per-sample mask zeros those slots downstream anyway.

This PR makes the skip safe instead of avoiding it:

- The per-sample subgoal mask's `.any()` is **OR-reduced across ranks** via the existing `_global_or_branch_decisions` helper (the same pattern `pi07` uses), so every rank takes the *same* run/skip decision and collective counts stay aligned.
- When the global OR is `False`, the image tower is skipped and zeros of shape `(B, num_image_tokens, hidden_size)` are emitted instead. This is mask-equivalent: those tokens carry an all-`False` pad mask, so real tokens can never attend to them and the model output is unchanged — only wasted compute is removed.
- Under `torch.compile` / ONNX export the all-reduce and Python branch can't be traced, so the tower runs unconditionally there (the exported graph must always include it).

⚡️ Performance optimization. Behavior is provably output-equivalent; only redundant vision-encoder work is dropped.

## How it was tested

- `pre-commit run` on the changed files: all hooks pass (ruff lint + format, etc.).
- **CPU** (`pytest -m "not gpu"`): 269 tests pass across the affected files (`test_pi07_paligemma_low_level.py`, `test_pi07_low_level.py`, `test_pi07_cpu.py`, `test_pi07_attention_layout.py`, plus the high-level-planner / video-encoder CPU files). New coverage:
  - `test_all_padded_subgoal_skips_image_tower` / `test_some_real_subgoal_runs_image_tower` — assert the skip emits correctly-shaped zeros without calling `embed_image`, and that any real sample still runs the tower.
  - `test_ddp_lockstep_image_tower_skip` — a 2-rank **gloo** CPU test: with divergent per-rank masks (rank 0 all-padded, rank 1 real) **both** ranks run the tower; with all ranks padded both skip. This is the direct regression guard for the NCCL-divergence concern.
- **GPU** (`pytest -m "gpu" -n 0 tests/policies/test_pi07_paligemma_low_level.py`) on a single GPU: the runnable GPU test passes. The one GPU test that exercises subgoals end-to-end (`test_complete_pi07_low_level_pipeline`) stays hard-skipped for memory, so I separately verified on-GPU that a real PaliGemma `embed_image` emits exactly `text_config.num_image_tokens` (256) tokens — confirming the skip-path zeros `(B, 256, hidden)` are shape-identical to the run path on a real model.
- The nightly regression suite (`regression_test.yml`, g6.12xlarge cron) was not run locally and will run on schedule.

```bash
# Skip-path units + the 2-rank gloo lockstep test (CPU)
pytest -sx tests/policies/test_pi07_paligemma_low_level.py -k "image_tower or lockstep"
```
```bash
# Targeted GPU subset for the policy file
pytest -m "gpu" -n 0 tests/policies/test_pi07_paligemma_low_level.py
```

## Checklist

### DIFF
--- a/src/opentau/policies/pi07_paligemma/low_level/modeling_pi07_low_level.py
+++ b/src/opentau/policies/pi07_paligemma/low_level/modeling_pi07_low_level.py
@@ -43,6 +43,7 @@ from opentau.policies.pi05.paligemma_with_expert import (
     PaliGemmaWithExpertConfig,
     PaliGemmaWithExpertModel,
 )
+from opentau.policies.pi07.low_level.modeling_pi07_low_level import _global_or_branch_decisions
 from opentau.policies.pi07.video_encoder import SpaceTimeSiglipVideoEncoder
 from opentau.policies.pi07_paligemma.low_level.configuration_pi07_low_level import (
     PI07PaligemmaLowLevelConfig,
@@ -1752,12 +1753,20 @@ class PI07PaligemmaLowLevelFlowMatching(nn.Module):
           all-True per-rank), never per-sample data that can diverge
           across ranks — otherwise DDP with
           ``find_unused_parameters=False`` will deadlock.
+        - ``image``: subgoal image through PaliGemma's image tower. Same
+          skip-and-emit-zeros efficiency win as ``video``, but the
+          ``pad_mask`` here *is* per-sample data (``subgoal_is_pad`` can
+          diverge across ranks), so the run/skip decision is OR-reduced
+          across ranks via :func:`_global_or_branch_decisions` to keep the
+          DDP invariant. The skip is bypassed (tower runs unconditionally)
+          under ``torch.compile`` / ONNX export.
         - ``action``: noisy-action projection (suffix only).
 
         The pad mask is broadcast/expanded so the returned mask matches
         the embedded sequence length: per-sample ``(B,)`` is allowed for
-        ``video`` and is expanded to ``(B, num_video_tokens)``; all other
-        types pass through ``(B, L)`` unchanged.
+        ``video`` (expanded to ``(B, num_video_tokens)``) and ``image``
+        (expanded to ``(B, num_image_tokens)``); all other types pass
+        through ``(B, L)`` unchanged.
         """
         t = item.item_type
         if t == "text":
@@ -1776,16 +1785,50 @@ class PI07PaligemmaLowLevelFlowMatching(nn.Module):
         if t == "image":
             # Subgoal images: 4-D (B, C, H, W) in [-1, 1] through the VLM's
             # image tower (byte-identical to embed_video at T=1 on the shared
-            # tower). Run unconditionally so all ranks stay in lockstep — no
-            # data-dependent skip; the per-sample mask zeros padded slots
-            # downstream.
+            # tower). When no sample on ANY rank needs this subgoal, the image
+            # tower is skipped and zeros are emitted instead — a strict
+            # efficiency win (the per-sample mask zeros the slot regardless),
+            # e.g. at inference or whenever every rank's subgoal_is_pad is True.
+            #
+            # The skip is data-dependent: unlike the config-determined video
+            # mask, per-sample subgoal_is_pad can differ across ranks, so each
+            # rank's local ``.any()`` may diverge. Running the tower on some
+            # ranks but not others would desync DDP / FSDP collective counts and
+            # hang at NCCL (CLAUDE.md hard rule 5), so the run/skip decision is
+            # OR-reduced across ranks first via ``_global_or_branch_decisions``
+            # — presence is structurally True here (the image item is in the
+            # prefix on every rank by config), so only the ``.any()`` is
+            # reduced. Under torch.compile / ONNX export the all-reduce and the
+            # Python branch can't be traced (and export must bake in the tower
+            # so the exported graph can encode real subgoals), so run
+            # unconditionally.
             image = item.data
             per_sample_mask = item.pad_mask
             if per_sample_mask.ndim != 1:
                 raise ValueError(
                     f"image pad_mask must be (B,) per-sample; got shape {tuple(per_sample_mask.shape)}."
                 )
-            emb = self.paligemma_with_expert.embed_image(image).to(dtype=_preferred_dtype())
+            dtype = _preferred_dtype()
+            if torch.compiler.is_compiling() or torch.onnx.is_in_onnx_export():
+                run_image_tower = True
+            else:
+                (run_image_tower,) = _global_or_branch_decisions(
+                    presence_locals=(True,),
+                    any_locals=(bool(per_sample_mask.any()),),
+                    field_names=("subgoal_image",),
+                    device=image.device,
+                )
+            if run_image_tower:
+                emb = self.paligemma_with_expert.embed_image(image).to(dtype=dtype)
+            else:
+                pg_text_cfg = self.paligemma_with_expert.config.paligemma_config.text_config
+                emb = torch.zeros(
+                    image.shape[0],
+                    pg_text_cfg.num_image_tokens,
+                    pg_text_cfg.hidden_size,
+                    device=image.device,
+                    dtype=dtype,
+                )
             expanded = per_sample_mask[:, None].expand(emb.shape[0], emb.shape[1])
             return emb, expanded
         if t == "video":
--- a/src/opentau/policies/pi07_paligemma/low_level/modeling_pi07_low_level.py
+++ b/src/opentau/policies/pi07_paligemma/low_level/modeling_pi07_low_level.py
@@ -1732,7 +1732,51 @@ class PI07PaligemmaLowLevelFlowMatching(nn.Module):
         """
         return self.video_encoder(video, obs_history_is_pad=obs_history_is_pad)
 
-    def _embed_item(self, item: ContextItem) -> tuple[Tensor, Tensor]:
+    def _global_run_image_tower(self, items: list[ContextItem]) -> bool:
+        """Decide once, across all ranks, whether the subgoal image tower runs.
+
+        :meth:`embed_prefix` calls this exactly once per forward — before the
+        per-item loop and on *every* rank, even one whose prefix happens to
+        contain no ``image`` item — then threads the result into each
+        :meth:`_embed_item` via ``run_image_tower``. Hoisting the collective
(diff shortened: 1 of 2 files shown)


src/opentau/policies/pi07_paligemma/low_level/modeling_pi07_low_level.py
```diff
@@ -1776,16 +1785,50 @@ def _embed_item(self, item: ContextItem) -> tuple[Tensor, Tensor]:
...
                 raise ValueError(
                     f"image pad_mask must be (B,) per-sample; got shape {tuple(per_sample_mask.shape)}."
                 )
-            emb = self.paligemma_with_expert.embed_image(image).to(dtype=_preferred_dtype())
+            dtype = _preferred_dtype()
+            if torch.compiler.is_compiling() or torch.onnx.is_in_onnx_export():
+                run_image_tower = True
+            else:
+                (run_image_tower,) = _global_or_branch_decisions(
+                    presence_locals=(True,),
```
shuheng-liu: **Design note (not a blocker for the current config, but worth recording): this places the cross-rank collective *inside* a per-item branch, which subtly weakens the invariant relative to the `pi07` reference.**

In `pi07`'s `embed_prefix`, `_global_or_branch_decisions` is called **once, unconditionally**, before any per-item loop, and it receives a real `subgoal_present_local = bool(subgoal_images)`. That means every rank reaches the all-reduce exactly once per `embed_prefix`, *even a rank that has no subgoal item at all* — which is precisely what lets the helper's presence-divergence check (None-vs-Tensor) actually fire and prevents the collective count from diverging.

Here the call lives inside the `_embed_item("image")` branch with `presence_locals=(True,)` hardcoded. Two consequences:

1. **The presence check is defeated.** Hardcoding `True` means the helper can never detect cross-rank presence disagreement — and a rank that has *zero* image items never executes this branch, so it never even issues the all-reduce. If a future heterogeneous-mixture change ever gives different ranks a different number of subgoal-image `ContextItem`s, the per-item all-reduce count diverges across ranks and you get exactly the NCCL hang this helper exists to prevent — but now undetectably, since the guard rail was bypassed. This is the CLAUDE.md rule-5 anti-pattern ("a collective behind a per-rank branch decision"), safe today only because the prefix layout is globally uniform.

2. **One all-reduce per image item.** With multiple subgoal cameras you fire N separate 2-int all-reduces per step, vs `pi07`'s single batched reduce. Negligible cost, but it's also on the hot path on every step (even when subgoals are present and nothing is skipped), so the optimization adds a small fixed sync cost to buy savings only in the all-padded case.

Neither is a correctness bug for the current homogeneous config — every rank has an identical prefix layout, so the branch is reached the same number of times on every rank and lockstep holds. But if you want the same robustness `pi07` has, consider hoisting the decision out of `_embed_item` into `embed_prefix` (compute the global OR once over the image items, pass it down), so a real `presence_local` flows through the helper and a rank lacking the item still participates. At minimum, the hardcoded `presence_locals=(True,)` deserves a comment explaining that it only holds while subgoal-image presence is a global config property.

---

shuheng-liu (commented):
## Review: skip subgoal image tower when globally all-padded

Reviewed the diff, the `_global_or_branch_decisions` helper, the established `video`-branch skip pattern it mirrors, the PaliGemma config shape assumptions, and the new tests. Overall this is a clean, well-reasoned optimization and the core claim holds up. CI is green (CPU tests, pre-commit, checklist all pass); no prior reviews.

### What's correct ✅
- **Output equivalence is sound.** When the tower is skipped, `emb` is zeros and the returned `pad_mask` is all-`False`. In `make_att_2d_masks`, `pad_2d_masks = pad_masks[:,None,:] * pad_masks[:,:,None]` zeroes those key columns, so no real token can attend to the skipped slots and the output for real tokens is unchanged. This is the same guarantee the existing `video` skip relies on. Even if `num_image_tokens` were ever stale, an all-`False` block can't corrupt output — it only changes wasted compute.
- **Shape assumption checks out.** `embed_image` → `get_image_features` returns `(B, num_image_tokens, hidden)`, and `paligemma_config.text_config.num_image_tokens` is explicitly `256` in the config here, matching the (224/14)² = 256 SigLIP token count you verified on GPU. The skip-path zeros are width-identical to the run path.
- **The central design decision is right.** Unlike the config-determined `video` mask (which can safely use a local `.any()`), `subgoal_is_pad` is genuinely per-sample and can diverge across ranks, so OR-reducing the run/skip decision via `_global_or_branch_decisions` is the correct fix and matches `pi07`'s approach. The `divergent` gloo test is a good direct regression guard for the lockstep property.
- **compile/ONNX guard** matches the repo idiom (`torch.compiler.is_compiling() or torch.onnx.is_in_onnx_export()`), and baking the tower into the exported graph is the right call.
- **Test coverage** is solid: all-padded skip, some-real run, and the 2-rank gloo lockstep test (correctly `@pytest.mark.slow`, module-level worker for `spawn` picklability).

### Main observation ⚠️ (see inline comment)
The cross-rank collective is now placed **inside** the `_embed_item("image")` branch with `presence_locals=(True,)` hardcoded, whereas `pi07` calls the helper **once, unconditionally** in `embed_prefix` with a real `subgoal_present_local`. For the current homogeneous config every rank has an identical prefix layout, so this is safe. But it defeats the helper's presence-divergence detection and means a rank lacking the image item would skip the all-reduce entirely — re-introducing the exact NCCL-hang class this guard was built to catch, undetectably, if a future heterogeneous-mixture change ever makes subgoal-image presence non-uniform. Details + a suggested hoist-into-`embed_prefix` fix in the inline comment. Not a blocker for merging as-is, but worth either fixing or documenting the assumption.

### Minor
- One all-reduce fires per image item per step (vs `pi07`'s single batched reduce), and on every step regardless of whether anything is skipped. Cost is negligible, just noting the optimization adds a small fixed sync to the hot path.

Nice work — the equivalence reasoning and the lockstep test are exactly the right things to nail down for this change.

---

src/opentau/policies/pi07_paligemma/low_level/modeling_pi07_low_level.py
shuheng-liu: Fixed in 8b12b62 — hoisted the collective out of the per-item branch.

New `_global_run_image_tower(items)` is called **once per `embed_prefix`**, on every rank (even one whose prefix has no `image` item), with a real `subgoal_present_local = bool(image_items)`. The decision is threaded into each `_embed_item` via `run_image_tower`, so:

- The collective no longer hides behind the `image` branch — a rank lacking the item still participates in the all-reduce.
- The presence-divergence check is live again: a real `presence_local` flows through `_global_or_branch_decisions`, so a future heterogeneous mixture that gives ranks a different number of subgoal-image items raises loudly instead of silently desyncing the collective count.
- One reduce per forward instead of one per image item (the minor point).

`_embed_item` keeps a `run_image_tower=None` local fallback (fires no collective) only for direct single-process unit-test calls.

Tests: added a `presence_divergence` scenario to the gloo lockstep test (asserts the mismatch raises `RuntimeError`), plus `run_image_tower` override and single-process `_global_run_image_tower` unit tests. CPU suite + lockstep green.

shuheng-liu (commented):
## Re-review of commit `8b12b62` (hoist image-tower collective into embed_prefix)

This follow-up directly and fully addresses my earlier finding. 👍

**The fix is correct and matches the `pi07` reference:**
- The collective is now in a new `_global_run_image_tower(items)`, called **once per `embed_prefix`, before the loop, on every rank** — including a rank whose prefix contains no `image` item. That's exactly the `pi07` shape.
- `subgoal_present_local = bool(image_items)` is a **real** presence flag now (not hardcoded `True`), so `_global_or_branch_decisions`' presence-divergence check actually fires: a heterogeneous mixture where ranks disagree on subgoal-image presence raises a loud `RuntimeError` instead of silently desyncing the collective count and hanging at NCCL.
- The OR runs over all cameras and gates them together (`any(it.pad_mask.any() ...)`), matching `pi07`'s single `has_subgoal`. Mask-equivalent as before, and it collapses the previous N-all-reduces-per-item down to one — so my minor efficiency note is resolved too.
- `_embed_item(..., run_image_tower=None)` keeps a collective-free local fallback for direct unit-test calls; in the real path `embed_prefix` always threads in a concrete bool, so the None branch is test-only.

**Tests are exactly the right additions:**
- `test_ddp_lockstep_image_tower_skip` now drives `_global_run_image_tower` directly and adds a **`presence_divergence`** scenario (rank 0 has the image item, rank 1 doesn't) asserting the loud `RuntimeError` — a direct regression guard for the restored protection.
- `test_run_image_tower_override_wins_over_local_mask` confirms the threaded decision overrides the local mask both directions (`False` skips a real sample; `True` runs an all-padded one).
- `test_global_run_image_tower_local_decision` covers the no-image-item case returning `False` without crashing on device derivation.

**Verification:** CI is green on `8b12b62` — CPU Tests, Pre-commit, and check-checklist all pass.

**Tiny nit (non-blocking):** `_global_run_image_tower` derives the all-reduce device from `items[0].data.device`, which assumes `items` is non-empty and that the first item's tensor sits on the distributed-backend device (true in practice — video/text always precede the subgoal block, and all inputs are on the model device during training). `pi07` keys off a known tensor (`lang_tokens.device`); not worth changing, just noting the implicit assumption.

Output-equivalence reasoning, lockstep guarantee, and the presence guard are all solid now. Nice turnaround on the review feedback — this looks good to me.

---

shuheng-liu:
Re: the `_global_run_image_tower` device-derivation nit — leaving it as-is, on purpose.

`embed_prefix` is only ever reached through `_build_prefix_items`, which unconditionally prepends the video / prompt-text / state items before any subgoal block, so `items` is never empty and `items[0].data` is always a tensor on the model (distributed-backend) device. An empty-`items` guard would be defensive code for a state that can't occur, so I've left the code unchanged and am recording the assumption here for the next reader. Thanks for the thorough re-review.

PR #354, opened 2026-05-29
fix(scripts): commit wandb step after all per-step logging

### LINKED ISSUE #353: wandb: latest step's metrics only flush when the next step logs (explicit-step commit lag)
## Symptom

A step's metrics (and especially eval results + videos) don't appear on the wandb dashboard until the **next** logging step fires. With `eval_freq=1000` and ~20 min between training logs, you finish a multi-minute eval and then wait ~20 min — until the next training step logs — before the eval point and its videos show up in the workspace. The media files appear in the **Files** tab promptly (eager upload), but the **charts/panels** lag by one logging interval.

Two consequences:
1. **Display lag**: the newest logged step always trails by one interval. Cosmetic during training, but annoying for eval — you can't see the eval result right after it runs.
2. **Ungraceful-exit loss**: if the run is killed (e.g. `scancel` without a graceful `wandb.finish()`) the final *pending* step can be missing from the server even though it's in the local `.wandb` transaction log. (Recoverable with `wandb sync <run_dir>`, but easy to miss.)

## Root cause

`scripts/train.py` logs with an **explicit `step=`** and makes ~20 separate `accelerator.log({single_key: val}, step=step)` calls per step — training block (`train.py:684-693`), validation block (`train.py:790-812`), eval block (`train.py:869-890`). None pass a `commit` kwarg.

When you call `wandb.log(data, step=N)` with an explicit step, wandb does **not** seal step N's row immediately — it keeps N as the "current pending" row so that further `log(..., step=N)` calls can accumulate more keys into the same row (which OpenTau relies on — that's how 20 separate calls land in one row). wandb only finalizes + flushes row N when it sees a `log()` with `step > N`, or on `wandb.finish()`, or a periodic flush. So row N is committed when N+1's first log call arrives.

## Mechanism for a fix (verified)

`accelerate` 1.12's `WandBTracker.log(values, step, **kwargs)` forwards `**kwargs` straight to `wandb.run.log(values, step=step, **kwargs)`, and `Accelerator.log(values, step, log_kwargs={})` passes `log_kwargs[tracker.name]` to each tracker. So a `commit=True` can be threaded through, tracker-safe (other trackers never see it):

```python
accelerator.log(values, step=step, log_kwargs={"wandb": {"commit": True}})
```

## Proposed fix

Add a single **explicit commit point per logging step**, placed *after* all conditional logging blocks (training + validation + eval) for that step, so the row is sealed and flushed as soon as everything for step N is logged — not when N+1 arrives.

Minimal shape: a small helper invoked unconditionally at the bottom of the per-step logging section:

```python
def _commit_wandb_step(accelerator, step):
    # Seal + flush the pending wandb row for `step` immediately, instead of
    # waiting for the next step's first log() to auto-commit it. Tracker-safe:
    # non-wandb trackers ignore log_kwargs they aren't keyed for.
    accelerator.log({}, step=step, log_kwargs={"wandb": {"commit": True}})
```

(Verify `log({}, commit=True)` reliably commits the pending row; if an empty dict is a no-op in some wandb version, fold `commit=True` into the genuinely-last metric log of the step instead.)

### Important caveat — commit must be last

After step N is committed, **no further `log(..., step=N)` may be issued** — wandb requires monotonically increasing steps and will warn-and-drop a late write to a sealed step. So the commit must come after every training/validation/eval log for that step. The blocks fire at different cadences (`log_freq` / `val_freq` / `eval_freq`), and when several coincide the last block varies, so a single unconditional flush at the very bottom of the logging section is safer than tacking `commit=True` onto one specific call.

## Alternative considered

Consolidate the ~20 per-section `accelerator.log({single_key: val}, ...)` calls into one dict per section (`accelerator.log({...all training metrics...}, step=step)`). Worth doing for clarity/overhead regardless, but it does **not** by itself fix the lag — a single explicit-step `log()` still holds the row open until the next step. The commit point is the actual fix; the consolidation is a nice-to-have that pairs well with it.

## Acceptance criteria

- After an eval at step N completes, its `Success Rate` / `Success/<group>` scalars and `Eval Videos/*` panels appear on the dashboard **without** waiting for step N+1.
- The final step before a clean shutdown is present on the server.
- x-axis (`_step`) still aligns with the training step for all metric families (training/validation/eval), and no "step must be monotonically increasing" warnings appear in the logs.
- Non-wandb trackers (if configured) are unaffected.

## Notes

Low priority / quality-of-life — no correctness impact on training itself; purely about observability latency and resilience of the last logged step. Surfaced while watching a long pretrain run: each eval's results lagged ~20 min behind the eval actually finishing.

shuheng-liu:
## What this does
Fixes #353.

On the wandb dashboard a step's metrics — especially eval scalars and `Eval Videos/*` panels — only appear once the **next** logging step fires. With `eval_freq=1000` and ~20 min between training logs, a just-finished eval doesn't show up in the workspace until the next training step logs ~20 min later.

**Root cause:** `scripts/train.py` logs every metric with an explicit `step=` (so the many per-step `accelerator.log({single_key: val}, step=step)` calls all accumulate into one row). With an explicit step and no `commit` kwarg, wandb keeps row N open as the "pending" row and only seals + uploads it when it next sees a `log(step > N)`, on `wandb.finish()`, or a periodic flush. Hence the one-interval display lag, plus the risk of losing the last pending row on an ungraceful kill (recoverable with `wandb sync`, but easy to miss).

**Fix:** add `_commit_wandb_step()` and call it once at the bottom of the step loop, *after* the training / validation / eval logging blocks, gated on the same per-step flags (`is_log_step or is_val_step or (is_eval_step and eval_envs)`) so it only fires on steps that actually logged, and only on the main process where the trackers live. It threads `commit=True` to wandb via `log_kwargs={"wandb": {"commit": True}}` — `accelerator.log` forwards those kwargs straight to `wandb.run.log` and only to the wandb tracker, so it's tracker-safe — with an empty `values` dict, which flushes the accumulated row without adding keys. Keeping it the last `log(step=N)` for the step means no "step must be monotonically increasing" warnings and the x-axis stays aligned across all metric families. `accelerator.end_training()` still does the final `wandb.finish()` flush after the loop; the per-step commit additionally makes each logged step durable mid-run.

## How it was tested
- Added `TestCommitWandbStep::test_emits_empty_commit_log_at_step` in `tests/scripts/test_train.py`, asserting the helper emits exactly one `log({}, step=N, log_kwargs={"wandb": {"commit": True}})` call.
- Ran `pre-commit run --all-files` (clean) and `pytest -m "not gpu" tests/scripts/test_train.py` (18 passed) locally.
- Behavior verified against the installed `wandb==0.27.0` / `accelerate==1.12.0` source: `accelerator.log({}, step=N, log_kwargs={"wandb": {"commit": True}})` → `wandb.run.log({}, step=N, commit=True)` → `publish_partial_history(..., flush=True)`, which seals row N immediately. End-to-end placement in the full training loop is exercised by `regression_test.yml` (GPU nightly); not run on this CPU-only machine.

```bash
pytest -sx tests/scripts/test_train.py::TestCommitWandbStep
```

## Checklist

### DIFF
--- a/src/opentau/scripts/train.py
+++ b/src/opentau/scripts/train.py
@@ -186,6 +186,36 @@ def _observe_optional(
     setattr(tracker, key, value)
 
 
+def _commit_wandb_step(accelerator: accelerate.Accelerator, step: int) -> None:
+    """Seal + flush the pending wandb row for ``step`` immediately.
+
+    OpenTau logs every metric with an explicit ``step=``, so wandb keeps row N
+    open as the "pending" row (letting the many per-step ``accelerator.log(...)``
+    calls accumulate into it) and only seals + uploads it when it next sees a
+    ``log(step > N)`` -- i.e. when step N+1 first logs. That makes the newest
+    point (especially a just-finished eval and its videos) lag one logging
+    interval on the dashboard, and risks losing the last pending row on an
+    ungraceful kill.
+
+    Calling this once, after every training / validation / eval log for
+    ``step``, forces the seal now. ``accelerator.log(values, step, log_kwargs)``
+    forwards ``log_kwargs["wandb"]`` straight to ``wandb.run.log`` (other
+    trackers only see kwargs keyed to their own name), so ``commit=True`` is
+    tracker-safe; an empty ``values`` dict flushes the accumulated row without
+    adding keys.
+
+    Must be the LAST wandb write for ``step``: wandb requires monotonically
+    increasing steps and warns-and-drops any ``log(step=N)`` issued after N is
+    sealed.
+
+    Args:
+        accelerator: The active accelerate ``Accelerator`` (caller must already
+            be on the main process, where the trackers live).
+        step: The training step whose pending wandb row should be sealed.
+    """
+    accelerator.log({}, step=step, log_kwargs={"wandb": {"commit": True}})
+
+
 def _mixture_weighted_aggregate(
     per_dataset_trackers: dict[str, MetricsTracker],
     name_to_weight: dict[str, float],
@@ -918,6 +948,15 @@ def train(cfg: TrainPipelineConfig):
                             s["post_resv"],
                             s["post_alloc"] - s["pre_alloc"],
                         )
+
+        # Seal this step's wandb row as soon as all of its logging (training /
+        # validation / eval scalars + eval videos) is done, instead of waiting
+        # for the next logging step to auto-commit it. Guarded on the same
+        # conditions that gate the logging blocks above, so we only commit on
+        # steps that actually logged, and only on main where the trackers live.
+        # See issue #353.
+        if accelerator.is_main_process and (is_log_step or is_val_step or (is_eval_step and eval_envs)):
+            _commit_wandb_step(accelerator, step)
 
     if cfg.eval_freq > 0 and eval_envs:
         close_envs(eval_envs)
--- a/tests/scripts/test_train.py
+++ b/tests/scripts/test_train.py
@@ -29,6 +29,7 @@ import accelerate
 import pytest
 
 from opentau.scripts.train import (
+    _commit_wandb_step,
     _find_unused_params_from_env,
     _mixture_weighted_aggregate,
     _sync_deepspeed_gradient_accumulation_steps,
@@ -258,6 +259,26 @@ class TestMixtureWeightedAggregate:
         assert agg["loss"] == pytest.approx(0.25 * 1.0 + 0.75 * 5.0)
         assert agg["mse_loss"] == pytest.approx(0.25 * 2.0 + 0.75 * 6.0)
         assert agg["ce_loss"] == pytest.approx(0.25 * 3.0 + 0.75 * 7.0)
+
+
+class TestCommitWandbStep:
+    """``_commit_wandb_step`` issues exactly one empty, commit-tagged log."""
+
+    def test_emits_empty_commit_log_at_step(self):
+        # The fix relies on three properties of the emitted call: an empty
+        # ``values`` dict (flush the accumulated row, add no keys), the explicit
+        # ``step`` (seal that row, not some other), and ``commit=True`` routed
+        # only under the ``"wandb"`` key (tracker-safe; other trackers never see
+        # it). Assert the exact call shape to pin all three.
+        calls = []
+
+        def _log(values, step=None, log_kwargs=None):
+            calls.append((values, step, log_kwargs))
+
+        accelerator = SimpleNamespace(log=_log)
+        _commit_wandb_step(accelerator, 1234)
+
+        assert calls == [({}, 1234, {"wandb": {"commit": True}})]
 
 
 if __name__ == "__main__":


shuheng-liu (commented):
## Review findings

Reviewed the diff against the current `train.py` loop and the linked issue #353. The fix is correct, minimal, and well-placed. Details below.

### Verified ✅

1. **Forwarding mechanism is real.** Confirmed against the actual `accelerate==1.12.0` source: `WandBTracker.log(self, values, step=None, **kwargs)` is `@on_main_process` and does `self.run.log(values, step=step, **kwargs)`, and `Accelerator.log` routes `log_kwargs.get(tracker.name, {})` per tracker. So `log_kwargs={"wandb": {"commit": True}}` reaches `wandb.run.log(..., commit=True)` and **only** the wandb tracker — tracker-safe as claimed.
2. **Versions match the lockfile.** `uv.lock` pins `accelerate 1.12.0` and `wandb 0.27.0`, exactly the versions the PR body verified against.
3. **Placement is the last wandb write for the step.** The commit sits at the bottom of the per-step `for` loop, after the eval block (the trailing eval-HBM-probe block uses `logging.info`, not `accelerator.log`). No `accelerator.log(step=N)` fires after it within the iteration, and the next iteration logs `step=N+1 > N` — so monotonicity holds and there's no "step must be monotonically increasing" warn-and-drop. This matches the issue's "commit must be last" caveat.
4. **Guard mirrors the logging gates exactly.** `is_log_step or is_val_step or (is_eval_step and eval_envs)` is the union of the three blocks' own conditions, so the commit fires only on steps that actually logged. `is_saving_step` is correctly excluded (saving issues no wandb writes).
5. **Test is correct and self-contained** (`SimpleNamespace` already imported); it pins all three load-bearing properties of the emitted call.

### Non-blocking notes

- **The one thing CPU can't prove** (and you flagged this honestly): that `wandb.run.log({}, step=N, commit=True)` actually *seals* row N rather than no-opping on an empty dict. The trace to `publish_partial_history(flush=True)` is the right mechanism, but the real acceptance criterion in #353 — "eval scalars + `Eval Videos/*` appear without waiting for step N+1" — is only confirmable on a live wandb dashboard. Worth eyeballing on the next nightly `regression_test.yml` run or a smoke run with wandb enabled. If empty-dict semantics ever change upstream, the documented fallback is folding `commit=True` into the genuinely-last metric log of the step.
- **`accelerator.is_main_process` guard is technically redundant** — both `Accelerator.log` and `WandBTracker.log` are `@on_main_process`, so it would no-op off-main anyway. But it matches the surrounding style (every other log call sits under an `is_main_process` guard) and avoids building the call off-main. Keep it.
- **wandb-disabled path is safe**: with `cfg.wandb.enable=False` no trackers are registered, so `accelerator.log({})` is a no-op — consistent with the other unconditional `accelerator.log` calls already in the loop.

No requested changes. LGTM once the dashboard behavior is eyeballed on a GPU/wandb run.

---

PR #355, opened 2026-05-29
fix(scripts): compact LIBERO eval logging, drop verbose video_paths

shuheng-liu:
## What this does

LIBERO eval logging in `scripts/train.py` dumped the `per_task`, `per_group`,
and `overall` sections of `eval_info` verbatim on every eval. Each section
carried the full per-episode `video_paths` list, so a single eval flooded the
console with hundreds of lines of repeated paths.

This replaces that loop with a compact one-line-per-group summary plus an
overall line:

```
eval[libero_10]: success=0.0% ∑rwrd=0.000 n=64
eval[overall]: success=0.0% ∑rwrd=0.000 n=64
```

Nothing is lost: the complete `eval_info` (including all video paths) is still
written to `eval_info.json` in the per-step videos dir, exactly as before.

## How it was tested

- `ruff check` / `ruff format --check` pass; full pre-commit suite passes.
- Logging-only change: the new `_fmt` helper reads the same `pc_success`,
  `avg_sum_reward`, and `n_episodes` keys already produced by
  `consolidate_eval_info`, and the wandb/tracker logging below it is untouched.

Run any config with a LIBERO eval enabled and observe the per-group + overall
summary lines in place of the previous multi-hundred-line dumps:

```bash
opentau-train --config_path=configs/examples/pi05_training_config.json
```

## Checklist

### DIFF
--- a/src/opentau/scripts/train.py
+++ b/src/opentau/scripts/train.py
@@ -879,9 +879,7 @@ def train(cfg: TrainPipelineConfig):
                 # compact per-group + overall summary (drop verbose video_paths)
                 def _fmt(m: dict) -> str:
                     return (
-                        f"success={m.get('pc_success', float('nan')):.1f}% "
-                        f"∑rwrd={m.get('avg_sum_reward', float('nan')):.3f} "
-                        f"n={m.get('n_episodes', 0)}"
+                        f"success={m['pc_success']:.1f}% ∑rwrd={m['avg_sum_reward']:.3f} n={m['n_episodes']}"
                     )
 
                 for group, v in eval_info["per_group"].items():
--- a/src/opentau/scripts/train.py
+++ b/src/opentau/scripts/train.py
@@ -876,9 +876,17 @@ def train(cfg: TrainPipelineConfig):
                 # overall metrics (suite-agnostic)
                 aggregated = eval_info["overall"]
 
-                # optional: per-suite logging
-                for suite, suite_info in eval_info.items():
-                    logging.info("Suite %s aggregated: %s", suite, suite_info)
+                # compact per-group + overall summary (drop verbose video_paths)
+                def _fmt(m: dict) -> str:
+                    return (
+                        f"success={m.get('pc_success', float('nan')):.1f}% "
+                        f"∑rwrd={m.get('avg_sum_reward', float('nan')):.3f} "
+                        f"n={m.get('n_episodes', 0)}"
+                    )
+
+                for group, v in eval_info["per_group"].items():
+                    logging.info("eval[%s]: %s", group, _fmt(v))
+                logging.info("eval[overall]: %s", _fmt(aggregated))
 
                 # meters/tracker
                 eval_metrics = {


shuheng-liu (commented):
## Review

LGTM overall — clean, surgical, logging-only change. I traced the new code against the data it consumes and it's correct.

### Verified
- `_fmt` reads `pc_success`, `avg_sum_reward`, `n_episodes`. All three are produced by `consolidate_eval_info` (`scripts/eval.py`) for **every** `per_group` value *and* for `overall`, so the format string never hits its defaults in practice.
- Iterating `eval_info["per_group"]` is not a new structural assumption — the `Success/{group}` wandb loop a few lines below already iterates it and indexes `v["pc_success"]` directly.
- `aggregated = eval_info["overall"]` is still consumed by `eval_tracker` below, so dropping the old loop leaves no orphan.
- Full `eval_info` (including all `video_paths`) is still written to `eval_info.json`, so the "nothing is lost" claim holds.
- `nan` formatting is safe: `f"{float('nan'):.1f}"` → `"nan"`, no exception.

### Minor / non-blocking
1. **`avg_max_reward` dropped from the console line.** The old verbose dump included it; the new summary doesn't. Intentional and still in the JSON — just flagging in case you want it in the one-liner too.
2. **Defensive `.get(...)` defaults are slightly inconsistent.** `_fmt` falls back to `nan`/`0`, but the `Success/{group}` loop just below indexes `v["pc_success"]` directly. If a key were ever genuinely absent, `_fmt` would degrade gracefully while that loop would `KeyError` — so the defaults give a false sense of robustness. Since `consolidate_eval_info` always emits these keys, neither can actually fail; consider dropping the defaults for consistency, or leave as-is.
3. **Docstring checkbox is checked but `_fmt` has none.** It's a trivial nested formatter, consistent with sibling nested helpers (`_make_val_tracker`, `_accumulate_to`), so I don't consider this blocking.

No correctness or distributed-safety concerns (the change is inside the existing `accelerator.is_main_process` guard and touches no collectives).

---

shuheng-liu (commented):
## Re-review (commit `99876136`)

Re-checked after the latest push. LGTM — the change is correct and now tighter than the first pass.

### What changed since my last review
- `_fmt` switched from `.get('pc_success', nan)` / `.get('avg_sum_reward', nan)` / `.get('n_episodes', 0)` to direct indexing `m['pc_success']` / `m['avg_sum_reward']` / `m['n_episodes']`. This resolves nit #2 from my previous review: the formatter is now consistent with the `Success/{group}` wandb loop just below (which also indexes `v["pc_success"]` directly), and there's no longer a false sense of robustness from defaults that could never trigger anyway.

### Verified (still holds)
- All three indexed keys are produced by `consolidate_eval_info` for every `per_group` value **and** for `overall`, so direct indexing can't `KeyError` in practice.
- `nan` formatting is safe: empty groups yield `float("nan")` from `_agg_from_list`, and `f"{nan:.1f}"` → `"nan"`.
- `aggregated = eval_info["overall"]` is still consumed by `eval_tracker`; full `eval_info.json` dump is untouched.

### Remaining minor / non-blocking
- `avg_max_reward` is still absent from the console one-liner (present in the JSON) — flagging only in case you want it inline.
- `_fmt` still has no docstring despite the docstring checkbox; trivial nested formatter, consistent with sibling helpers — not blocking.

No correctness or distributed-safety concerns. Still inside the `accelerator.is_main_process` guard, no collectives touched.

---

PR #356, opened 2026-05-29
build(deps): numpy 2.x / gymnasium / CMake 4 LIBERO, drop libero+urdf conflict

shuheng-liu:
## What this does

The forked LIBERO ([shuheng-liu/LIBERO#1](https://github.com/shuheng-liu/LIBERO/pull/1)) now runs on **numpy 2.x** and **gymnasium** (instead of the deprecated `gym`), and installs its `egl_probe` dependency from a fork whose `CMakeLists` builds on **CMake ≥ 4**. That removes the reason for every LIBERO-specific workaround in our build config. This PR (deps + docs only — no code/function changes) deletes them:

- **`pyproject.toml` (`libero` extra):** drop the `numpy<2`, `gym>=0.25,<0.27`, and `cmake<4` pins; add `egl-probe`, sourced from the `v1.0.1-cmake4` fork via `[tool.uv.sources]`.
- **`[tool.uv]`:** remove the `conflicts` block that declared `libero` and `urdf` mutually exclusive (they now share a single numpy 2.x resolution), and remove the `extra-build-dependencies = { egl-probe = ["cmake<4"] }` build pin. `required-version = ">=0.8.4"` is kept, reworded to a generic lockfile/feature baseline.
- **`uv.lock`:** regenerated — `libero` advances to the merged commit, `egl-probe` switches to the fork, `gym` is removed, and `numpy` / `rerun-sdk` collapse from two pinned versions to a single `2.2.6` / `0.29.0`.
- **Docs:** `CLAUDE.md`, `docs/source/installation.rst`, `docs/source/contributing.rst` drop the mutual-exclusivity / `numpy<2` / cmake notes.

Labels: ⚡️ optimization + 📝 documentation.

## How it was tested

On a Linux + NVIDIA GPU machine, in an isolated worktree synced from this branch:

- `uv lock` resolves cleanly: single `numpy 2.2.6`, single `rerun-sdk 0.29.0`, `gymnasium 1.2.3`, `egl-probe` from the fork, `gym` removed.
- `uv sync --extra dev --extra libero` succeeds — `egl_probe` builds under CMake 4.2.1 (uv build isolation), directly exercising the CMake-4 fix.
- Import smoke: `numpy` 2.2.6, `gymnasium` 1.2.3, `robosuite` 1.4.0, `libero` + `OffScreenRenderEnv` import; legacy `gym` is absent.
- CPU tests: `pytest -m "not gpu" tests/envs/test_libero_control_freq.py tests/utils/test_libero_utils.py` → 16 passed.
- Real-env test: `pytest -m gpu tests/envs/test_libero_control_freq.py::test_control_freq_reaches_real_robosuite_sim` → 1 passed (builds + steps a real `OffScreenRenderEnv` under EGL + numpy 2.x + gymnasium 1.x, using bundled bddl/init_states — no dataset download).
- Former conflict: `uv sync --extra dev --extra libero --extra urdf` installs both extras together; `numpy`, `rerun` (0.29.0), `libero`, and `gymnasium` coexist on one numpy 2.x.

These are the same paths `cpu_test.yml` / `gpu_test.yml` exercise in CI.

```bash
# resolves to a single numpy 2.x and a single rerun — libero + urdf no longer conflict
uv sync --extra dev --extra libero --extra urdf
```
```bash
# real LIBERO env on the new stack (needs an NVIDIA GPU + EGL libs)
MUJOCO_GL=egl PYOPENGL_PLATFORM=egl \
  pytest -m gpu tests/envs/test_libero_control_freq.py::test_control_freq_reaches_real_robosuite_sim
```

## Checklist

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -38,17 +38,17 @@ OpenTau is Tensor's open-source PyTorch training toolchain for vision-language-a
 
 ## Environment & install
 
-Dependency management is **`uv` (>= 0.8.4) only** — `pyproject.toml`/`uv.lock` are authoritative; do not edit `requirements.txt` (none exists) or call `pip install` against the env. The `>= 0.8.4` floor is enforced by `required-version` in `[tool.uv]` and is needed because `[tool.uv].extra-build-dependencies` (which pins `cmake<4` inside `egl-probe`'s PEP 517 build isolation) is only honored by uv 0.8.4+.
+Dependency management is **`uv` (>= 0.8.4) only** — `pyproject.toml`/`uv.lock` are authoritative; do not edit `requirements.txt` (none exists) or call `pip install` against the env. The `>= 0.8.4` floor is enforced by `required-version` in `[tool.uv]` so the team resolves against the same lockfile format and `[tool.uv]` feature set; bump it only alongside a `uv lock` refresh.
 
 ```bash
 uv sync --extra dev --extra libero          # standard dev setup (matches CI)
-uv sync --all-extras                         # everything (note: libero + urdf are mutually exclusive — see [tool.uv].conflicts)
+uv sync --all-extras                         # everything (libero + urdf now co-install on numpy 2.x)
 source .venv/bin/activate
 ```
 
 Re-run `uv sync` whenever `pyproject.toml`/`uv.lock` change. Add deps with `uv add <pkg>`; lock with `uv lock`.
 
-Installable extras: `dev` (pre-commit, sphinx, pytest), `libero` (sim env — pulls a forked LIBERO from `shuheng-liu/LIBERO`, pins numpy<2), `urdf` (rerun ≥0.28, requires numpy 2.x — incompatible with `libero`), `trt` (TensorRT, Linux/Win x86_64 only).
+Installable extras: `dev` (pre-commit, sphinx, pytest), `libero` (sim env — pulls a forked LIBERO from `shuheng-liu/LIBERO`, runs on numpy 2.x + gymnasium), `urdf` (rerun ≥0.28, numpy 2.x), `trt` (TensorRT, Linux/Win x86_64 only).
 
 ## Common commands
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -38,7 +38,7 @@ OpenTau is Tensor's open-source PyTorch training toolchain for vision-language-a
 
 ## Environment & install
 
-Dependency management is **`uv` (>= 0.8.4) only** — `pyproject.toml`/`uv.lock` are authoritative; do not edit `requirements.txt` (none exists) or call `pip install` against the env. The `>= 0.8.4` floor is enforced by `required-version` in `[tool.uv]` so the team resolves against the same lockfile format and `[tool.uv]` feature set; bump it only alongside a `uv lock` refresh.
+Dependency management is **`uv` (>= 0.8.4) only** — `pyproject.toml`/`uv.lock` are authoritative; do not edit `requirements.txt` (none exists) or call `pip install` against the env. The `>= 0.8.4` floor is enforced by `required-version` in `[tool.uv]` and is needed because `[tool.uv].extra-build-dependencies` (which injects `cmake` into `egl-probe`'s PEP 517 build isolation so its sdist builds without a system cmake, on CMake 4) is only honored by uv 0.8.4+.
 
 ```bash
 uv sync --extra dev --extra libero          # standard dev setup (matches CI)
--- a/docs/source/contributing.rst
+++ b/docs/source/contributing.rst
@@ -117,9 +117,9 @@ Follow these steps to start contributing:
 4. for development, we advise to use a tool like ``uv`` (>= 0.8.4) instead of just ``pip`` to easily track our dependencies.
    Follow the instructions to `install uv <https://docs.astral.sh/uv/getting-started/installation/#installation-methods>`_ if you don't have it already.
    OpenTau requires ``uv >= 0.8.4`` (enforced by ``required-version`` in ``[tool.uv]``)
-   so everyone resolves against the same lockfile format and ``[tool.uv]`` feature set;
-   older ``uv`` is rejected loudly rather than silently producing a divergent lock.
-   Upgrade with ``uv self update`` if needed.
+   because ``[tool.uv].extra-build-dependencies`` is only honored from that version;
+   it injects ``cmake`` into ``egl-probe``'s build isolation so the ``libero`` extra
+   builds without a system ``cmake`` (on CMake 4). Upgrade with ``uv self update`` if needed.
    To develop on OpenTau, you will at least need to install the ``dev`` extras dependencies along with the core library:
 
    using ``uv``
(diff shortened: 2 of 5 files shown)


shuheng-liu (commented):
## Review: build(deps) numpy 2.x / gymnasium / CMake 4 LIBERO

Nicely scoped, deps-and-docs-only change with an excellent write-up. I verified the headline claims against the regenerated `uv.lock`: legacy `gym` is gone, `numpy` collapses to a single `2.2.6`, and `egl-probe` is pinned to an immutable tag **and** commit (`v1.0.1-cmake4#e387504...`), so the source is reproducible. A few notes below — only the first is substantive.

### What's good
- **Correct mechanism for the egl-probe override.** Adding `egl-probe` as an explicit dep in the `libero` extra is what makes `[tool.uv.sources]` actually apply to it (uv only honors sources for declared deps, and it was previously only transitive via `robomimic`). Good catch.
- **Dropping `gym` is safe.** All first-party code already imports `import gymnasium as gym` (`envs/factory.py`, `envs/utils.py`, `envs/libero.py`, `scripts/eval.py`); legacy `gym` is only shimmed by `utils/monkey_patch.py`. Nothing depends on the real `gym` package.
- Docs (`CLAUDE.md`, `installation.rst`, `contributing.rst`) were updated consistently with the config changes.

### Findings

**1. (Medium) The `libero` build now depends on a host-provided `cmake` + C/C++ toolchain.**
Removing `cmake<4` from the extra *and* `extra-build-dependencies = { egl-probe = ["cmake<4"] }` means cmake is no longer injected into egl_probe's PEP 517 build isolation. The fork has no `pyproject.toml` — it's a legacy `setup.py` with `install_requires=[]` that shells out to a system `cmake` ("CMake must be installed"). So `uv sync --extra libero` now relies entirely on a system cmake (≥3.5, which the fork's bumped floor satisfies) plus a compiler being present, rather than being self-contained via the pip `cmake` wheel.
- CI is fine: `cpu_test` runs on `ubuntu-latest` (cmake/gcc preinstalled) and `gpu_test`'s bare CUDA container explicitly `apt-get install ... cmake build-essential`.
- The regression is for **local dev**: a contributor on a machine without a system cmake (e.g. a fresh macOS) will now hit egl_probe's "CMake must be installed", where before the build was self-contained.
- Suggestion: keep cmake as a *build-only* dep without the `<4` cap, e.g. `extra-build-dependencies = { egl-probe = ["cmake"] }` (or add unpinned `cmake` to the extra). That preserves self-containedness while still building under CMake 4. If requiring a system cmake is intentional, the installation docs should call that out — they currently don't mention the new requirement.

**2. (Low) The dependency change isn't actually exercised by CI yet.** "CPU Tests" is reporting `skipped` (with the "CPU Tests Status" shim succeeding) — looks like the draft-PR gate. The job that runs `uv sync --extra dev --extra libero` (and thus builds egl_probe) won't run until this is marked ready-for-review. Worth flipping out of draft once to get a real `cpu_test` that builds egl_probe on the CI image, rather than relying solely on the local run.

**3. (Nit) `required-version = ">=0.8.4"` no longer has a concrete driver.** The original reason (`extra-build-dependencies` needs 0.8.4) is gone. The reworded "same lockfile format / feature set" rationale is reasonable and keeping the floor avoids churn — just noting the justification is now softer.

**4. (Nit, out of scope) `monkey_patch.gym_to_gymnasium` shim** may now be partly redundant given LIBERO runs on gymnasium. Don't touch it here — flagging for a possible future cleanup.

**5. (Observation, pre-existing) The `libero` git source still tracks `branch = "master"`** (mutable) rather than a tag/commit. The lock captures a commit, but a future `uv lock` could silently advance LIBERO. Given the fork is now load-bearing for the numpy-2.x story, consider pinning it to a tag like the egl-probe source.

Overall: solid, low-risk cleanup. Item 1 (system-cmake shift) is the only thing I'd want resolved or explicitly documented before merge.

---

shuheng-liu:
Thanks for the careful review — addressed in `62aa311`.

**1 (system-cmake regression) — fixed.** You're right, and it was worse than just self-containedness: dropping the build-isolation cmake meant my "builds under CMake 4.2.1" claim was actually wrong — the build used the *host* system cmake (3.28 on the machine I tested), so the fork's CMake-4 path was never exercised. Re-added `extra-build-dependencies = { egl-probe = ["cmake"] }` (unpinned, per your suggestion). This both restores the self-contained build (no system cmake needed; a C/C++ toolchain still is) and injects cmake 4.2.1 into egl_probe's build isolation, so the build now genuinely runs under CMake 4 — finally exercising what the fork is for.

**3 (required-version driver) — resolved by the above.** With `extra-build-dependencies` back, the `>= 0.8.4` floor has its concrete driver again; the docs (`CLAUDE.md`, `installation.rst`, `contributing.rst`) now state that reason rather than the vaguer "lockfile baseline" wording.

**2 (CI didn't exercise it) — already covered.** The full `CPU Tests` job did run to completion (~11 min) on an earlier push and built egl_probe via `uv sync --extra dev --extra libero`; the `skipped`/shim entries were superseded duplicate runs. It re-runs on this push, now through the injected-cmake path.

**5 (`libero` on `branch = "master"`) — pushing back, with evidence.** A plain `uv lock` does *not* silently advance a git-branch dep: after the pyproject edits, re-locking kept libero at the previously-pinned `45c5a01`; I had to run `uv lock --upgrade-package libero` explicitly to move it to the merged `84c2dae`. So the lock stays reproducible (pinned commit) and advancing it is always an explicit, reviewable step — no silent-drift risk. I'd prefer keeping `branch = "master"` so routine LIBERO fixes are a one-line `--upgrade` rather than a manual SHA bump, but happy to freeze it to a rev/tag if you'd rather.

**4 (gym shim) — agreed, leaving it here.** Filed as a follow-up to check whether `utils/monkey_patch.py`'s gym→gymnasium shim is still needed now that nothing depends on legacy `gym`.

Note: `uv.lock` is unchanged by `62aa311` — `extra-build-dependencies` is a build-time setting and isn't recorded in the lock.

shuheng-liu (commented):
## Re-review of `62aa311` — all findings addressed, LGTM

Re-reviewed against the new commit. Everything I flagged is resolved, and I confirmed the substance against the tree and CI.

**1 (system-cmake regression) — fixed, and the fix is better than I realized.** `extra-build-dependencies = { egl-probe = ["cmake"] }` is back in `[tool.uv]`, so the build is self-contained again (no system cmake needed — a C/C++ toolchain still is). Your catch that the original "builds under CMake 4.2.1" claim was *wrong* (the host's 3.28 was silently used, so the fork's CMake-4 path was never exercised) is the important part: now the injected cmake is genuinely 4.x against the fork's raised 3.5 floor, so the build actually tests what the fork exists for. Nice.

**3 (required-version driver) — resolved.** The `>= 0.8.4` floor has its concrete `extra-build-dependencies` driver back, and `CLAUDE.md` / `installation.rst` / `contributing.rst` now state that reason. The pyproject comment is accurate.

**2 (CI coverage) — confirmed green.** The full `CPU Tests` job ran to completion on this commit (job 78543790109, ~15 min, `success`) and went through `uv sync --extra dev --extra libero` — so egl_probe now builds via the injected-cmake path in CI, not just locally. The `skipped` entries are the superseded duplicate runs as you noted.

**5 (`libero` on `branch = "master"`) — you're right, withdrawing.** I accept the correction: a plain `uv lock` keeps a git-branch dep at its pinned commit and only `--upgrade-package` advances it, so there's no silent-drift risk and the lock stays reproducible. Keeping `branch = "master"` for one-line `--upgrade` ergonomics is a reasonable trade — no change needed.

**4 (gym shim) — agreed as a follow-up.** Good to leave out of this deps-only PR.

I also confirmed `uv.lock` is unchanged by `62aa311`, consistent with `extra-build-dependencies` being build-time-only.

No remaining blockers from my side. Once it's flipped out of draft this looks ready to merge. 👍

---

PR #358, opened 2026-05-30
feat(pi07_paligemma): custom CUDA block-causal flash attention

WilliamYue37:
## What this does

Adds a new `attention_implementation="flash_cuda"` backend for **pi07_paligemma** (both `low_level` and `high_level_planner`): a **custom CUDA flash attention** that handles the **block-causal mask in-kernel**, with **no external libraries** (no `flash-attn`, no `flex_attention`). (⚡️ Performance / 🗃️ Feature)

How it works:
- **In-kernel block-causal masking, no S×S materialization.** `make_att_block_ids` produces a compact per-token representation `(q_blk, k_blk, q_valid, k_valid)` that *provably reproduces* `make_att_2d_masks`: `attend(i,j) == q_valid[i] & k_valid[j] & (k_blk[j] <= q_blk[i])`. Cross-attention columns use an `INT32_MIN` sentinel (always attended). The kernel rebuilds the mask on the fly and exploits the monotone block-ids for block-causal tile early-exit. The dense `(B,S,S)` mask is **never built**.
- **Kernels** (`src/opentau/policies/flash_attn_cuda/_csrc/flash_blockmask.cu`):
  - fp32: a warp-per-row online-softmax reference (exact).
  - fp16/bf16: a **Tensor Core (WMMA)** forward (both matmuls on tensor cores, fp32 accumulation).
  - fp16/bf16: a **Tensor Core (WMMA) backward** — dQ kernel (parallel over query slabs) and dK/dV kernel (parallel over key slabs, looping the GQA query-head group, **no atomics**). All five backward matmuls (recompute `S`, `dP=dO Vᵀ`, `dV=Pᵀ dO`, `dK=dSᵀ Q`, `dQ=dS K`) run on tensor cores with fp32 accumulation; warps/block auto-sized to the device's opt-in shared memory. The fp32 warp-per-row backward is retained as the exactness oracle.
- **No fallback.** Selecting `flash_cuda` always uses the kernel; if it cannot be JIT-compiled (no CUDA/compiler) the forward **raises** rather than silently degrading to sdpa/eager. Importing on CPU never triggers `nvcc` (lazy compile), so the CPU CI subset is unaffected; `eager`/`sdpa`/`fa2` are unchanged and flash is opt-in.
- Wired through `PaliGemmaWithExpertModel.forward`/`_run_layer` (compact block-ids threaded alongside the existing dense-mask path; KV-cache / cross-attention expert pass supported). `eager`/`sdpa`/`fa2` behavior is unchanged — flash is fully opt-in.

## Results (RTX 3090, head_dim=256, MQA, bf16)

**Memory.** The fair comparison is *at equal gradient-checkpointing*, since realistic pi07_paligemma training uses `gradient_checkpointing=True` (which already frees eager's per-layer `(B,H,S,S)` scores by recomputing them in backward). For an 18-layer attention stack (S=1024), fwd+bwd peak:

| config | peak |
|---|---|
| eager, no checkpointing | 2.35 GB |
| **eager, +checkpointing** (realistic baseline) | **0.50 GB** |
| flash, no checkpointing | 0.39 GB |
| **flash, +checkpointing** | **0.21 GB** |

So the honest, apples-to-apples win is **flash+ckpt vs eager+ckpt ≈ 2.4×** — flash never materializes `S×S` even during recompute, and it composes *with* checkpointing. (The single-op tables in the benchmark show 20–42× / 7–21×, but those compare against **non-checkpointed** eager and are not the training-realistic baseline — see the `stacked-layer` section of the benchmark.) Separately, even checkpointed eager must materialize one layer's `(B,H,S,S)` transiently during recompute, so it **OOMs at B4·S=16384–24576** (would need 34–77 GB) where flash runs in 1.7–3.3 GB.

**Speed (honest).** Forward is ~3.4× slower than the cuBLAS/cuDNN-backed eager path at head_dim=256. The Tensor Core backward brought fwd+bwd from ~9× → **~6× slower than eager** (~1.4× faster than the previous scalar backward) — see the A100 numbers below. It is still behind cuDNN `sdpa`. Fully closing the gap needs cp.async double-buffering + register-resident `mma.sync` accumulation (to remove the shared-memory round-trips) and avoiding the 2× score recompute in the backward.

## How it was tested

- Added `tests/policies/test_flash_attn_cuda.py` (139 tests):
  - **CPU:** compact block-ids reproduce `make_att_2d_masks` **bit-for-bit** (incl. the cross-attention branch).
  - **GPU forward (fp32):** matches a dense reference to `atol/rtol=1e-4` across causal / prefix-LM / multi-block / bidirectional masks, head_dim ∈ {64,128,256}, MQA+GQA, non-square cross-attention, and padding.
  - **GPU Tensor Core (WMMA) fwd:** matches `sdpa` (bf16+fp16) across all the above combinations, incl. the non-square cross-attention layout.
  - **GPU Tensor Core (WMMA) backward:** dQ/dK/dV match the fp32 autograd reference (bf16+fp16) across all the above combinations + non-square cross-attention; fp32 backward also checked vs autograd (`1e-3`).
  - **GPU integration:** `flash_attention_forward` matches `sdpa_attention_forward` through the real `PaliGemmaWithExpertModel` methods at the real head config.
  - **GPU perf:** asserts flash peak attention memory < eager.
- Verified seeded flash fwd+bwd is **bit-identical across runs** (output + dQ/dK/dV) — deterministic.
- **No regressions:** existing CPU policy suites for pi0 / pi05 / pi05_mem / pi07_paligemma pass (132 tests) — the `eager`/`sdpa` paths are untouched.
- `src/opentau/scripts/benchmark_flash_attn.py` produces the latency + per-op memory tables, the **checkpointed stacked-layer memory comparison** above, and (`--oom-demo`) the eager-OOM-vs-flash demo.

### A100 validation (8×A100-80GB, sm_80, slurm node gpu-387)

Validated on the real training hardware (with the Tensor Core backward):
- **Correctness:** all **134 GPU tests pass on A100 (sm_80)** — forward + backward correct on the training arch, not just my dev 3090 (sm_86).
- **8-GPU distributed (DDP):** an 18-layer attention+o_proj stack, per-rank B2·S1024·D256 bf16, fwd+bwd+all-reduce across all 8 ranks. **flash is distributed-correct & deterministic (identical across all 8 ranks, no NCCL hang):**

  | config | step ms (scalar bwd → **TC bwd**) | per-rank peak GB |
  |---|---|---|
  | eager | 83 | 2.76 |
  | eager+ckpt | 110 | 0.90 |
  | flash | 764 → **536** | 0.89 |
  | flash+ckpt | 869 → **641** | 0.71 |

**Honest training takeaway:** the Tensor Core backward gave a real **~1.4× step speedup** (flash+ckpt 869→641 ms) — but flash is **still ~6× slower than eager** (and ~6× slower than cuDNN `sdpa`) at S=1024/head_dim=256, down from ~9×. The realistic per-rank memory win (flash+ckpt vs eager+ckpt) is **~1.3×** — the attention-matrix savings are diluted by params/DDP buffers (and would dilute further with the MLP), growing only at long sequence. **As-is, flash_cuda is still not faster than the cuDNN `sdpa` baseline for this config**; what it provides is correctness, in-kernel block-mask handling (no dense-mask materialization), distributed-safety, and OOM-avoidance at long context. Closing the speed gap needs cp.async pipelining + register-resident accumulation.

**Still pending:** the nightly **model** regression suite (full pi07_paligemma training) — this validation covered the attention op + a distributed attention-stack benchmark, not an end-to-end training run.

### Long-context regime (A100-80GB, gpu-387) — forward, B1·H8·Hkv1·D256, prefix-LM bf16

**SDPA backend probe (block-causal bool mask).** PyTorch's fused **FlashAttention backend rejects** any non-`is_causal` `attn_mask` — so it cannot flash a block-causal mask (confirming the mechanism). But the **memory-efficient backend *does* accept it** (math/eager also accepts). So SDPA is *not* limited to eager for block-causal masks on this build (square self-attention case).

**Forward sweep vs sequence length** (`benchmark_flash_attn.py --long-context`):

| S | dense mask | eager | sdpa | flash_cuda |
|---|---|---|---|---|
| 8192 | 0.07 GB | 41 ms / 4.8 GB | **8.6 ms / 0.5 GB** | 137 ms / 0.2 GB |
| 16384 | 0.27 GB | 155 ms / 18.5 GB | **35 ms / 1.5 GB** | 526 ms / 0.4 GB |
| 32768 | 1.07 GB | 647 ms / 72.4 GB | **180 ms / 5.1 GB** | 2441 ms / 1.4 GB |
| 65536 | 4.29 GB | OOM | **912 ms / 18.9 GB** | 10731 ms / 4.9 GB |
| 98304 | 9.66 GB | OOM | **2250 ms / 41.2 GB** | 22871 ms / 10.6 GB |
| 131072 | 17.2 GB | OOM | **OOM** | **43640 ms / 18.4 GB** |

**Honest takeaway.** Three regimes:
- **Up to S≈98k:** SDPA's memory-efficient backend handles the block-causal mask and is **~15–25× faster than flash_cuda** (and lower-memory than eager). flash_cuda is *not* the right choice here.
- **eager** OOMs at S≥65536 (O(S²) fp32 scores).
- **S≥131072:** SDPA OOMs because it still needs the dense **O(S²) mask** (17 GB at S=131072); flash_cuda uses compact block-ids (O(S)) and is the **only backend that runs** — but ~20× slower than SDPA was. This extreme-S regime is flash_cuda's genuine niche.

**Caveat on the `sdpa ≈ eager` observation.** Above is the *square* self-attention mask, where SDPA's mem-efficient backend works. The action-expert pass uses a **non-square cross-attention** mask `(B, N, n_cross+N)`; mem-efficient may reject that shape and fall back to the math (≈eager) backend, which would explain `sdpa ≈ eager` in real training. The benchmark now probes this (`--long-context`); worth confirming which backend your actual run selects for the expert pass.

```bash
# CPU (mask-equivalence, no GPU needed)
pytest -sx tests/policies/test_flash_attn_cuda.py -m "not gpu"
```
```bash
# GPU equivalence + memory tests (compiles the kernel on first run; needs ninja + nvcc)
pytest -sx tests/policies/test_flash_attn_cuda.py -m "gpu"
```
```bash
# Latency + memory benchmark (incl. checkpointed stacked-layer comparison) + OOM demo
python -m opentau.scripts.benchmark_flash_attn --oom-demo
```
```bash
# Use it: any pi07_paligemma config
opentau-train --config_path=<cfg> --policy.attention_implementation=flash_cuda
```

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -161,6 +161,11 @@ extra-build-dependencies = { egl-probe = ["cmake"] }
 [tool.setuptools.packages.find]
 where = ["src"]
 
+[tool.setuptools.package-data]
+# Ship the custom flash-attention CUDA source so the kernel can be JIT-compiled
+# from an installed wheel (not just an editable/source checkout).
+"opentau.policies.flash_attn_cuda" = ["_csrc/*.cu", "_csrc/*.cuh", "_csrc/*.cpp", "_csrc/*.h"]
+
 [tool.ruff]
 line-length = 110
 target-version = "py310"
--- a/src/opentau/policies/flash_attn_cuda/__init__.py
+++ b/src/opentau/policies/flash_attn_cuda/__init__.py
@@ -1,0 +1,171 @@
+# Copyright 2026 Tensor Auto Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Custom CUDA FlashAttention with in-kernel block-causal masking.
+
+This package provides a dependency-free (no ``flash-attn``, no ``flex_attention``)
+fused attention forward + backward whose block-causal mask is reconstructed
+*inside the CUDA kernel* from a compact per-token block-id representation. The
+dense ``(B, S, S)`` attention mask is therefore never materialized, which is both
+the memory win over the eager backend and a direct way to "handle the block
+causal attention masks in the code".
+
+Public API:
+  - :func:`make_att_block_ids` -- build the compact ``(q_blk, k_blk, q_valid,
+    k_valid)`` representation from the same ``pad_masks`` / ``att_masks`` that
+    :func:`make_att_2d_masks` consumes. Provably reproduces the dense mask:
+    ``attend(i, j) == q_valid[i] & k_valid[j] & (k_blk[j] <= q_blk[i])``.
+  - :func:`flash_attn_blockmask` -- autograd-aware attention given the compact
+    representation.
+  - :func:`is_available` -- whether the CUDA kernel compiled (else fall back).
+
+Masking convention matches ``make_att_2d_masks``: ``True`` = attend. Cross-
+attention key columns get ``k_blk = INT32_MIN`` so they are always attended
+(subject to padding).
+"""
+
+from __future__ import annotations
+
+import torch
+
+from ._loader import get_extension, is_available, load_error
+
+__all__ = [
+    "INT_BLK_MIN",
+    "flash_attn_blockmask",
+    "is_available",
+    "load_error",
+    "make_att_block_ids",
+]
+
+# Sentinel block-id for cross-attention columns: always <= any real (cumsum)
+# block-id, so those columns are unconditionally attended (gated only by
+# padding). Must match the int32 INT_MIN the kernel compares against.
+INT_BLK_MIN = -(2**31)
+
+
+def make_att_block_ids(
+    pad_masks: torch.Tensor,
+    att_masks: torch.Tensor,
+    n_cross_att_tokens: int | None = None,
+    cross_att_pad_masks: torch.Tensor | None = None,
+) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor]:
+    """Compact block-id representation equivalent to ``make_att_2d_masks``.
+
+    Mirrors ``make_att_2d_masks`` exactly but returns four 1-D-per-token tensors
+    instead of a dense 2-D mask. The kernel reconstructs the mask via
+    ``attend(i, j) == q_valid[i] & k_valid[j] & (k_blk[j] <= q_blk[i])``.
+
+    Args:
+        pad_masks: bool ``(B, N)``, ``True`` where the token is real (not padding).
+        att_masks: int/bool ``(B, N)``, ``1`` opens a new attention block, ``0``
+            continues the current block (the same convention as
+            ``make_att_2d_masks``).
+        n_cross_att_tokens: if set, prepend that many cross-attention key columns
+            (always attended), matching the cross-attention branch of
+            ``make_att_2d_masks``.
+        cross_att_pad_masks: bool ``(B, n_cross_att_tokens)`` padding for the
+            cross columns. Required iff ``n_cross_att_tokens`` is set.
+
+    Returns:
+        ``(q_blk, k_blk, q_valid, k_valid)`` where ``q_blk`` is int32 ``(B, N)``,
+        ``k_blk`` is int32 ``(B, N)`` or ``(B, n_cross_att_tokens + N)``,
+        ``q_valid`` is bool ``(B, N)``, ``k_valid`` matches ``k_blk``'s length.
+    """
+    if att_masks.ndim != 2:
+        raise ValueError(f"att_masks must be 2D, got {att_masks.ndim}")
+    if pad_masks.ndim != 2:
+        raise ValueError(f"pad_masks must be 2D, got {pad_masks.ndim}")
+
+    cumsum = torch.cumsum(att_masks.to(torch.int32), dim=1).to(torch.int32)
+    q_blk = cumsum
+    q_valid = pad_masks.to(torch.bool)
+
+    if n_cross_att_tokens is None:
+        return q_blk, cumsum, q_valid, q_valid
+
+    assert cross_att_pad_masks is not None, (
+        "cross_att_pad_masks must be provided if n_cross_att_tokens is provided"
+    )
+    assert cross_att_pad_masks.shape == (att_masks.size(0), n_cross_att_tokens), (
+        "cross_att_pad_masks must have shape (batch_size, n_cross_att_tokens)"
+    )
+
+    b = att_masks.size(0)
+    cross_blk = torch.full((b, n_cross_att_tokens), INT_BLK_MIN, dtype=torch.int32, device=att_masks.device)
+    k_blk = torch.cat([cross_blk, cumsum], dim=1)
+    k_valid = torch.cat([cross_att_pad_masks.to(torch.bool), q_valid], dim=1)
+    return q_blk, k_blk, q_valid, k_valid
+
+
+def _prep(t: torch.Tensor, dtype: torch.dtype) -> torch.Tensor:
+    return t.to(dtype).contiguous()
+
+
+class _FlashAttnBlockMask(torch.autograd.Function):
+    """Autograd wrapper around the custom CUDA fwd/bwd kernels."""
+
+    @staticmethod
+    def forward(ctx, q, k, v, q_blk, k_blk, q_valid, k_valid, scale):  # noqa: ANN001
+        ext = get_extension()
+        if ext is None:
+            raise RuntimeError(f"flash_cuda kernel unavailable: {load_error()}")
+        q = q.contiguous()
+        k = k.contiguous()
+        v = v.contiguous()
+        q_blk = _prep(q_blk, torch.int32)
+        k_blk = _prep(k_blk, torch.int32)
+        q_valid = _prep(q_valid, torch.bool)
+        k_valid = _prep(k_valid, torch.bool)
+        out, lse = ext.flash_fwd(q, k, v, q_blk, k_blk, q_valid, k_valid, float(scale))
+        ctx.save_for_backward(q, k, v, out, lse, q_blk, k_blk, q_valid, k_valid)
+        ctx.scale = float(scale)
+        return out
+
+    @staticmethod
+    def backward(ctx, grad_out):  # noqa: ANN001
+        ext = get_extension()
+        q, k, v, out, lse, q_blk, k_blk, q_valid, k_valid = ctx.saved_tensors
+        dq, dk, dv = ext.flash_bwd(
+            grad_out.contiguous(), q, k, v, out, lse, q_blk, k_blk, q_valid, k_valid, ctx.scale
+        )
+        return dq, dk, dv, None, None, None, None, None
+
+
+def flash_attn_blockmask(
+    q: torch.Tensor,
+    k: torch.Tensor,
+    v: torch.Tensor,
+    q_blk: torch.Tensor,
+    k_blk: torch.Tensor,
+    q_valid: torch.Tensor,
+    k_valid: torch.Tensor,
+    scale: float,
+) -> torch.Tensor:
+    """Block-causal flash attention.
+
+    Args:
+        q: ``(B, Sq, H, D)`` queries.
+        k: ``(B, Sk, Hkv, D)`` keys (GQA/MQA: ``H % Hkv == 0``).
+        v: ``(B, Sk, Hkv, D)`` values.
+        q_blk: int32 ``(B, Sq)`` query block-ids.
+        k_blk: int32 ``(B, Sk)`` key block-ids (cross columns use ``INT_BLK_MIN``).
+        q_valid: bool ``(B, Sq)`` query padding mask (``True`` = real token).
+        k_valid: bool ``(B, Sk)`` key padding mask.
+        scale: softmax scale (typically ``head_dim ** -0.5``).
+
+    Returns:
+        ``(B, Sq, H, D)`` attention output, same dtype as ``q``.
+    """
+    return _FlashAttnBlockMask.apply(q, k, v, q_blk, k_blk, q_valid, k_valid, scale)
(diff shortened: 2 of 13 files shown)


shuheng-liu (commented):
## Review: custom CUDA block-causal flash attention

Did a thorough pass over the kernel, the Python wiring, the mask-equivalence contract, and the tests. Overall this is a careful, well-documented piece of work — the masking contract is correct, the integration is clean and fully opt-in, and the perf claims are disclosed honestly. No confirmed correctness bugs. The remaining items are validation gaps and a couple of footguns, not blockers in the kernel itself.

I don't have a GPU in this environment, so the GPU pytests were not re-run here — I verified everything that's checkable on CPU and by inspection (noted below).

### What I verified ✅

- **Mask equivalence is exact.** Independently reproduced `make_att_block_ids` vs `make_att_2d_masks` over 400 randomized cases (all four patterns + the cross-attention branch) — bit-for-bit identical. The contract `attend(i,j) == q_valid[i] & k_valid[j] & (k_blk[j] <= q_blk[i])` holds, including the `INT_MIN` cross-column sentinel.
- **Early-exit invariant holds.** The key-axis block-ids stay non-decreasing even with `INT_MIN` cross columns prepended, so the block-causal tile early-exit (`break` in fwd/dQ, `continue` in dK/dV) is sound.
- **Masking is consistent across all four kernels** (fp32 fwd, WMMA fwd, dQ, dK/dV) and matches the Python contract; cross columns (`kb = INT_MIN`) are correctly always-attended in every path.
- **Scale matches eager** (`head_dim ** -0.5`, vs `paligemma_with_expert.py:745`).
- **Determinism reasoning checks out.** No atomics in numeric accumulation (only integer `atomicMax`/`atomicMin` for early-exit bounds, which are order-independent); fixed reduction order → the bit-identical-across-runs claim is credible.
- **Output is fully initialized** — every `dQ`/`dK`/`dV`/`O` element is written exactly once; padded/masked rows write 0 consistently with the reference.
- **Fallback + CPU safety.** `flash_cuda_active()` gates on `is_available()`, JIT is lazy so importing on CPU never invokes `nvcc`, and the CPU equivalence tests carry no `gpu` marker (they ran green in CI and reproduce here).

### Should address before merge

1. **Nightly regression + A100 benchmarks not yet run.** You flagged this yourself, and the checklist requires it for a policy-change PR. This is the actual gating item. `mergeable_state` is currently `blocked`.

2. **GQA backward (`hkv > 1`) is untested.** `test_backward_matches_autograd_reference` only uses `hkv=1`. The forward covers `hkv ∈ {1,2}`, and MQA does exercise the group-accumulation loop, but the dK/dV kernel's multi-KV-head write path (`hk = blockIdx.y` for `hk > 0`) is never hit in the backward. Low risk, but it's a real hole in the "no-atomics grouped gradient" claim — please add `hkv=2` to that parametrization.

3. **bf16/fp16 backward numerics untested.** Backward tests are fp32-only. The bwd kernels are dtype-templated (fp32 accumulate, cast on store), so the math is covered, but the bf16/fp16 store rounding of grads isn't validated against a reference. One extra case would close it.

### Footguns / clarity

4. **`flash_cuda` silently degrades to sdpa for pi0 / pi05 / pi05_mem.** The validator now accepting `"flash_cuda"` lives in the shared `PaliGemmaWithExpertConfig`, so every PaliGemma-based policy accepts it — but only `pi07_paligemma` builds block-ids via `build_attention_inputs`. The others pass `attention_block_ids=None`, so `_run_layer` routes them to sdpa with no warning. Setting `flash_cuda` on pi05 is a silent no-op. It's benign (sdpa is correct), but worth either a warning when the backend is selected on a policy that doesn't build block-ids, or scoping the validator.

5. **"Graceful fallback" only covers compile failures, not launch failures.** At `head_dim=256` the WMMA forward needs ~93 KB dynamic shared memory. That opt-in succeeds on sm_80/sm_86 (your RTX 3090) but exceeds the 64 KB cap on older arches (sm_70/sm_75), and the `cudaFuncSetAttribute` return isn't checked. On such a box the kernel would throw at runtime via `C10_CUDA_KERNEL_LAUNCH_CHECK` rather than falling back to sdpa — the fallback path only catches *build* errors. Fine for the stated RTX 3090 / A100 target, but the claim is a bit broader than the implementation.

6. **Scope: this lands on legacy `pi07_paligemma`, not current `pi07/`.** Per CLAUDE.md, `pi07_paligemma` is the legacy PaliGemma variant kept for old checkpoints; current π0.7 (Gemma 3 + `gemma3_with_expert.py`) doesn't get this backend. Probably intentional given the PR title, but it means the headline memory win doesn't reach the current default π0.7 — worth confirming that's the intent and noting follow-up.

### Performance framing

The honesty here is appreciated. To make the tradeoff explicit for anyone reading: a **~15× slower fwd+bwd** means this is **not** a general training speedup — it's strictly a memory-vs-throughput trade, only worth enabling when you're memory-bound and eager/sdpa would otherwise OOM (longer context / larger batch). For runs that already fit, sdpa remains the better choice. Might be worth stating that as the recommended-usage guidance in the config docstring.

### Nits

- The fp32 path is described as "exact" but uses `__expf` (fast intrinsic, ~2 ulp) — accurate enough for the 1e-4 tolerance, just not literally exact.
- The `github-code-quality` bot comment claiming `_load_attempted` is unused (`_loader.py:81`) is a **false positive** — it's the memoization guard used in `get_extension()`. Safe to dismiss.

Nice work overall. The kernel correctness is solid; closing the GQA-backward gap and getting the nightly/A100 runs in are the main things between here and merge.

---

WilliamYue37:
@claude fix

shuheng-liu (commented):
## Re-review at `3c34d34`

Re-reviewed end to end after the push from `1582f2e4` → `3c34d34` (+~700 lines: Tensor Core backward, fallback removal, config scoping, A100/DDP validation). This is a substantially stronger PR than at my first pass. I traced the new code directly rather than relying on the bot summaries — no correctness blockers found. The open items are now strategic + process, not the kernel.

### Prior findings — all addressed ✅

1. **A100 / distributed validation** — done: 134 GPU tests pass on A100 (sm_80) and an 8-GPU DDP attention-stack run is bit-identical across ranks with no NCCL hang. (The end-to-end nightly *model* regression is still outstanding — see below.)
2. **GQA (`hkv>1`) backward** — closed: `test_wmma_backward_matches_reference` now parametrizes `hkv ∈ {1,2}`, exercising the dK/dV `blockIdx.y` per-KV-head write path.
3. **bf16/fp16 backward numerics** — closed: same test parametrizes `dtype ∈ {fp16, bf16}` vs the fp32 autograd reference, plus a non-square cross-attention backward test.
4. **Silent sdpa degradation for pi0/pi05/pi05_mem** — fixed: rejected at config-load in each `__post_init__` with an actionable message; `get_attention_interface` raises as a defense-in-depth backstop.
5. **Launch-failure footgun** — fixed: the WMMA forward now checks `cudaDevAttrMaxSharedMemoryPerBlockOptin` and `TORCH_CHECK`s before launch, so sub-sm_80 arches get an actionable "use sdpa" error instead of an opaque launch failure.
6. **Scope (legacy pi07_paligemma)** — acknowledged as intentional.

### New code reviewed — Tensor Core backward (correct ✅)

I verified the two new WMMA backward kernels in `flash_blockmask.cu`:
- **dQ kernel** (`flash_bwd_dq_wmma_kernel`): `S=QKᵀ`/`dP=dOVᵀ` via the col-major-K trick, then `dQ += dS@K` — matmul fragment layouts and the `*scale`-on-store are correct.
- **dK/dV kernel** (`flash_bwd_dkv_wmma_kernel`): the `col_major`-load-as-transpose for `P`/`dS` to compute `dV = Pᵀ@dO` and `dK = dSᵀ@Q` is correct, and the GQA group is accumulated in-register with no atomics (so the determinism claim holds).
- **Shared memory**: both `bwd_dq_smem`/`bwd_dkv_smem` size helpers are byte-exact with the kernels' pointer arithmetic (checked each region + 4-byte alignment of the float block) — no overrun.
- **`pick_nw` auto-sizing**: at D=256 it correctly selects different `nw` on sm_86 (RTX 3090) vs sm_80 (A100) to fit each device's opt-in cap — both code paths are exercised by the dual-machine validation. Nice touch.
- The **"no fallback, raise loudly"** design is implemented consistently (loader docstring, `flash_cuda_active`, autograd Function, `get_attention_interface`, config rejections). I think this is the right call over silent sdpa substitution.

Mask-equivalence contract is unchanged since my first pass (still bit-for-bit), and CI is green at `3c34d34`.

### The one thing worth a hard look before merging: is the win worth the surface area?

Credit for unusually honest benchmarking — but read together, the author's own A100 numbers make the practical case narrow:
- At S=1024 / head_dim=256 (realistic config), flash_cuda is **~6× slower than *both* eager and cuDNN sdpa**, and the per-rank memory win in the real 8-GPU run is only **~1.3×** (the ~2.4× on the isolated attention stack dilutes once params/DDP buffers dominate).
- The long-context sweep shows sdpa's **memory-efficient backend does accept the block-causal mask** and is **15–25× faster** up to S≈98k. flash_cuda's *only* uncontested niche is **S≥131072**, where sdpa OOMs on the dense O(S²) mask and flash is the only backend that runs.

So the question for us as maintainers: do pi07_paligemma runs actually hit that extreme-S regime, and is ~1200 lines of custom CUDA worth maintaining for it? The strongest practical case is the bit the author flagged: the **action-expert non-square `(B, N, n_cross+N)` mask**, where sdpa's mem-efficient backend may reject the shape and fall back to math (≈eager). If that's what real training selects, flash_cuda could genuinely help there at normal sequence lengths — that's the regime I'd want measured before committing, more than the long-context tables.

### Remaining gate + nits

- **Merge gate:** the nightly **model** regression (full pi07_paligemma training) still hasn't run — required for a policy change. The A100 work validated the op + a distributed attention-stack microbenchmark, not an end-to-end training step.
- **Minor (untested):** the new `flash_cuda` config rejection in pi0/pi05/pi05_mem has no test — a 3-line `pytest.raises(ValueError, match="flash_cuda")` per config would lock in the loud-failure behavior against a future backbone-validator refactor.
- **Very minor:** `flash_bwd` has no explicit opt-in-smem `TORCH_CHECK` like the forward — `pick_nw` falls through to `nw=1` even if that still doesn't fit. Harmless today (the forward needs the most smem and its guard fires first in any autograd flow), but a one-line check would make `flash_bwd` self-contained.

Net: correctness and engineering are solid and all my prior points are handled. The decision is now a maintainer judgment call on the value/maintenance tradeoff plus the pending regression run — not the code.

---

shuheng-liu (approved):

PR #359, opened 2026-05-30
fix(datasets): dual-split feature-mapping collision + norm-distribution diagnostic

shuheng-liu:
## What this does

Two linked changes (🐛 Bug + 🗃️ Feature).

**Bug fix — `(robot_type, control_mode)` feature-mapping collision.**
`DATA_FEATURES_NAME_MAPPING` was keyed by `repo_id` alone. When one `repo_id` is used by two mixture entries with different control modes — e.g. `control_mode="joint"` mapping `actions -> action_joint` and `control_mode="ee"` mapping `actions -> action_ee` — `DatasetConfig.__post_init__` registered both under the same key, so the second clobbered the first and **both control modes silently read `action_ee`**; `action_joint` never reached training. The collision was total: stats standardization, per-sample data loading, and the DataLoader-worker re-injection all resolve via the same `repo_id`-keyed global. On a wide co-training mixture this silently turned every `::joint` head into a duplicate of its `::ee` head, so joint-space action supervision was entirely absent.

Fix: a composite `repo_id::control_mode` key (`feature_mapping_key` / `resolve_feature_mapping` in `standard_data_format_mapping.py`), registered alongside the plain `repo_id` (kept as a back-compat fallback for built-in defaults, single-mode datasets, and camera consumers). Training-critical read sites resolve composite-first against each module's local symbol (so existing patched tests keep working); the aux camera/BPE consumers keep the `repo_id` fallback with TODO breadcrumbs.

**Feature — `scripts/diagnose_norm_distribution.py`.**
A CPU-only, streaming, parallel diagnostic that reads the *real* per-frame state/action values from dataset parquet (the configured episode subset), normalizes them with the per-`(robot_type, control_mode)` head stats, and reports the distribution of normalized values per `(robot_type, control_mode, dim)` — plus per-head plots, a health heatmap, a data-vs-metadata staleness report, and a data-derived corrected-stats artifact. It detects stale `meta/stats.json` / `episodes_stats` (outliers removed from the parquet but never recomputed in the stats metadata), which every metadata-derived view misses. Histograms are fixed-size (bounded memory) and workers are pure-numpy (torch-free).

## How it was tested

- Added `test_feature_mapping_dual_split` in `tests/datasets/test_feature_mapping_dual_split.py`: asserts the two mappings coexist, the resolver returns the correct column per mode, and `DatasetMixtureMetadata` produces distinct joint vs ee action stats (the regression this closes).
- Added accumulator + worker tests in `tests/scripts/test_diagnose_norm_distribution.py`: streaming histogram vs brute-force numpy, merge associativity, NaN/Inf handling, and a parquet round-trip.
- `pre-commit run --all-files` clean; `pytest -m "not gpu" -n auto` green for the dataset/config/script suites (remaining local failures are environment-only: macOS video-decode backends and the uninstalled `libero` extra).
- Verified on a real co-training mixture: every dual-split robot's `::joint` and `::ee` heads went from byte-identical action stats to distinct, and the diagnostic surfaced stale metadata (real data max ≈ O(1) vs metadata max ≈ 3e4 on the affected datasets).
- GPU pytests (`pytest -m "gpu"`) and nightly regression are expected to run in CI (`gpu_test.yml` / `regression_test.yml`); this change is in the shared data/normalization path but touches no policy modeling code, and CUDA was unavailable locally.

```bash
pytest -sx tests/datasets/test_feature_mapping_dual_split.py
pytest -sx tests/scripts/test_diagnose_norm_distribution.py
```
```bash
python src/opentau/scripts/diagnose_norm_distribution.py \
    --config=<DatasetMixtureConfig.json> --output_dir=outputs/norm_diag \
    --num_workers=16 --max_episodes_per_dataset=5
```

## Checklist

### DIFF
--- a/src/opentau/configs/default.py
+++ b/src/opentau/configs/default.py
@@ -153,11 +153,24 @@ class DatasetConfig:
                 f"got {self.tolerance_s} for {self.repo_id or self.vqa}."
             )
 
-        # If data_features_name_mapping is provided, upsert it into the global DATA_FEATURES_NAME_MAPPING
+        # If data_features_name_mapping is provided, upsert it into the global
+        # DATA_FEATURES_NAME_MAPPING. Register under the plain repo_id (back-compat
+        # fallback, last-wins) AND, when this entry carries a real control mode,
+        # under the composite `repo_id::control_mode` key. The composite key lets
+        # two entries that share a repo_id but use different action columns
+        # (e.g. control_mode="joint" -> action_joint vs control_mode="ee" ->
+        # action_ee) coexist instead of the second silently clobbering the first.
         if self.data_features_name_mapping is not None:
-            from opentau.datasets.standard_data_format_mapping import DATA_FEATURES_NAME_MAPPING
+            from opentau.datasets.standard_data_format_mapping import (
+                DATA_FEATURES_NAME_MAPPING,
+                feature_mapping_key,
+            )
 
             DATA_FEATURES_NAME_MAPPING[self.repo_id] = self.data_features_name_mapping
+            if self.repo_id is not None:
+                key = feature_mapping_key(self.repo_id, self.control_mode)
+                if key != self.repo_id:
+                    DATA_FEATURES_NAME_MAPPING[key] = self.data_features_name_mapping
 
 
 @dataclass
--- a/src/opentau/datasets/dataset_mixture.py
+++ b/src/opentau/datasets/dataset_mixture.py
@@ -77,7 +77,7 @@ from torch.utils.data import ConcatDataset, DataLoader, Dataset, Sampler
 from opentau.configs.train import TrainPipelineConfig
 from opentau.datasets.compute_stats import aggregate_stats
 from opentau.datasets.lerobot_dataset import BaseDataset, DatasetMetadata
-from opentau.datasets.standard_data_format_mapping import DATA_FEATURES_NAME_MAPPING
+from opentau.datasets.standard_data_format_mapping import DATA_FEATURES_NAME_MAPPING, feature_mapping_key
 
 
 class _TaggedDataset(Dataset):
@@ -276,7 +276,8 @@ class DatasetMixtureMetadata:
         # the standardize loop below clobbers `metadata.stats`.
         raw_dims: list[tuple[int, int]] = []
         for m in metadatas:
-            name_map = DATA_FEATURES_NAME_MAPPING[m.repo_id]
+            key = feature_mapping_key(m.repo_id, getattr(m, "info", {}).get("control_mode"))
+            name_map = DATA_FEATURES_NAME_MAPPING[key if key in DATA_FEATURES_NAME_MAPPING else m.repo_id]
             raw_dims.append(
                 (
                     int(m.stats[name_map["state"]]["mean"].shape[-1]),
@@ -286,7 +287,9 @@ class DatasetMixtureMetadata:
 
         # convert each metadata stats to the standard data format
         for metadata in metadatas:
-            metadata.stats = self._to_standard_data_format(metadata.repo_id, metadata.stats)
+            metadata.stats = self._to_standard_data_format(
+                metadata.repo_id, metadata.stats, getattr(metadata, "info", {}).get("control_mode")
+            )
 
         # Per-dataset stats kept for diagnostic / back-compat consumers
         # (e.g. `aggregated_action_stats` for the BPE codec).
@@ -472,7 +475,7 @@ class DatasetMixtureMetadata:
         return agg["actions"]
 
     def _to_standard_data_format(
-        self, repo_id: str, stats: dict[str, dict[str, np.ndarray]]
+        self, repo_id: str, stats: dict[str, dict[str, np.ndarray]], control_mode: str | None = None
     ) -> dict[str, dict[str, np.ndarray]]:
         """Convert statistics to the standard data format.
 
@@ -482,6 +485,10 @@ class DatasetMixtureMetadata:
         Args:
             repo_id: Repository ID used to look up feature name mapping.
             stats: Statistics dictionary with dataset-specific feature names.
+            control_mode: Control mode used (with ``repo_id``) to resolve the
+                feature-name mapping, so dual-split entries (joint vs ee) pick
+                their own action column. ``None`` falls back to the plain
+                ``repo_id`` entry.
 
         Returns:
             Statistics dictionary with standard feature names and padded vectors.
@@ -490,7 +497,8 @@ class DatasetMixtureMetadata:
             KeyError: If a required feature is missing from stats or if required
                 statistics (mean, std, min, max) are missing.
         """
-        name_map = DATA_FEATURES_NAME_MAPPING[repo_id]
+        key = feature_mapping_key(repo_id, control_mode)
+        name_map = DATA_FEATURES_NAME_MAPPING[key if key in DATA_FEATURES_NAME_MAPPING else repo_id]
         features_without_stats = ["prompt", "response", "advantage"]
 
         standard_stats = {}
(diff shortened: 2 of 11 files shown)


shuheng-liu (commented):
## Review — dual-split feature-mapping collision + norm diagnostic

I traced the bug fix through every read site and the diagnostic end-to-end. **The core fix is correct, the diagnosis is accurate, and the regression test closes the exact hole.** Findings below are maintainability nits, not blockers.

### What's right

- **Root cause + fix are sound.** Keying `DATA_FEATURES_NAME_MAPPING` by `repo_id` alone did make the second mixture entry clobber the first; the `repo_id::control_mode` composite with a plain-`repo_id` fallback resolves it. All four training-critical read sites are converted consistently: `DatasetConfig.__post_init__` (registration), `DatasetMixtureMetadata` (both the `raw_dims` snapshot loop **and** `_to_standard_data_format`), `LeRobotDataset._get_feature_mapping_key`, `resolve_delta_timestamps`, and the worker re-injection in `_get_worker_name_mapping_overrides`. None left on the old key.
- **Config-override vs on-disk consistency holds.** Registration uses the config's `control_mode`; the stats loop resolves via `meta.info["control_mode"]`, which `_apply_metadata_overrides` writes the override through *before* this runs — so the two keys match. `resolve_delta_timestamps` correctly handles the one path that runs *before* overrides apply by preferring `dataset_cfg.control_mode` then on-disk. That ordering subtlety is the easiest thing to get wrong here and it's handled.
- **Sentinel semantics line up** with `compute_norm_key`: `""`/whitespace falls through the `if cm:` falsy guard, `"unknown"` via the frozenset — so column resolution and the norm-head split agree on what counts as "no control mode."
- **`_normalize` MIN_MAX matches the policy** verbatim (`(x-min)/(max-min+EPS)*2-1`, normalize.py:336-338). Histograms are genuinely bounded-memory and the merge is associative.
- Good test coverage on the resolver, key construction, sentinels, and the end-to-end "joint vs ee heads have distinct stats" regression.

### Findings

**1. Duplicated sentinel set with no sync guard (the one I'd actually fix).**
`_MISSING_CONTROL_MODE_VALUES` in `standard_data_format_mapping.py` is a hand-copy of `_NORM_KEY_MISSING_VALUES` in `dataset_mixture.py`, documented to "mirror" each other but with nothing enforcing it. If someone later adds e.g. `"none"` to one and not the other, column-resolution and the norm-head split silently diverge — which re-introduces *exactly* the class of silent-collision bug this PR fixes. Cheapest guard: a one-line test asserting the two frozensets are equal (avoids an import-direction/cycle decision while still failing loudly on drift).

**2. `_build_tasks` is untested and tightly coupled to `DatasetMixtureMetadata` internals.**
It reaches into `per_norm_key_stats`, `norm_key_to_index`, `dataset_to_norm_index`, `norm_key_to_dataset_names` and drives the constructor with a `SimpleNamespace` cfg shim. I confirmed all those attributes exist today, but the unit tests only exercise the accumulator/worker/histogram math — the metadata-wiring half of the script has no coverage and will rot silently if those internals are renamed. Acceptable for an offline diagnostic, but worth a comment noting the coupling (or a smoke test over a tiny fixture mixture).

**3. Minor diagnostic nits (non-blocking):**
- `head_info` derives `robot_type`/`control_mode` by splitting the norm key on `::`, but a *fallback* norm key is the dataset name (a `repo_id` like `org/name` with no `::`) → `robot_type` becomes the full name, `control_mode` empty. Display-only, fine, just noting.
- `np.asarray(table.column(col).to_pylist(), ...)` round-trips an Arrow column through a Python list; `to_numpy(zero_copy_only=False)` / stacking would skip that. Pure perf, irrelevant to correctness.
- `_emit_corrected_stats` reports `count` as `acc.count.max()` across dims — slightly arbitrary when NaNs make per-dim counts differ.

### Verdict
Ship-worthy. I'd land finding #1 (cheap insurance against re-opening the bug) and optionally jot a coupling note for #2; the rest are nits. Note the policy-change checkbox is correctly left unchecked — this is shared data/normalization path but no modeling code, so the GPU/regression gate doesn't strictly apply, though a CI GPU run wouldn't hurt given it's in the normalization path.


---

shuheng-liu:
Thanks for the thorough trace — addressed in `221edb3`.

**#1 — sentinel drift (landed).** Added `test_missing_control_mode_sentinels_stay_in_sync` asserting `feature_mapping_key`'s `_MISSING_CONTROL_MODE_VALUES` equals `compute_norm_key`'s `_NORM_KEY_MISSING_VALUES`, so a future edit to one and not the other fails loudly in CI instead of silently re-opening the collision. I considered the stronger fix — collapsing the two into one definition via import (the direction is already clean: `dataset_mixture` imports from `standard_data_format_mapping`, no cycle) — but went with the guard test per your suggestion to keep the well-tested `compute_norm_key` path untouched. Happy to switch to the single-source-of-truth import if you'd prefer prevention over detection.

**#2 — `_build_tasks` coupling (landed).** Documented the intentional coupling to `DatasetMixtureMetadata` internals (`per_norm_key_stats`, `norm_key_to_index`, `dataset_to_norm_index`, `norm_key_to_dataset_names`) in the docstring, with a "rename these and update here" note. A smoke test over a tiny fixture mixture is the right follow-up but needs a fixture dataset, so I left it as a TODO rather than bloat this PR.

**#3 — nits (intentionally left, rationale):**
- *fallback-key `robot_type`/`control_mode` split* — kept. A fallback head genuinely has no `(robot_type, control_mode)` (that's why it fell back to the dataset name), so labeling it with the dataset name and an empty mode is the honest display; there's nothing truer to show.
- *`to_pylist()` round-trip* — kept; mirrors the existing `fit_fast_tokenizer` reader and is pure perf. Easy swap to `to_numpy(zero_copy_only=False)` if a profiler ever flags it.
- *`corrected_stats` count = `count.max()`* — kept. It's a diagnostic artifact (not consumed by training without a converter), and `max` = the cleanest dim's frame count. If you'd feed it straight into a normalizer I'll emit per-dim counts instead.

On the GPU/regression note: agreed it's data/normalization path with no modeling change, so the gate doesn't strictly apply — happy to kick a CI GPU run if you want the extra signal before merge.

shuheng-liu (commented):
## Follow-up review — commit `221edb3`

Re-reviewed the delta since my last pass (head now `221edb3`). The new commit is a tight, surgical response to my two findings — nothing else changed.

**Finding #1 (sentinel drift) — resolved.** `test_missing_control_mode_sentinels_stay_in_sync` is exactly the guard I wanted: it imports `_MISSING_CONTROL_MODE_VALUES` and `_NORM_KEY_MISSING_VALUES` and asserts equality, so any future drift between the column resolver and the norm-head split fails loudly instead of silently re-opening the collision. I verified at import time that both resolve to `{'unknown'}`, so the assertion passes. The docstring also correctly names *why* drift matters (read-column vs stats-head disagreement).

**Finding #2 (`_build_tasks` coupling) — resolved as documented.** The new docstring NOTE enumerates the exact `DatasetMixtureMetadata` internals it depends on (`per_norm_key_stats`, `norm_key_to_index`, `dataset_to_norm_index`, `norm_key_to_dataset_names`) and states the rename-fragility + where coverage comes from. That's the right call for an offline diagnostic — a fixture-backed smoke test would be nicer but isn't worth the cost here.

**No new issues introduced.** The diff is +22/−1 across just those two files; I re-read both hunks and they're confined to a docstring and a new test (no behavior change in the bug-fix path).

**Still open (optional, from prior review):** the minor diagnostic perf/cosmetic nits — `to_pylist()` round-trip vs `to_numpy`, `count.max()` in `_emit_corrected_stats`, and the `::`-split display parsing for fallback norm keys. All non-blocking; leave or address at your discretion.

Net: both actionable findings closed. This is good to go from my side once CI is green.


---

PR #361, opened 2026-05-30
fix(deps): bump pyopengl-accelerate to 3.1.10 for numpy 2.x (libero)

shuheng-liu:
## What this does
Fixes a numpy 2.x ABI crash in the `libero` extra (🐛 Bug).

#356 migrated the `libero` extra to numpy 2.x (dropping the `numpy<2` / `gym<0.27` pins) but kept `pyopengl-accelerate==3.1.7`, whose prebuilt manylinux wheels are compiled against the numpy 1.x C-ABI. As soon as LIBERO builds an OSMesa offscreen render context (`MjRenderContextOffscreen` → `mujoco/osmesa` → `OpenGL_accelerate.numpy_formathandler`), the numpy ABI check raises:

```
ValueError: numpy.dtype size changed, may indicate binary incompatibility.
            Expected 96 from C header, got 88 from PyObject
```

which crashes `check_mujoco.py` and every rank during LIBERO env creation.

This bumps `PyOpenGL` / `pyopengl` / `pyopengl-accelerate` from `==3.1.7` to `==3.1.10` in the `libero` extra. From 3.1.9 upstream regenerates the Cython wrappers on its build machines, so the 3.1.10 manylinux wheels are built against numpy 2.x and are ABI-compatible. (A plain reinstall of 3.1.7 does not help — it pulls the same prebuilt numpy-1 wheel; the version has to move.) `uv.lock` is regenerated and the diff is limited to the pyopengl entries — the LIBERO git pin is untouched.

## How it was tested
- On Linux (x86_64, cp310) confirmed `pyopengl-accelerate==3.1.10` imports `OpenGL_accelerate.numpy_formathandler` cleanly under `numpy==2.2.6` (the version the `libero` extra now resolves), where `3.1.7` raised the ABI `ValueError`.
- Ran a full LIBERO training job (8×A100, DeepSpeed ZeRO-2, bf16) past every previously-failing point:
  - `check_mujoco.py` part 2 now prints `Render OK, image shape: (64, 64, 3)` → `success`.
  - LIBERO env creation completes on all ranks and training proceeds.
  - The step-200 LIBERO sim eval runs clean: grid-summary videos render via OSMesa and `eval[overall]: success=0.0% n=64` is logged across all 8 ranks with no ABI error (0% success is expected this early in training).

```bash
python -c "import numpy; print('numpy', numpy.__version__); import OpenGL_accelerate.numpy_formathandler; print('OpenGL_accelerate numpy handler import OK')"
```

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -129,11 +129,15 @@ libero = [
     "robosuite==1.4.0",
     "thop==0.1.1.post2209072238",
     "mujoco>=3.3.5",
-    "PyOpenGL==3.1.7",
+    # pyopengl-accelerate ships a Cython C-extension compiled against a specific
+    # numpy ABI: the 3.1.7 wheels are numpy-1 only and crash with "numpy.dtype
+    # size changed" under numpy 2.x (hit on the libero osmesa render path). 3.1.9+
+    # ship wheels rebuilt against numpy 2; keep PyOpenGL pinned in lockstep.
+    "PyOpenGL==3.1.10",
     "libero",
-    "pyopengl-accelerate==3.1.7 ; sys_platform == 'linux'",
+    "pyopengl-accelerate==3.1.10 ; sys_platform == 'linux'",
     "mujoco>=3.1.6 ; sys_platform == 'linux'",
-    "pyopengl==3.1.7 ; sys_platform == 'linux'",
+    "pyopengl==3.1.10 ; sys_platform == 'linux'",
 ]
 urdf = [
     "rerun-sdk>=0.28.2",
--- a/uv.lock
+++ b/uv.lock
@@ -2275,9 +2275,9 @@ requires-dist = [
     { name = "py-spy", marker = "extra == 'dev'", specifier = ">=0.4.0" },
     { name = "pymunk", specifier = ">=6.6.0" },
     { name = "pynput", specifier = ">=1.7.7" },
-    { name = "pyopengl", marker = "sys_platform == 'linux' and extra == 'libero'", specifier = "==3.1.7" },
-    { name = "pyopengl", marker = "extra == 'libero'", specifier = "==3.1.7" },
-    { name = "pyopengl-accelerate", marker = "sys_platform == 'linux' and extra == 'libero'", specifier = "==3.1.7" },
+    { name = "pyopengl", marker = "sys_platform == 'linux' and extra == 'libero'", specifier = "==3.1.10" },
+    { name = "pyopengl", marker = "extra == 'libero'", specifier = "==3.1.10" },
+    { name = "pyopengl-accelerate", marker = "sys_platform == 'linux' and extra == 'libero'", specifier = "==3.1.10" },
     { name = "pyserial", specifier = ">=3.5" },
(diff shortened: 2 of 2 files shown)


shuheng-liu (commented):
## Review — bump pyopengl-accelerate to 3.1.10 for numpy 2.x

**Verdict: correct, minimal, and well-scoped.** Lockfile-consistent dependency bump that directly resolves the numpy 2.x ABI crash on the LIBERO OSMesa render path. I verified the upstream rationale and it holds up.

### What's good
- **Root cause and fix are accurate.** Upstream's own changelog confirms the mechanism: *"As of release 3.1.9 we no longer check in the Cython generated code … relying on the build machines to generate the wrappers."* So 3.1.9+ wheels are regenerated against current (numpy 2) headers, which is precisely why 3.1.10 clears the `numpy.dtype size changed` error that the pinned 3.1.7 wheel hit. cp310 manylinux/musllinux x86_64 wheels for 3.1.10 are on PyPI (uploaded 2025-08-18), so the A100 target installs from a wheel — no source build needed. (The `Expected 96 … got 88` numbers are numpy's own canonical example for this error class, so the description reads correctly.)
- **Lockfile is consistent.** `uv.lock` updates all three `requires-dist` entries plus both package stanzas, and CI's `uv sync --extra dev --extra libero` is green — confirming the lock matches `pyproject.toml`.
- **Versions kept in lockstep.** `PyOpenGL` and `pyopengl-accelerate` ship together; bumping both to 3.1.10 (not just the accelerate package) avoids a core/accelerate ABI skew. Good call.
- **Testing hits the path CPU CI can't.** The crash only fires once an OSMesa context is built, so the end-to-end A100 run (check_mujoco → env creation → step-200 sim eval rendering) is the right validation. As a bonus it also exercises the other numpy-2 C-extensions in the extra (mujoco / robosuite / robomimic), so #356's migration looks clean beyond just pyopengl.
- Correctly scoped: no policy/training code touched, so the unchecked "policy-related changes" box is right and GPU/regression gating doesn't apply.

### Non-blocking notes
1. **Pre-existing redundancy (not introduced here):** the `libero` extra lists the same project twice — `PyOpenGL==3.1.10` (no marker, all platforms) and `pyopengl==3.1.10 ; sys_platform == 'linux'`. The unconditional entry already covers linux, so the linux-markered one is redundant. The PR correctly bumps both in lockstep; flagging only as a future cleanup, not for this PR.
2. **Heads-up on the slimmer sdist:** because 3.1.9+ stopped vendoring the generated Cython `.c`, the 3.1.10 sdist is ~22 KB vs 3.1.7's ~562 KB and now needs Cython at build time. Irrelevant for the cp310 x86_64 target (wheel exists), but any platform lacking a prebuilt wheel would now require a Cython build env. The lock also drops the old i686 accelerate wheels — fine given the `sys_platform == 'linux'` x86_64 target.
3. **Trivial:** the explanatory comment sits on the `PyOpenGL` line (pure-python) while describing the `pyopengl-accelerate` C-extension's ABI issue. Reads fine; could live on the accelerate line.

Nothing here blocks merge. (Leaving this as a comment since GitHub won't let me formally Approve my own PR.)

---

PR #362, opened 2026-05-30
feat(diagnose_norm): per-dim histograms vs N(0,1), replace heatmaps

shuheng-liu:
## What this does

Improves the read-out of the CPU-only normalization diagnostic `src/opentau/scripts/diagnose_norm_distribution.py`.

Previously each `(robot_type, control_mode)` head was rendered as a 2-D density **heatmap** (dim × bin); its x-axis was the histogram *bin index* but was labeled "normalized z (std units)", so you couldn't read off the actual σ and a healthy channel looked the same as an under-normalized one.

This replaces that with a small-multiple **grid of per-dim histograms of the normalized values** (state + action dims in one figure), each overlaid with the ideal `N(0, 1)` the normalization should produce, on a true σ x-axis (symlog once a dim's extremes exceed 10σ):

- a **healthy** dim hugs the green `N(0, 1)` curve,
- an **under-normalized / stale-stats** dim is a thin spike at 0 (`z_std << 1`),
- a **heavy-tailed** dim is too wide.

The global `health_heatmap.png` is likewise replaced by a single `summary_zstd_distribution.png` (distribution of realized `z_std` over all dims, the same split by feature, a staleness-vs-`z_std` scatter separating the two failure modes, and the heavy-tail mass).

Purely the plotting layer — no change to the streaming accumulator, the JSON/CSV reports, or `corrected_stats.json`.

## How it was tested

- `pre-commit run --all-files` clean (ruff lint + format, etc.).
- `pytest -m "not gpu"` green, including a new matplotlib-guarded smoke test `test_plot_writes_histograms_and_summary` in `tests/scripts/test_diagnose_norm_distribution.py` that drives `_plot` on a synthetic one-head mixture and asserts both the per-head grid and `summary_zstd_distribution.png` are written.
- Ran the script end-to-end against a real π0.7 pretrain dataset mixture and eyeballed the figures: under-normalized heads show the expected thin spikes at 0, healthy heads track `N(0, 1)`, and discrete gripper DOFs show as clean bimodal histograms.

```bash
pytest -sx tests/scripts/test_diagnose_norm_distribution.py
```

```bash
python src/opentau/scripts/diagnose_norm_distribution.py \
    --config=<path/to/train_mixture.json> \
    --output_dir=outputs/norm_diag --num_workers=8 --max_episodes_per_dataset=50
```

The figures land in `outputs/norm_diag/`: one `head__<robot>__<mode>.png` per head plus `summary_zstd_distribution.png`.

## Checklist

### DIFF
--- a/src/opentau/scripts/diagnose_norm_distribution.py
+++ b/src/opentau/scripts/diagnose_norm_distribution.py
@@ -43,7 +43,10 @@ Run (where the parquet data lives, e.g. a training node)::
 ``--config`` is a ``DatasetMixtureConfig`` JSON (``$ref`` includes are
 resolved). ``pyarrow`` is already a (transitive) dependency; plotting also needs
 ``matplotlib`` (e.g. ``uv sync --extra libero``) — without it the JSON/CSV
-reports are still written and only the figures are skipped.
+reports are still written and only the figures are skipped. The figures are one
+per-head grid (``head__<robot>__<mode>.png``) of per-dim normalized-value
+histograms overlaid with the ideal ``N(0, 1)``, plus a global ``z_std`` health
+summary (``summary_zstd_distribution.png``).
 """
 
 from __future__ import annotations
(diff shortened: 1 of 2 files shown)


shuheng-liu (commented):
## Review: per-dim histograms vs N(0,1)

Read the full script (not just the diff) plus the test. Net: a clean, well-scoped plotting-layer change that does exactly what the description says — the streaming accumulator, the JSON/CSV reports, and `corrected_stats.json` are all untouched, and the new read-out is a real improvement over the bin-index heatmap. A true σ x-axis with an N(0,1) overlay makes "under-normalized spike at 0" vs "heavy-tailed" obvious at a glance, where the old heatmap's x-axis was a bin index mislabeled as std units. The density math is right (`counts / (n * width)`, normalized by the *total* count so overflow mass is legitimately the missing area under the curve), the `Agg` backend + `plt.close` avoid figure leaks, and `geom` is computed once and threaded through cleanly.

### How I verified
No `.venv` / `opentau` install in my environment, so I did **not** run `pytest` itself. Instead I loaded the actual PR module file standalone (its heavy imports are all lazy) and drove the real `_plot` / `_plot_head_histograms` / `_plot_summary` / `_finalize_report` on synthetic mixtures:

- **Both features, healthy data (the test's own scenario + the realistic case):** both `head__*.png` and `summary_zstd_distribution.png` written, no warnings. So the new test's two assertions hold.
- **Heavy-tailed dim (exercises the `symlog` branch):** clean, both figures written.
- **Single-feature mixture / all-dims-constant:** see finding #1 — figures still written, but with cosmetic warnings.

### Findings

**1. (minor, edge-case) Degenerate inputs emit cosmetic warnings in `_plot_summary`.** Two cases, both confirmed against the real code:
- A mixture with only `state` *or* only `actions` across all heads → `ax[0, 1].legend()` warns `No artists with labels found to put in legend`.
- Every dim constant (so `z_std ≡ 0`, `zstd.max() == 0`) → `ax[1, 0].set_ylim(0, 0)` plus a `non-positive ylim on a log-scaled axis` warning.

Importantly, I initially worried these could fail CI — they can't: the repo's `[tool.pytest.ini_options]` has **no `filterwarnings`**, so these stay console warnings and **the figures still render in every case**. Both scenarios are also unlikely in practice (real VLA mixtures carry both state and actions, and a whole mixture of dead dims won't happen). So this is low priority. If you want it tidy, one guard covers the common one:
```python
if ax[0, 1].get_legend_handles_labels()[0]:
    ax[0, 1].legend()
```

**2. (minor, UX) The `>bins=%` annotation effectively never fires.** `~fin` is only the two ±inf overflow bins, i.e. mass beyond ±1e5σ. Anything between the visible `xhi` (capped at 1e3) and 1e5σ is drawn outside `set_xlim` (invisible) and isn't counted in `>bins` either. The docstring is technically accurate ("the two open overflow bins"), and `z_std` in the panel title plus `frac_gt10` in the summary already capture spread, so it's a small gap, not a bug. If the intent is "fraction off the visible axis," compute the tail relative to `xhi` instead.

**3. (nit, pre-existing) `head_info` is unused in `_plot`.** Confirmed it was already unused before this PR (the old heatmap path didn't reference it either), so this isn't introduced here — flagging only because the signature is otherwise noise. `main` is the sole caller if you ever want to drop it.

### Optional
The new test covers the happy path (both features, healthy data). A single-feature case would lift coverage and pin down finding #1's behavior; a wide/heavy-tailed dim would lock in the `symlog` branch.

### Not verified
No `pytest` run (package not installed here), no GPU run (N/A — non-policy change), and no end-to-end run against a real mixture (no dataset in this env), so the visual check rests on your manual eyeballing noted in the description.

Overall LGTM — none of the above blocks merge; #1 and #2 are polish you can take or consciously waive.

---

shuheng-liu (commented):
### Correction to finding #1

My empirical probe actually contradicts what I wrote in #1 — correcting the record:

- **The single-feature case does *not* warn.** I claimed `ax[0, 1].legend()` would emit "No artists with labels found to put in legend" for a state-only (or actions-only) mixture. It doesn't: an empty `ax.hist([...])` still registers a labeled (empty) `BarContainer`, so the legend has a handle. The `get_legend_handles_labels()` guard I suggested is therefore unnecessary. Withdrawn.

- **The genuine degenerate-case warning is different.** When *every* dim is constant, all `z_std == 0`, so `zstd.max() == 0` and `ax[1, 0].set_ylim(0, min(6.0, 0 * 1.05))` becomes `set_ylim(0, 0)` → `UserWarning: Attempting to set identical low and high ylims makes transformation singular`. This is on the scatter axis (linear), not a log-axis legend issue. The figure still renders.

Everything else stands: it's cosmetic (no `filterwarnings` in pytest config, figures render in all cases), and an all-dead-dims mixture is not realistic. If you want it gone, the right fix targets the ylim, not the legend:
```python
ax[1, 0].set_ylim(0, max(1.0, min(6.0, float(zstd.max()) * 1.05)) if len(zstd) else 1.0)
```

Apologies for the mischaracterization — findings #2 (`>bins` never fires) and #3 (`head_info` unused) are unaffected, and the overall LGTM holds.

---

PR #363, opened 2026-05-30
fix(datasets): pool norm stats over selected episodes + add excluded_episodes

shuheng-liu:
## What this does

Two coupled changes to dataset normalization (🐛 Bug + 🗃️ Feature):

**1. Normalization now pools SELECTED-episode stats, not all-episode.**
Per-norm-head normalization constants were pooled from each dataset's all-episode aggregate (`LeRobotDatasetMetadata.stats`), not the selected-episode subset the mixture actually trains on (`DatasetConfig.episodes`, typically a ~10% slice). A single corrupt or out-of-selection episode could inflate the normalization `std` by 20–4600×, under-scaling real channels until they were effectively dead to the policy.

`LeRobotDataset.__init__` already computed the selected-episode aggregate (`self.stats`) but discarded it for normalization. It now propagates that aggregate onto `self.meta.stats` — the single channel every consumer reads (the mixture's `per_norm_key_stats`, `aggregated_action_stats`, the BPE codec, and the norm diagnostic). The ImageNet camera override in `datasets/factory.py` runs afterward and layers onto the same dict, so cameras stay ImageNet-normalized; val-split `Subset`s and VQA datasets are unaffected.

**2. New `DatasetConfig.excluded_episodes` denylist.**
A per-dataset list of episode indices to drop. It takes precedence over `episodes` (an index present in both is excluded) and removes those episodes from both the training frames and the stats aggregate. This lets operators drop corrupt episodes that fall *inside* the trained selection via config, without editing the dataset on disk.

Both behaviors share two helpers in `lerobot_dataset.py` — `resolve_selected_episodes` and `aggregate_selected_stats` — so the dataset loader, `fit_fast_tokenizer.py`, and `diagnose_norm_distribution.py` use one implementation of the selection logic (including the v2.0 fallback, since v2.0 datasets have no per-episode stats to recompute from).

## How it was tested

CPU suite (`pytest -m "not gpu"`), all green, plus `pre-commit run --all-files` clean:

- Added 12 tests in `tests/datasets/test_datasets.py`: `excluded_episodes` precedence / None / empty-raise; out-of-selection corrupt episode excluded from `meta.stats`; corrupt-in-selection dropped via `excluded_episodes`; a negative control proving the fixture injects real poison; and unit tests for `resolve_selected_episodes` / `aggregate_selected_stats` (incl. no-selection identity and v2.0 fallback).
- Added `test_norm_head_stats_use_selected_episodes` in `tests/datasets/test_dataset_mixture.py`: end-to-end, `per_norm_key_stats` pools the selected episodes.
- `tests/datasets/` → 508 passed, 7 skipped; `tests/scripts/` tokenizer + norm-diagnostic tests → 20 passed.

Not a policy-code change, so the GPU / regression suites were not run locally (no local GPU); they run in nightly CI.

```bash
pytest -sx tests/datasets/test_datasets.py -k "excluded_episodes or selected"
pytest -sx "tests/datasets/test_dataset_mixture.py::TestWeightedDatasetMixtureIntegration::test_norm_head_stats_use_selected_episodes"
```

Config usage — drop known-bad episodes from a mixture dataset:
```json
{ "repo_id": "your-org/dataset-with-bad-eps", "excluded_episodes": [12, 47] }
```

## Checklist

### DIFF
--- a/src/opentau/configs/default.py
+++ b/src/opentau/configs/default.py
@@ -61,6 +61,9 @@ class DatasetConfig:
             Defaults to None.
         episodes: List of episode indices to use from the dataset. If None, all
             episodes are used. Defaults to None.
+        excluded_episodes: List of episode indices to drop from this dataset.
+            Takes precedence over `episodes`: an index present in both is
+            excluded. If None (default), no episodes are excluded.
         image_transforms: Configuration for image transformations. Defaults to
             ImageTransformsConfig().
         revision: Git revision of the dataset repository to use. Defaults to None.
@@ -109,6 +112,7 @@ class DatasetConfig:
     # Root directory where the dataset will be stored (e.g. 'dataset/path').
     root: str | None = None
     episodes: list[int] | None = None
+    excluded_episodes: list[int] | None = None
     image_transforms: ImageTransformsConfig = field(default_factory=ImageTransformsConfig)
     revision: str | None = None
     use_imagenet_stats: bool = True
--- a/src/opentau/configs/default.py
+++ b/src/opentau/configs/default.py
@@ -63,7 +63,10 @@ class DatasetConfig:
             episodes are used. Defaults to None.
         excluded_episodes: List of episode indices to drop from this dataset.
             Takes precedence over `episodes`: an index present in both is
-            excluded. If None (default), no episodes are excluded.
+            excluded. If None (default), no episodes are excluded. Note: on
+            legacy v2.0 datasets (no per-episode stats) the listed episodes are
+            dropped from training, but normalization stats stay the global
+            (all-episode) aggregate — only v2.1+ can recompute them.
         image_transforms: Configuration for image transformations. Defaults to
             ImageTransformsConfig().
         revision: Git revision of the dataset repository to use. Defaults to None.
--- a/src/opentau/datasets/factory.py
+++ b/src/opentau/datasets/factory.py
@@ -230,6 +230,7 @@ def make_dataset(
             cfg.repo_id,
             root=cfg.root,
             episodes=cfg.episodes,
+            excluded_episodes=cfg.excluded_episodes,
             delta_timestamps=dt_mean,
             delta_timestamps_std=dt_std,
             delta_timestamps_lower=dt_lower,
--- a/src/opentau/datasets/lerobot_dataset.py
+++ b/src/opentau/datasets/lerobot_dataset.py
@@ -1531,7 +1531,11 @@ class LeRobotDataset(BaseDataset):
             # normalization std. The ImageNet camera override in
             # datasets/factory.py runs afterward and layers onto this same dict.
             # (v2.0 datasets have no per-episode stats, so this block is skipped
-            # and meta.stats stays the global aggregate.)
+            # and meta.stats stays the global aggregate.) This intentionally
+            # aliases `self.stats` and `self.meta.stats` to one object; nothing
+            # reads the `LeRobotDataset.stats` attribute for normalization, so
+            # the shared reference (later padded in place by the mixture) is
+            # benign.
             self.meta.stats = self.stats
 
         if self.episodes is None:
--- a/src/opentau/datasets/lerobot_dataset.py
+++ b/src/opentau/datasets/lerobot_dataset.py
@@ -90,6 +90,7 @@ import math
 import shutil
 import traceback
 from abc import abstractmethod
+from collections.abc import Iterable
 from concurrent.futures import ThreadPoolExecutor
 from pathlib import Path
 from typing import Callable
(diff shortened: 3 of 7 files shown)


shuheng-liu:
## Review notes — pool norm stats over selected episodes + `excluded_episodes`

**Verdict:** Solid, well-tested fix for a real and impactful bug. I found no blocking correctness issues. A few minor caveats below — none need to block merge, but two are worth a one-line doc/comment.

### Verified correct
- **`self.meta.stats` mutation is per-instance and safe.** Each `LeRobotDataset` builds its own `self.meta` (`lerobot_dataset.py:1399`); the factory's `ds_meta` (`factory.py:210`) is a separate throwaway used only for delta-timestamp resolution. No shared/cached metadata is clobbered across datasets, and the mixture already reassigns `m.stats` via `_to_standard_data_format` (`dataset_mixture.py:290`), so propagating onto `meta.stats` is consistent with the existing pattern.
- **ImageNet-override ordering claim holds.** `factory.py:258-264` runs after `__init__` and layers camera keys *in place* onto `dataset.meta.stats`, so cameras stay ImageNet-normalized on top of the selected aggregate. The mixture reads `[m.stats for m in metadatas]` (`dataset_mixture.py:296`) → `per_norm_key_stats`, exactly as the PR claims.
- **Version gating is consistent** (`>= v2.1` in `__init__` vs `< v2.1` fallback in `aggregate_selected_stats`) and **selection is deterministic** (`sorted(...)`), so this won't perturb seeded-run reproducibility.
- **Error paths handled:** the emptied-selection `ValueError` is caught by the tokenizer's broad `except Exception` and by the diagnostic's `try/except + continue`.
- **Tests are strong** — the negative control (`test_corrupt_episode_inside_selection_is_inflated_without_exclusion`) proves the fixture really poisons, so the positive tests can't false-pass. The v2.0-fallback and no-selection-identity unit tests pin the helper contract well.

### Minor points (non-blocking)

1. **v2.0 asymmetry deserves a user-facing caveat.** On v2.0 datasets (no per-episode stats), `excluded_episodes` drops *frames* but the `>= v2.1` stats block is skipped, so `meta.stats` stays the global aggregate — a corrupt episode still poisons normalization. `aggregate_selected_stats`'s docstring notes the fallback, but the `DatasetConfig.excluded_episodes` docstring in `default.py` (what operators actually read) doesn't. Someone denylisting a bad episode on a v2.0 dataset *specifically to fix norm* would be surprised it's a no-op for stats. Suggest one line there.

2. **`dataset.stats` ↔ `dataset.meta.stats` aliasing is intentional but unexplained at the source.** After `self.meta.stats = self.stats` they're the same object, so the in-place ImageNet override also mutates `dataset.stats` (it silently gains camera stats). Benign today — nothing reads the `LeRobotDataset.stats` *attribute* for normalization — but a latent footgun. The test asserts the aliasing (`dataset.stats is dataset.meta.stats`); a one-line comment at the assignment explaining *why* they're shared would save a future reader.

3. **Diagnostic: the second `resolve_selected_episodes` call (`diagnose_norm_distribution.py:347`) isn't guarded.** It's safe only because the earlier `aggregate_selected_stats` call already `continue`d past any emptied-selection dataset. That's an implicit coupling between two call sites passing the same args — if they ever drift, the second raises unhandled. Computing `selected` once and reusing it (or a brief comment) would harden it.

4. **Trivial:** `excluded_episodes=[]` is correctly a no-op via `if not excluded_episodes`, but only `None` is unit-tested. Cosmetic.

### CI
Pre-commit, checklist, CodeQL, and the auto-review job are green; **CPU Tests still in progress** at time of writing. Checklist correctly filled (docstrings ✓, policy-change ✗ — no policy code touched, so GPU/regression suites are legitimately N/A here).

---

shuheng-liu:
Thanks for the review — addressed the minor points in f94e6d0:

1. **v2.0 caveat** — added to the operator-facing `DatasetConfig.excluded_episodes` docstring in `default.py`: on v2.0 the listed episodes are dropped from training but normalization stats stay the global aggregate (only v2.1+ can recompute).
2. **`self.stats` ↔ `self.meta.stats` aliasing** — added a comment at the assignment explaining the intentional shared reference and why it's benign (the `LeRobotDataset.stats` attribute is never read for normalization).
3. **Diagnostic's second `resolve_selected_episodes`** — went with the documented-invariant option rather than threading a parallel `kept_eps` list through the second loop's zip: added a comment noting the call can't raise because emptied-selection datasets are already dropped by `aggregate_selected_stats` in the metadata loop above. Threading a parallel list would widen the (already nested-tuple) unpacking and risk an alignment bug for no real gain in a startup-only diagnostic — happy to switch to store-and-reuse if you'd prefer the stronger guarantee.
4. **`excluded_episodes=[]`** — added explicit no-op assertions to the `resolve_selected_episodes` unit test.

shuheng-liu:
**Re-review of f94e6d0** — all four points resolved, and since the commit is docs/comment/test-only there's no behavioral change to re-verify.

1. v2.0 caveat — reads well and is in the right place (the operator-facing config docstring). ✅
2. `self.stats` ↔ `self.meta.stats` aliasing comment — accurate. ✅
3. Diagnostic invariant — agreed, the documented-invariant route is the right call here, **no change needed.** I confirmed the invariant holds robustly: the second loop's `m` is the same metadata object appended in the metadata loop, called with the same `dc.episodes`/`dc.excluded_episodes`, so `aggregate_selected_stats` has already dropped any emptied-selection dataset before the second `resolve_selected_episodes` runs. And if the two sites ever did drift, the failure is a loud `ValueError` at diagnostic startup, not a silent mis-normalization. Threading a parallel `kept_eps` list through that already-nested `zip(..., strict=True)` would add alignment surface for no runtime benefit in a startup-only tool. (If you ever want it DRY without touching the zip, resolve once in the metadata loop where `aggregate_selected_stats` already runs and stash the result alongside `kept_cfgs` — purely optional.)
4. Empty-denylist no-op assertions — good. ✅

No new issues. CI: pre-commit / checklist / auto-review green; **CPU Tests + CodeQL still running** on the new commit.

---

PR #364, opened 2026-05-30
test: cover skip_normalization_weights with N!=M per-norm-key heads

shuheng-liu:
## What this does

Adds end-to-end regression tests for the interaction between `skip_normalization_weights=True` and the per-`(robot_type, control_mode)` normalization heads (stacked `(D, *feat_shape)` `Normalize` / `Unnormalize` buffers, #347). Specifically it covers loading a checkpoint whose saved norm heads number **N** into a finetuning policy whose mixture has **M** heads, with **N != M** — the documented use case for the knob (finetuning a checkpoint whose stats were aggregated over a *different* dataset mixture).

Until now the skip-normalization tests exercised the strip / inf-guard helpers only with single-row buffers; there was no test that strips a genuine multi-head checkpoint and reloads into a different-sized mixture. The new tests build real (backbone-free) `PreTrainedPolicy` subclasses with stacked multi-head buffers and replay the normalization-relevant steps of a per-policy `_load_as_safetensor` override (promote → strip → `load_state_dict` → assert → inject), pinning that:

- the strip is keyed by buffer **name** (`is_norm_buffer_key`), not shape, so it drops stacked buffers of any head count; the freshly-built M-head buffers keep the new mixture's values and the checkpoint's sentinel rows never leak through;
- the one-shot flag resets and `_norm_key_to_index` reflects the new heads;
- **without** the strip, `N != M` is a hard `load_state_dict` size mismatch that raises regardless of `strict` (the legacy promotion shim only rescues single-row → `(1, *feat)`, not an N → M re-stack) — i.e. skip is the *required* knob when the head count changes;
- after the round-trip, inference-time `_resolve_dataset_index` routes the new `(robot_type, control_mode)` keys to the right rows and old checkpoint keys are no longer routable.

Test-only change — no policy/source code modified. Label: `test`.

## How it was tested

- Added `TestSkipNormalizationMultiHeadRoundTrip` (4 tests; one parametrized over shrink `N=3→M=2` and grow `N=2→M=3`) plus typed helpers in `tests/policies/test_pretrained_skip_normalization.py`.
- `pytest -m "not gpu" tests/policies/test_pretrained_skip_normalization.py` → 33 passed.
- Sibling normalization suites, to confirm no regression: `tests/policies/test_normalize_per_dataset.py`, `tests/policies/test_save_pretrained_skip_stats.py`, `tests/datasets/test_norm_key_aggregation.py` → 37 passed.
- `pre-commit run --files ...` → all hooks pass (ruff, ruff-format, pyupgrade, license header, bandit, …).
- CPU-only; no GPU or regression runs required (no policy source changed).

```bash
pytest -sx tests/policies/test_pretrained_skip_normalization.py
```

## Checklist

### DIFF
--- a/tests/policies/test_pretrained_skip_normalization.py
+++ b/tests/policies/test_pretrained_skip_normalization.py
@@ -51,7 +51,7 @@ from opentau.policies.pi07_paligemma.high_level_planner.configuration_pi07_high_
 from opentau.policies.pi07_paligemma.low_level.configuration_pi07_low_level import (
     PI07PaligemmaLowLevelConfig,
 )
-from opentau.policies.pretrained import PreTrainedPolicy
+from opentau.policies.pretrained import PreTrainedPolicy, is_norm_buffer_key
 from opentau.policies.value.configuration_value import ValueConfig
 
 # Parallel list of (cls, display_id) — display IDs disambiguate the two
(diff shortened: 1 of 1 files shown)


shuheng-liu (commented):
## Review

**Verdict: LGTM 👍** — non-blocking comments only. Well-targeted test-only PR that adds real regression coverage for a genuine gap, faithfully exercises the production load path, and is correct.

**Scope:** 1 file, `+309/−1`, test-only (no policy/source changed).

### Verification performed
Beyond reading the diff, I ran it locally against the real source helpers:

- **New `TestSkipNormalizationMultiHeadRoundTrip`:** 5 items pass (`shrink_3_to_2`, `grow_2_to_3`, inject, non-skip-raises, inference-resolve).
- **Full modified file:** `33 passed` — matches the PR description.
- **Sibling suites** (`test_normalize_per_dataset`, `test_save_pretrained_skip_stats`, `test_norm_key_aggregation`): `37 passed` — matches the PR description.
- **Lint/format:** `ruff check` clean; `ruff format --check` already formatted.

Cross-checked the load-bearing claims against source: `_replay_norm_load` matches `PI0Policy._load_as_safetensor` step-for-step (incl. `effective_strict = strict and not stripped_keys`); the `(N, *feat)` / `buffer_<key>` ParameterDict access matches `normalize.py`; `compute_norm_key` emits `"{rt}::{cm}"` (so the order-swapped `[1, 0]` routing assertion is right); `FeatureType(str, Enum)` makes the string-keyed `norm_map` resolve; and `_DummyConfig` implements exactly the abstract members of `PreTrainedConfig`.

### Strengths
- **Faithful to the real path** — leans on the actual shared helpers (`_promote_legacy_norm_buffers_in_state_dict`, `_strip_*`, `_assert_normalize_buffers_initialized`, `_inject_stats`, `_resolve_dataset_index`) rather than mocks.
- **Sentinel-tagged values** (`_CKPT_BASE` vs `_NEW_BASE`) prove provenance, not just row count — the new mixture's rows survive and no checkpoint sentinel row leaks.
- **The contrast test earns its place:** `test_non_skip_mismatched_head_count_raises_size_mismatch` shows *why* the knob is required (size mismatch raises regardless of `strict`).
- **Inference routing test** uses order-swapped keys (`[1, 0]`) to prove a real lookup, and the old-key `ValueError` proves the rebuilt name map is live, not the checkpoint's.
- Parametrized shrink + grow; backbone-free (~0.04s); typed Google-style docstrings.

### One thing to consider (non-blocking)
`_replay_norm_load` reimplements the override glue (promote → strip → `effective_strict` → assert) rather than driving a real `_load_as_safetensor`. No CPU test drives a real override with skip end-to-end, so override drift (e.g. dropping the `effective_strict` gate in a policy's `modeling_*.py`) would stay green here. The docstring acknowledges this, and the glue is identical across all overrides, so the tradeoff (vs. a multi-GB backbone) is reasonable — flagging only to keep it a conscious decision.

### Nits (trivial)
- The new helper keys `norm_map` with strings (`{"STATE": ...}`) while the existing helper at the top of the file uses enum keys (`{FeatureType.STATE: ...}`). Both work (str-enum); matches `test_normalize_per_dataset.py`. Cosmetic.
- `match="size mismatch"` couples to torch's internal message wording — stable in practice.

No production code touched; CPU-only, deterministic, fast — CI risk is negligible.

---

PR #365, opened 2026-05-31
fix: history_state_drop masks history, keeps current state, zeros post-norm

shuheng-liu:
## What this does

🐛 Bug fix for the `history_state_drop_prob` augmentation (default `0.3`).

When the drop fired, the dataset zeroed the **entire** `observation.state` — including the *current* step — in **raw, pre-normalization** units. Because state is `MEAN_STD`-normalized inside the policy, a raw `0` maps to `(0 - mean)/(std + EPS) = -mean/std`, an out-of-distribution extreme (unbounded as `std → 0`). The model's force-unmask (`state_mask[:, -1] = True`) then conditioned the *current* state token on that garbage value — so ~30% of training steps saw a corrupted "current" proprioceptive state instead of merely dropping the history.

The image path was already correct (`IDENTITY`-normalized, historical frames zeroed, current frame kept, masked) and is unchanged.

Fix — drop history by **masking**, never by a pre-normalization zero:

- **Dataset** (`_emit_optional_keys`): no longer zeros `state`; it only marks `obs_history_is_pad` all-True (the drop signal). Historical camera zeroing is unchanged.
- **Each temporal policy** (`pi07`, `pi05_mem`, `pi07_paligemma`): in `embed_prefix` / `_build_prefix_items`, *after* `normalize_inputs` and *before* `state_proj`, zeros the masked (historical) state via `masked_fill` while keeping the current step real and attended. Defense-in-depth: the masked history is provably ignored by attention, and even if that mask regressed the content is a clean post-norm zero — no history leak, no `-mean/std`.

Verified that boundary history padding is **clamped** (real rows via `np.clip`), so no raw zero ever reaches normalization after this change.

Behavioral note: during drop steps the model now conditions on the real current state instead of the `-mean/std` ghost — an intended change in training dynamics, so loss curves will differ from pre-fix runs.

## How it was tested

- Rewrote `test_history_drop_zeros_state` → `test_history_drop_marks_pad_keeps_state_content` (asserts `state` is untouched and `obs_history_is_pad` flips all-True) and added a temporal-state-preserved assertion to the camera test, in `tests/datasets/test_optional_keys.py`.
- Added a post-norm-zero test to each temporal policy: `tests/policies/test_pi07_cpu.py`, `tests/policies/test_pi05_mem.py`, `tests/policies/test_pi07_paligemma_low_level.py` (masked steps → zero at/before projection, current step preserved).
- CPU suite: `pytest -m "not gpu and not slow" tests/datasets tests/policies tests/configs` → **1000 passed, 9 skipped**.
- Determinism (repo policy for data-path / modeling changes): the change adds/removes only deterministic tensor ops (drops a `torch.zeros_like`, adds a `masked_fill`) and touches no RNG, so the augmentation draw sequence is unchanged. Confirmed empirically on the real `_emit_optional_keys` drop path — same seed → bit-identical augmentation sequence, different seed diverges, state preserved on every sample.
- GPU pytests (`pytest -m "gpu"`) for the three changed policies are running on internal GPU infra; the checklist box is checked ahead of their completion as agreed.

```bash
# CPU coverage for the change
pytest -sx tests/datasets/test_optional_keys.py tests/policies/test_pi07_cpu.py \
          tests/policies/test_pi05_mem.py tests/policies/test_pi07_paligemma_low_level.py
```

```bash
# GPU coverage (CUDA box)
pytest -m "gpu" -n 0 tests/policies/test_pi07_low_level.py \
                     tests/policies/test_pi05_mem_gpu.py \
                     tests/policies/test_pi07_paligemma_low_level.py
```

## Checklist

### DIFF
--- a/src/opentau/configs/default.py
+++ b/src/opentau/configs/default.py
@@ -219,11 +219,16 @@ class DatasetMixtureConfig:
             ``history_interval`` attribute (defaults to 1 when the policy
             doesn't define one), so observations are sampled at timesteps
             :math:`t - (T-1)k,\; t - (T-2)k,\; \ldots,\; t`.
-        history_state_drop_prob: Probability of dropping historical frames and
-            ``observation.state`` together during a single ``__getitem__`` call
-            (zeros out ``state`` and historical camera frames; sets
-            ``obs_history_is_pad`` to all True). Must be in ``[0, 1]``. Defaults
-            to 0.3.
+        history_state_drop_prob: Probability of dropping the observation
+            *history* during a single ``__getitem__`` call. When it fires, the
+            historical steps are masked via ``obs_history_is_pad`` (set all True)
+            and the historical camera frames are zeroed; the current step —
+            current ``observation.state`` and current camera frame — is kept.
+            ``state`` is deliberately NOT zeroed here: it is MEAN_STD-normalized
+            downstream, so the dropped history is zeroed *after* normalization
+            inside the policy (zeroing a raw state pre-normalization would map
+            0 to ``-mean/std``, an out-of-distribution extreme). Must be in
+            ``[0, 1]``. Defaults to 0.3.
         subgoal_drop_prob: Probability of dropping all subgoal images during a
             single ``__getitem__`` call. Must be in ``[0, 1]``. Defaults to 0.75.
         subgoal_end_of_segment_prob: Probability of sampling the subgoal frame
--- a/src/opentau/datasets/lerobot_dataset.py
+++ b/src/opentau/datasets/lerobot_dataset.py
@@ -951,8 +951,13 @@ class BaseDataset(torch.utils.data.Dataset):
         ``prepare_metadata`` falls through to its default-pad path).
 
         Dropout order:
-            1. ``history_state_drop_prob``: zero ``state`` and historical camera
-               frames; mark ``obs_history_is_pad`` all True.
+            1. ``history_state_drop_prob``: mark ``obs_history_is_pad`` all True
+               (the policy masks the historical steps) and zero the historical
+               camera frames. ``state`` is left intact here — the dropped history
+               is zeroed *after* normalization inside the policy's
+               ``embed_prefix``, since zeroing a raw ``state`` before MEAN_STD
+               normalization would map 0 to ``-mean/std``. The current step
+               (current state + current camera frame) is always kept.
             2. ``subgoal_drop_prob``: zero every ``subgoalK`` and set the
                single ``subgoal_is_pad`` flag to True (subgoals are all-or-none).
             3. ``response_drop_prob``: only when subgoals were NOT dropped, mask
@@ -983,10 +988,15 @@ class BaseDataset(torch.utils.data.Dataset):
                 return False
             return bool(torch.rand(()) < prob)
 
-        # (1) History + observation.state drop.
+        # (1) History drop. Mark the historical observation steps as padded via
+        # `obs_history_is_pad` (the policy masks them out) and zero the historical
+        # *camera* frames. We deliberately do NOT zero `state` here: state is
+        # MEAN_STD-normalized downstream, so zeroing a raw state would map
+        # 0 -> -mean/std (an out-of-distribution extreme). The dropped history is
+        # instead zeroed *after* normalization inside the policy's `embed_prefix`.
+        # The current step — current state and current camera frame — is kept.
         drop_hist = _roll(self.history_state_drop_prob)
         if drop_hist:
-            standard_item["state"] = torch.zeros_like(standard_item["state"])
             if self.n_obs_history is not None:
                 standard_item["obs_history_is_pad"] = torch.ones(self.n_obs_history, dtype=torch.bool)
                 if self.n_obs_history > 1:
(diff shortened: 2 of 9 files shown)


shuheng-liu:
GPU pytests for the three changed policies passed on an internal GPU dev box:

`pytest -m "gpu" -n 0 tests/policies/test_pi07_low_level.py tests/policies/test_pi05_mem_gpu.py tests/policies/test_pi07_paligemma_low_level.py`

→ **11 passed, 4 skipped, 58 deselected**, 0 failures (69.9s).

The 4 skips are gated full-pipeline integration tests (real Gemma 3 / PaliGemma weights aren't downloaded there); CI's nightly `gpu_test.yml` covers those. The state-masking paths this PR touches are exercised by the passing `pi05_mem_gpu` forward / backward / causality / end-to-end tests and the `pi07_paligemma` obs-history regression test.

shuheng-liu:
## Review — `fix: history_state_drop masks history, keeps current state, zeros post-norm`

**Verdict: looks correct and well-scoped.** I reviewed the change in context (not just the diff). The root-cause analysis is accurate and the fix is sound. One scope question worth confirming before merge, plus a couple of minor notes.

### What it does
Stops the dataset from zeroing a *raw* `observation.state` on a history drop (raw `0` → `(0-mean)/(std+eps) = -mean/std`, an OOD extreme that the force-unmask `state_mask[:, -1] = True` then fed into the *current* state token). Instead the drop is signalled via `obs_history_is_pad`, and the three temporal policies zero the masked steps **after** normalization, **before** `state_proj`.

### Correctness — verified
- **The load-bearing invariant holds.** The fix only works if the `masked_fill(0.0)` runs on already-normalized state. Confirmed for all three policies on **both** the training and inference paths: `normalize_inputs` → `prepare_state` → `embed_prefix`/`_build_prefix_items`.
  - pi07: `modeling_pi07_low_level.py:819`+`:832` (train), `:731` (infer) → masked_fill at `:1528`
  - pi05_mem: `modeling_pi05.py:603`/`:668` → masked_fill at `:1018`
  - pi07_paligemma: `modeling_pi07_low_level.py:1090→1094` / `1153→1168` → masked_fill at `:930`
- **Mask semantics correct.** `masked_fill(~state_mask, 0.0)` zeros padded/historical steps and preserves the current step (`state_mask[:, -1] = True`) and any genuinely-real history.
- **No aliasing bug.** `masked_fill` (non-inplace) returns a fresh tensor, so `batch["state"]` is untouched — important for pi07_paligemma, which re-reads `batch["state"]` at `:1114` after building the prefix items.
- **FSDP-safe (CLAUDE.md §5).** `num_state_tokens` is still `T` (was `state_emb.shape[1]`, now `state.shape[1]` — identical), prefix layout and `att_masks` are unchanged, and no new per-rank branch is introduced. No all-gather count drift.
- **Determinism-safe (CLAUDE.md §3).** Removes a `torch.zeros_like`, adds a `masked_fill`; the `_roll` draw sequence is untouched, so seeded runs stay bit-identical. (Logic-verified; relying on CI/your reported runs for execution.)
- **Complete.** `obs_history_is_pad` is consumed for state in exactly these three policies (grep-confirmed); pi0/pi05/pi06/value and the high-level planners don't read it, so no policy was missed.
- **No residual raw-zero path.** Boundary/start-of-episode history is clamped via `np.clip(idx + delta_idx, ep_start, ep_end - 1)` (`lerobot_dataset.py:1938`), so a raw zero never reaches normalization — the "clamped" claim checks out.
- Uses `rearrange(~state_mask, "b t -> b t 1")` rather than `.unsqueeze(-1)` — matches CLAUDE.md §4. 👍

### Tests
New tests are well-formed and assert the right things (state untouched at the dataset level; masked steps → 0 / current preserved at the policy level). The pi07_paligemma test's exactly-zero assertion is valid because the stub `state_proj` is a `Linear` with `ones_` weight and **`zeros_` bias**, so `proj(0) == 0`. I did not execute them locally (CI covers the CPU suite).

### One thing to confirm — the dataset change is global, not just the temporal policies
`enable_optional_key_dropout` defaults `True` for training (`lerobot_dataset.py:734`) and `history_state_drop_prob` defaults to `0.3` (`configs/default.py:332`), so this drop path fires for **every** policy trained from a dataset mixture — including the non-temporal ones (pi0/pi05/pi06/value) that ignore `obs_history_is_pad`.

For those policies the drop is now effectively a **no-op for state**, whereas before it zeroed their single state → the same `-mean/std` garbage ~30% of steps. So this PR *also* (silently) fixes them — which is good — but:
- it's a training-dynamics change for pi0/pi05/pi06/value that the PR description doesn't mention (the "loss curves will differ" note is framed only around temporal policies; it applies to those too), and
- `history_state_drop_prob` becomes inert for non-temporal policies, with no warning.

Suggest a sentence in the PR description (and maybe a one-line warning/no-op note when `history_state_drop_prob > 0` but the policy has no history) so this isn't a silent surprise for existing non-temporal configs.

### Minor
- The ~10-line masked_fill block + comment is triplicated across the three modeling files. Defensible (they're intentionally-separate forks and §2 favors not over-abstracting), but worth flagging the drift risk — if the masking convention ever changes, all three must move in lockstep.
- The checklist boxes for GPU pytests + regression are checked "ahead of completion as agreed" — PR is draft / `mergeable_state: blocked`, so please confirm those actually go green before merge.


---

PR #366, opened 2026-05-31
fix: guard zero-variance dims in Normalize + pad-aware outlier warning

shuheng-liu:
## What this does

A defensive **zero-variance guard** in the shared `Normalize`/`Unnormalize` (`src/opentau/policies/normalize.py`): when a feature dim's `std` (MEAN_STD) or `max - min` range (MIN_MAX) is ~0 (`|·| < EPS`), snap it to 1 so a value that *deviates* from that "constant" dim normalizes to a bounded `(x - mean)` instead of dividing by `~EPS` and exploding to ~1e8 (the "Outlier normalized state" blow-up). Dims with `std >= EPS` are **bit-identical** to before; the snap is mirrored in `Unnormalize` so `Unnormalize(Normalize(x))` round-trips exactly on a snapped dim.

The guard **warns loudly, but only when the snap actually changes the result** — i.e. when a value really deviates from a dim the stats call constant (stale stats / a value zeroed upstream). A genuinely-constant dim with correct stats maps to `0` and stays silent. The per-step deviation check is gated behind a one-time cached precheck, so models with no constant dims pay nothing on the hot path.

Separately, makes pi07_paligemma's `_warn_state_action_outliers` **pad-aware**: it ignores timesteps the model doesn't attend to (padded history via `obs_history_is_pad`, padded action steps via `action_is_pad`), always keeping the current state frame, so a masked-out frame can't trip the warning. Falls back to scanning all timesteps when no mask is present (back-compatible).

Both warnings **re-fire when a dim's magnitude exceeds the largest already warned for it** (instead of suppressing every repeat after the first), so a steadily-worsening dim is resurfaced — while a steady offender at equal/smaller magnitude still doesn't spam each step. The pi07_paligemma outlier-warning default ceiling (`warn_outlier_threshold`) is also loosened from `10` to `32`: it targets gross ~1e8 blow-ups, so a higher ceiling avoids tripping on benign large-but-finite normalized dims.

🐛 Bug

## How it was tested

- Added `TestZeroVarianceGuard` + `TestSnapWarning` in `tests/policies/test_normalize_per_dataset.py`: a value deviating from a zero-std dim is bounded (`~ -mean`) not exploded; a matching value normalizes to `0`; `std >= EPS` dims are bit-identical to the legacy `(x - mean) / (std + EPS)`; `Unnormalize(Normalize(x))` round-trips on a zero-std dim; the snap warning fires on a real deviation, stays silent when the value matches the constant or no constant dim exists, and re-fires only when the deviation exceeds the last warned magnitude.
- Added pad-awareness + re-warn cases in `tests/policies/test_pi07_paligemma_outlier_warning.py`: a huge value in a masked history slot does not warn; the current (last) frame still warns even when all history is marked padded; real (unpadded) history is still scanned; a padded action step does not warn; a larger magnitude re-warns while an equal/smaller one is suppressed.
- `pre-commit run` clean; `pytest -m "not gpu" -n auto` over the affected files → **160 passed** (plus the pi07_paligemma CPU suite green after the threshold change).
- `pytest -m "gpu" -n 0 tests/policies/test_pi07_paligemma_low_level.py` on a CUDA GPU → **1 passed, 1 skipped** (the full-pipeline test carries an unconditional `@pytest.mark.skip` for GPU memory; the runnable video-encoder masking test passed). Nightly regression runs via `regression_test.yml`.

```bash
pytest -sx tests/policies/test_normalize_per_dataset.py::TestZeroVarianceGuard tests/policies/test_normalize_per_dataset.py::TestSnapWarning
pytest -sx tests/policies/test_pi07_paligemma_outlier_warning.py
```

## Checklist

### DIFF
--- a/src/opentau/policies/normalize.py
+++ b/src/opentau/policies/normalize.py
@@ -22,6 +22,7 @@ datasets the policy was trained on. ``forward`` accepts a per-sample
 row per sample before broadcasting.
 """
 
+import logging
 import sys
 
 import numpy as np
@@ -31,6 +32,82 @@ from torch import Tensor, nn
 from opentau.configs.types import FeatureType, NormalizationMode, PolicyFeature
 
 EPS = 1e-8  # Small epsilon value for numerical stability in normalization
+
+# A genuinely-zero std (|std| < EPS) marks a constant/padding feature dim. The guard in
+# ``Normalize`` / ``Unnormalize`` snaps such a std (or a zero MIN_MAX range) to 1 so a value
+# that *deviates* from that "constant" can't divide by ~EPS and blow up to ~1e8 (the
+# "Outlier normalized state" failure). When the dim is truly constant and its stats are
+# right, every value equals the mean, the deviation is 0, and the snap is a no-op.
+_SNAP_WARN_TOL = 1e-2  # raw-unit |value - mean| above this at a snapped dim => real deviation => warn.
+
+# ``(feature_key, dim)`` offenders already warned about, so each fires once per process
+# (mirrors ``_WARNED_OUTLIER_KEYS`` in the pi07_paligemma policy).
+_WARNED_SNAP_KEYS: set[tuple[str, int]] = set()
+
+
+def _warn_snapped_deviation(
+    key: str,
+    zero_var_mask: Tensor,
+    deviation: Tensor,
+    dataset_index: Tensor,
+    dataset_names: list[str] | None,
+) -> None:
+    """Loudly warn (once per ``(key, dim)``) when the zero-variance guard changed the result.
+
+    The guard snaps a ``std`` (or ``max - min``) of ~0 to 1. That only changes the output when
+    the input *deviates* from the dim's constant ``mean`` (resp. ``min``): a genuinely-constant
+    dim with correct stats has ``deviation == 0`` and stays silent. A non-trivial ``deviation``
+    means the stats are stale or the value was zeroed upstream (e.g. a dropped frame) — surface
+    the offending ``feature``/``dim`` so it can be traced.
+
+    Log-only and called behind ``Normalize._snapping_possible()``; reads tensors without
+    mutating them and fires no collective, so the ``forward`` graph stays identical across
+    ranks (CLAUDE.md rule 5).
+
+    Args:
+        key: Feature name (e.g. ``"observation.state"``).
+        zero_var_mask: Bool ``(B, *feat)`` — True where the gathered std/range was snapped.
+        deviation: ``batch_val - mean`` (MEAN_STD) or ``batch_val - min`` (MIN_MAX), the raw
+            pre-scale numerator; broadcasts against ``zero_var_mask`` onto the batch shape.
+        dataset_index: ``(B,)`` per-sample dataset row, used to name the offending dataset.
+        dataset_names: Ordered dataset names parallel to the buffer rows, or ``None``.
+    """
+    trigger = zero_var_mask & (deviation.abs() > _SNAP_WARN_TOL)
+    if not bool(trigger.any()):
+        return
+    # Collapse every axis except the trailing feature axis to get per-dim offenders. The
+    # inserted-axis count is data-dependent (see ``_gather_and_broadcast``), so a computed
+    # dim tuple is clearer here than a fixed einops pattern.
+    feat_axes = tuple(range(trigger.ndim - 1))
+    per_dim = trigger.any(dim=feat_axes) if feat_axes else trigger
+    dims = torch.nonzero(per_dim, as_tuple=False).flatten().tolist()
+    fresh = [d for d in dims if (key, d) not in _WARNED_SNAP_KEYS]
+    if not fresh:
+        return
+    for d in fresh:
+        _WARNED_SNAP_KEYS.add((key, d))
+    # Provenance: the sample with the largest deviation among the triggers.
+    magnitude = deviation.abs() * trigger
+    sample_axes = tuple(range(1, magnitude.ndim))
+    per_sample = magnitude.amax(dim=sample_axes) if sample_axes else magnitude
+    worst_sample = int(torch.argmax(per_sample))
+    worst_val = float(magnitude.max())
+    ds = None
+    if dataset_names is not None:
+        idx = int(dataset_index[worst_sample])
+        if 0 <= idx < len(dataset_names):
+            ds = dataset_names[idx]
+    logging.warning(
+        "Normalize zero-variance guard fired: feature '%s' dim(s)=%s have std≈0 but a value "
+        "deviates from the constant by up to %.3g (raw units, dataset=%s). std was snapped to 1 "
+        "so only the mean is subtracted (no ~1e8 blow-up), but this almost always means stale "
+        "normalization stats or a value zeroed upstream — inspect this feature's stats. "
+        "Repeats for these (feature,dim) pairs are suppressed.",
+        key,
+        fresh,
+        worst_val,
+        ds,
+    )
 
 
 def _materialize(tensor: Tensor) -> Tensor:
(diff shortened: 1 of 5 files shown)


shuheng-liu (commented):
## Review: zero-variance Normalize guard + pad-aware outlier warning

Read the full diff plus surrounding context in `normalize.py` and `modeling_pi07_low_level.py`. Overall this looks correct and the invariants it claims hold up under inspection. Nothing blocking — the one actionable item is a test gap.

### What I verified
- **std ≥ EPS is genuinely bit-identical.** `torch.where(std_is_zero, 1, std)` is a no-op for non-constant dims, so `deviation / (std + EPS)` equals the legacy expression exactly. The `Unnormalize` mirror uses the same `1 + EPS`, so `Unnormalize(Normalize(x))` round-trips exactly on a snapped dim (forward divides by `1 + EPS`, inverse multiplies by `1 + EPS`). Same holds for the MIN_MAX `(x - min)/(1+EPS)` ↔ `(z+1)/2 · (1+EPS) + min` pair.
- **rule-5 safe.** Both warning helpers are log-only and fire no collectives; the snap is a deterministic elementwise op, so ranks can't diverge. The `is_compiling()` / `is_in_onnx_export()` short-circuits keep the data-dependent reductions out of the trace while the snap itself (the actual fix) still applies under compile/export.
- **`set` → `dict` for `_WARNED_OUTLIER_KEYS` is fully applied** — no leftover `.add()` call sites — and `_attended_steps_mask` mutates a fresh `~pad.bool()` tensor, so `keep[:, -1] = True` doesn't alias `batch["obs_history_is_pad"]`.
- `_attended_steps_mask` faithfully matches `_build_prefix_items` (`state_mask = ~obs_history_is_pad; state_mask[:, -1] = True`).

### Findings

**1. (Should fix) The MIN_MAX snap path is untested.** The zero-range snap was added to both the `Normalize` and `Unnormalize` MIN_MAX branches, but `TestZeroVarianceGuard` / `TestSnapWarning` use `MEAN_STD` only; the pre-existing MIN_MAX test covers the non-zero-range path. Suggest a MIN_MAX analog: a value deviating from a `max == min` dim stays bounded (instead of `/EPS`), the round-trip holds on a zero-range dim, and the snap warning fires.

**2. (Clarify) `warn_outlier_threshold` 10.0 → 32.0.** That's a >3× jump that silences everything in the 10–32 band, on top of the new pad-awareness already removing false positives. For MEAN_STD a genuine normalized value rarely exceeds ~10 std, so 32 is quite permissive. What's the rationale for 32 specifically? A one-line comment next to the default would keep the constant from looking arbitrary.

**3. (Minor) Fallback diverges from the model when `obs_history_is_pad` is absent.** In `_build_prefix_items`, a missing `obs_history_is_pad` means only the last frame is attended (`state_mask = zeros; [:, -1] = True`). But `_attended_steps_mask` returns `None` → scan all timesteps in that case, so a large value in a history slot the model would actually mask can still trip the warning — the same false-positive class this PR is closing. If `obs_history_is_pad` can be `None` here with `T > 1`, consider keeping only the last frame in the fallback to match the model. (If it's never `None` at this site, ignore.)

**4. (Low) No off-switch for the Normalize per-step sync.** `_warn_snapped_deviation`'s `bool(trigger.any())` runs every step whenever `_snapping_possible()` is true — which is basically always under heterogeneous co-training, since padded tail dims have std=0. The marginal cost is small: it's gated behind `not is_compiling()`, and `forward` already syncs unconditionally via the `isinf` asserts. But unlike `_warn_state_action_outliers` (which has the `threshold <= 0` disable) there's no knob to turn this off. Worth a knob or a note for parity, given the codebase otherwise treats per-step syncs as opt-out-able.

**5. (Nit) `_snapping_possible` cache can go stale.** It's memoized on the first `forward`; if a zero-std dim is introduced afterward via `_inject_stats` / `load_state_dict`, the cached `False` sticks. Harmless in the normal load-then-train order (and `forward` asserts finite stats first), but invalidating `_snap_active` in `_inject_stats`, or a one-line note, would harden it.

### CI
Pre-commit, check-checklist, CodeQL, and the `review` job are green; one CPU Tests job was still in progress at review time. GPU + regression run nightly.

---

shuheng-liu:
Thanks for the review — addressed all five in `5f1bc30`.

**1. MIN_MAX snap path tested.** Added `test_min_max_zero_range_is_bounded_and_round_trips` (a value deviating from a `max == min` dim stays bounded instead of `/EPS`, and `Unnormalize(Normalize(x))` round-trips on the zero-range dim) and `test_min_max_zero_range_warns_on_deviation` (silent when the value `== min == max`, warns naming the dim on a real deviation).

**2. `warn_outlier_threshold` 32 rationale.** Kept the value (intentional loosening) and documented it inline: the failure it targets drives normalized values to ~1e8, so 32 flags those by a wide margin while tolerating benign large-but-finite dims — with the zero-variance guard + pad-awareness already removing the common false positives. Happy to dial it back if you'd prefer something tighter.

**3. Fallback now mirrors the model.** `_attended_steps_mask` for `state` no longer returns `None` (scan-all) when `obs_history_is_pad` is absent — it builds `zeros; [:, -1] = True`, exactly matching `_build_prefix_items`' no-mask branch (current frame only). Added `test_no_mask_3d_state_scans_current_frame_only`. In training the dataset always emits the mask, so this only affected the standalone helper / non-dataset callers, but it's now faithful regardless.

**4. Off-switch added for parity.** `_SNAP_WARN_TOL <= 0` now disables the snap warning *and* skips its per-step `trigger.any()` sync (the gate short-circuits before `_snapping_possible()`), mirroring the outlier warning's `threshold <= 0`.

**5. Cache-staleness documented.** Added a note on `_snapping_possible` that the cache assumes the load-then-train order; the snap itself runs unconditionally every `forward`, so a stale cache only delays a *warning*, never a blow-up. Left it as a note rather than wiring invalidation into `_inject_stats` — can harden if you'd rather.

CPU suite green (41 in the two touched files, 160 across the affected set).

shuheng-liu (commented):
## Re-review of `5f1bc30` — all five points addressed

Pulled the new commit and diffed it against the SHA I first reviewed (`34377028`). Each finding is handled, and I re-derived the new tests by hand rather than just trusting the labels.

**1. MIN_MAX snap path — now covered. ✓** `test_min_max_zero_range_is_bounded_and_round_trips` (a deviating value at a `max == min` dim stays ~5 instead of ~3e8, and `Unnormalize∘Normalize` recovers it) and `test_min_max_zero_range_warns_on_deviation` (silent when the value `== min == max`, warns naming dim 3 otherwise). Traced the `(x-min)/(1+EPS)·2−1` ↔ inverse arithmetic — both hold.

**2. Threshold 32 — documented. ✓ (one note, no change requested)** The inline rationale reads fine. Worth being conscious of: now that the zero-variance guard bounds the ~1e8 blow-ups, the residual job of `warn_outlier_threshold` is mostly *nonzero*-variance outliers — and at 32 a genuine ~10–32σ MEAN_STD outlier (corrupt data, or wrong-but-nonzero stats) is now silent where it warned before. You've explicitly accepted that trade and offered to dial back; flagging only so it's a conscious choice.

**3. No-mask fallback — now mirrors the model. ✓** `_attended_steps_mask` builds `zeros; [:, -1] = True` when `obs_history_is_pad` is absent, matching `_build_prefix_items`' no-mask branch (current frame only); a present-but-shape-mismatched mask still falls back to scan-all. `test_no_mask_3d_state_scans_current_frame_only` confirms a value in a non-current history slot is ignored while the current frame still trips. Correct.

**4. Snap-warning off-switch — added. ✓ (minor)** `_SNAP_WARN_TOL > 0` now gates the block *before* `_snapping_possible()`, so `<= 0` disables both the warning and its per-step host sync — good parity with `warn_outlier_threshold <= 0`. Minor: it's a module-level constant (global to all `Normalize` instances) rather than a config field like the outlier threshold — fine for a tuning constant, just not per-model.

**5. Cache staleness — documented. ✓** The `_snapping_possible` docstring now states the load-then-train assumption and that a stale cache only delays a *warning*, never the snap (numerical safety is unconditional). That was the suggested resolution.

No new issues in the delta.

### CI (`5f1bc30`)
Pre-commit, check-checklist, CodeQL, and the `review` job are green; the gating CPU Tests Status is green with one CPU Tests run still in progress when I checked. I couldn't execute the suite locally (no torch in this env), so this rests on CI plus a line-by-line read of the new tests.

This looks ready to me. (Posting as `COMMENT` since GitHub blocks self-approval on your own PR.)

---

PR #367, opened 2026-05-31
Fix fake_tensor_training.py crash on torch 2.10 (FakeTensor .to swap)

shuheng-liu:
## What this does

`src/opentau/scripts/fake_tensor_training.py` crashed on the currently-locked torch (2.10.0), so the no-allocation training smoke it provides was unusable:

```
RuntimeError: Cannot swap t1 because it has weakref associated with it
RuntimeError: _apply(): Couldn't swap Linear.weight
```

Root cause: modern torch's `nn.Module._apply` hardcodes the `torch.utils.swap_tensors` path for `FakeTensor` parameters (`... or isinstance(param, FakeTensor)`), and `swap_tensors` rejects them because `FakeTensorMode` tracks them with weakrefs. The `set_overwrite_module_params_on_conversion` / `set_swap_module_params_on_conversion` flags can't steer around it, so the existing `torch_fake_tensor_module_to_patch()` no longer makes the post-hoc `.to(device)` work — and disabling the swap flag does **not** help.

Fix: build the `nn.Linear` layers directly on `device` inside the `FakeTensorContext` instead of constructing on CPU and calling `.to(device)`. The params are created as FakeTensors already on the target device — no swap, and the huge `large_hidden_dim` weights still cost no real memory. Only this one script uses `FakeTensorContext`, so nothing else is affected. (🐛 Bug)

## How it was tested

CPU-only, no network, no GPU:

```bash
uv run python src/opentau/scripts/fake_tensor_training.py --device=cpu
```

- Exits 0 and prints the FakeTensor params (size `1e9`, i.e. no real allocation), per-step `symbolic loss = zuf0..zuf4`, and the final `Symbolic mean loss: FloatTrueDiv(zuf0 + ... + zuf4, 5.0)`.
- Ran twice: the symbolic loss series is bit-identical. The "dummy numpy array" column differs run-to-run by design — that is the pre-existing `torch_fake_tensor_to_numpy_patch`, which returns `np.random.rand(...)`; unrelated to this change.
- `pre-commit run --files src/opentau/scripts/fake_tensor_training.py` passes (ruff lint + format, line-length, typos, license header, bandit).

Note: `auto_torch_device()` returns `mps` on Apple Silicon, so the repro uses explicit `--device=cpu`.

```bash
uv run python src/opentau/scripts/fake_tensor_training.py --device=cpu
```

Before this PR it raises `RuntimeError: _apply(): Couldn't swap Linear.weight`; after it, it prints the symbolic per-step losses and a final symbolic mean loss without raising.

## Checklist

### DIFF
--- a/src/opentau/scripts/fake_tensor_training.py
+++ b/src/opentau/scripts/fake_tensor_training.py
@@ -44,12 +44,18 @@ def main(args: Args):
     # ---------------------------------------------------------------
     # Everything inside this context will use FakeTensorMode
     with FakeTensorContext():
-        # Create a model in FakeTensorContext shouldn't cost real memory for model parameters.
+        # Build the layers directly on `device` rather than constructing on CPU and calling
+        # `.to(device)` afterwards. Modern torch routes Module.to() for FakeTensor params through
+        # torch.utils.swap_tensors, which rejects them ("Cannot swap t1 because it has weakref
+        # associated with it") — and it hardcodes that swap path for FakeTensors, so the
+        # set_{overwrite,swap}_module_params_on_conversion flags can't avoid it. Creating the
+        # params on `device` inside FakeTensorMode keeps them symbolic, so the huge
+        # `large_hidden_dim` weights still cost no real memory.
         model = torch.nn.Sequential(
-            torch.nn.Linear(args.dim_in, args.large_hidden_dim),
-            torch.nn.Linear(args.large_hidden_dim, args.large_hidden_dim),
-            torch.nn.Linear(args.large_hidden_dim, args.dim_out),
-        ).to(device)
+            torch.nn.Linear(args.dim_in, args.large_hidden_dim, device=device),
+            torch.nn.Linear(args.large_hidden_dim, args.large_hidden_dim, device=device),
+            torch.nn.Linear(args.large_hidden_dim, args.dim_out, device=device),
+        )
         optimizer = torch.optim.SGD(model.parameters(), lr=0.001)
     # End of FakeTensorContext
     # ---------------------------------------------------------------
--- a/src/opentau/utils/monkey_patch.py
+++ b/src/opentau/utils/monkey_patch.py
@@ -134,12 +134,21 @@ def torch_full_patch():
 
 @_run_once_only
 def torch_fake_tensor_module_to_patch():
-    """Fix torch.nn.Module.to(device) behavior in FakeTensorMode.
+    """Prefer overwriting (not in-place updating) module params on `.to()` conversion.
 
-    Without this patch, Module.to(device) is a no-op in FakeTensorMode, leading
-    to device mismatch errors. This patch enables proper device conversion.
+    Historically this flag made ``Module.to(device)`` behave sensibly in FakeTensorMode
+    (see https://github.com/pytorch/pytorch/issues/119665). On modern torch (confirmed on
+    the locked 2.10) it is no longer sufficient: ``nn.Module._apply`` hardcodes the
+    ``torch.utils.swap_tensors`` path for any ``FakeTensor`` param
+    (``... or isinstance(param, FakeTensor)``), and ``swap_tensors`` rejects them because
+    ``FakeTensorMode`` tracks them with weakrefs — neither this flag nor
+    ``set_swap_module_params_on_conversion`` can steer around it. So the supported pattern
+    is to construct modules directly on the target device inside FakeTensorMode
+    (``nn.Linear(..., device=device)``) rather than building on CPU and calling
+    ``.to(device)`` afterwards.
 
-    See https://github.com/pytorch/pytorch/issues/119665 for more details.
+    The flag is retained as a harmless, defensive default (it is a no-op for FakeTensor
+    params); it only affects the now-unused ``Module._apply`` conversion path.
     """
     torch.__future__.set_overwrite_module_params_on_conversion(True)
 
--- a/tests/scripts/test_fake_tensor_training.py
+++ b/tests/scripts/test_fake_tensor_training.py
@@ -1,0 +1,55 @@
+# Copyright 2026 Tensor Auto Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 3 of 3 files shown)


shuheng-liu (commented):
## Review: Fix `fake_tensor_training.py` crash on torch 2.10

**Verdict: LGTM.** Small, well-scoped bug fix with an accurate root-cause writeup. The fix is correct and is actually *more* version-robust than the code it replaces. Everything below is non-blocking.

### What it does
Replaces the post-construction `.to(device)` on the `nn.Sequential` of `nn.Linear` layers with per-layer `device=device` kwargs, so parameters are created as FakeTensors already on the target device inside `FakeTensorContext`.

### Correctness — confirmed
- **Root cause checks out.** On the locked torch (2.10.0) `nn.Module._apply` forces the `torch.utils.swap_tensors` path for `FakeTensor` params, and `swap_tensors` rejects weakref-tracked fake tensors; the `set_{overwrite,swap}_module_params_on_conversion` flags can't steer around it.
- **The new pattern is canonical.** Building inside an active `FakeTensorMode` with `device=` means the factory calls inside `Linear.reset_parameters` are intercepted by fake mode, so params stay symbolic (`size 1e9`, no real allocation) and never touch the swap machinery.
- **Blast radius is as claimed.** `fake_tensor_training.py` is the only consumer of `FakeTensorContext` in the repo (`run_with_fake_tensor` is defined but unused), so nothing else is affected.
- **Version-robust bonus.** `pyproject` pins `torch>=2.7.1` (lock = 2.10.0); the new construction path behaves identically across that range, whereas the old `.to()` path was the fragile one.
- The remaining `x.to(device=device)` / `y.to(device=device)` in the step loop are *tensor*-level `.to()` on real dataloader tensors entering fake mode (`allow_non_fake_inputs=True`) — a different code path from `Module._apply`, so leaving them untouched is correct.

### Suggestions (optional follow-ups)
1. **`torch_fake_tensor_module_to_patch()` is now effectively dead, and its docstring is stale.** With the only `Module.to(device)` call gone, its `set_overwrite_module_params_on_conversion(True)` is no longer exercised — yet it's still applied on every `FakeTensorContext()` and sets a *process-global* torch flag that affects real modules too. The docstring still claims it "enables proper device conversion," which this PR's own analysis shows is no longer true on 2.10. Consider dropping the patch (and its import in `fake_tensor.py`) or at least correcting the docstring. Fine as a separate PR to keep this one surgical.
2. **No regression coverage.** Nothing under `tests/` or `.github/` exercises this script, which is why a hard crash on the locked torch slipped through. A cheap `-m "not gpu"` smoke test (`Args(device="cpu")` → exit 0, params are `FakeTensor`) would guard the next torch bump. Run it in a **subprocess** — `FakeTensorContext` installs never-undone global monkey patches (`torch.isinf`, `FakeTensor.numpy`, `Beta.__init__`, the conversion flag) that would otherwise leak into the rest of the pytest session.

### CI / mergeability
All current checks are green — `CPU Tests`, `Pre-commit Hooks`, CodeQL, the review bot, and `check-checklist`. The lone red mark is an early `check-checklist` run at PR-creation time, superseded by a passing re-run after the checklist was completed. `mergeable_state: blocked` therefore looks like a required-review gate, not a CI failure.

*Reviewed via static analysis + CI inspection; I did not execute the script (CPU Tests is green and the PR documents a clean local repro).*

---

shuheng-liu:
Thanks for the review. Addressed both non-blocking suggestions in `2d50842`:

**2. Regression coverage** — added `tests/scripts/test_fake_tensor_training.py`, a `-m "not gpu"` smoke that runs the script via `subprocess` with `--device=cpu` and asserts it exits 0, keeps params symbolic (`FakeTensor`), and reaches `Symbolic mean loss`. Subprocess isolation is deliberate, exactly per your note — `FakeTensorContext` installs never-undone global patches (`torch.isinf`, `FakeTensor.numpy`, `Beta.__init__`, the conversion flag) that would otherwise leak into the session. This guards the next torch bump. Passes in CI (CPU Tests green).

**1. Stale/dead patch docstring** — corrected the `torch_fake_tensor_module_to_patch()` docstring so it no longer claims it "enables proper device conversion"; it now documents that modern torch forces the `swap_tensors` path for `FakeTensor` params regardless of the conversion flags, and points to the construct-on-device pattern. Kept the patch body (a harmless, forward-compatible global default) — full removal of the patch plus its `fake_tensor.py` import is the more invasive change, so I left that as the separate surgical PR you suggested.

Also confirmed out of curiosity: the script runs fine on the default `mps` device too (params come back as `FakeTensor(device='mps:0', ...)`), since FakeTensorMode only tracks metadata and never dispatches to the backend — so the `--device=cpu` in the repro is for determinism/portability, not a correctness requirement.

shuheng-liu (commented):
## Re-review (updated to `2d50842` — now 2 commits / 3 files)

Both suggestions from the previous review were addressed cleanly. This is ready to go.

### ✅ Suggestion 1 — stale patch / docstring
`torch_fake_tensor_module_to_patch()` keeps the flag (a reasonable defensive choice) and the docstring now accurately explains that on torch 2.10 the `swap_tensors` path is hardcoded for `FakeTensor` params, that neither conversion flag can steer around it, and that constructing on `device` is the supported pattern. Clear and correct.

### ✅ Suggestion 2 — regression coverage
`tests/scripts/test_fake_tensor_training.py` is exactly the right shape:
- Runs the script in a **subprocess** (the module docstring even captures the never-undone global-monkey-patch rationale), so nothing leaks into the pytest session.
- Matches repo conventions — the `tests/scripts/__init__.py` package and the `pytest.main([__file__, "-v"])` `__main__` guard mirror `test_train.py` / `test_profile_step.py`.
- `@pytest.mark.slow` is registered and, crucially, **not** excluded by the CPU job's `-m "not gpu"` filter — so it actually runs in `cpu_test.yml`. The latest run executed and passed it, which means the fix is now exercised end-to-end on the locked torch 2.10 in CI (the gap that let this regression through is now closed).
- Assertions are meaningful: exit 0, `"FakeTensor"` in stdout (params stayed symbolic — no real allocation of the 1e9 hidden dim), and `"Symbolic mean loss"` (loop reached the end).

### Optional nits (non-blocking)
- **`timeout=600`** is generous for a pure shape-inference smoke that finishes in seconds. A tighter bound (e.g. 120s) would surface a hang faster in CI without pinning an `-n auto` worker for 10 min on failure. Your call.
- **Docstring precision:** "harmless, defensive default … only affects the now-unused `Module._apply` path" is true in practice, but `set_overwrite_module_params_on_conversion(True)` is a *process-global* flag that also changes real-module `.to()` semantics — it's harmless only because this script is the sole place the context is built (and now in a subprocess). Fine as-is; just flagging in case you want to be precise.

### CI
All checks green on `2d50842` — CPU Tests (incl. the new test), Pre-commit, CodeQL, check-checklist, review bot. No superseded failures this round. `mergeable_state: blocked` is the required-review gate, not a CI failure.

**LGTM.**

---

PR #368, opened 2026-06-01
Outlier warning always reaches wandb (gather to rank 0); port to pi07

shuheng-liu:
## What this does

Fixes a logging bug. The normalized state/action **outlier warning**
(`logging.warning("Outlier normalized ...")`) fires inside the policy `forward`
on whichever data-parallel rank holds the offending sample. But `wandb.init()`
runs on rank 0 only and wandb captures only rank-0's console, so on an N-rank
run the warning reached wandb only ~1/N of the time (1/8 on an 8-rank run).

Splits **detection** (pure, collective-free, stays in `forward`) from **logging**:

- New `src/opentau/policies/outlier_utils.py::detect_state_action_outliers` returns
  offender records — no logging, no cross-step dedup, no collectives — so `forward`
  stays rule-5 safe.
- `src/opentau/scripts/train.py::log_outlier_records_distributed` gathers records to
  rank 0 (a cheap count reduction gates a `gather_object`, so the branch is
  rank-aligned across ranks), dedups **globally**, and emits the existing warning on
  rank 0 — so it always reaches wandb no matter which rank detected the outlier.
  Wired into both `update_policy` and the validation loop, gated by a rank-uniform
  `warn_outlier_threshold` check so policies without that field (pi0/pi05/pi06) are
  unaffected.
- Global dedup replaces the old per-rank dedup, so a dim that bounces between ranks
  is warned once instead of up to once per rank.

Also **ports** the warning to the current `pi07` low-level policy (it previously
existed only in the legacy `pi07_paligemma` policy) and adds `warn_outlier_threshold`
to the `pi07` config.

Labels: 🐛 Bug, 🗃️ Feature

## How it was tested

- Added `tests/policies/test_outlier_detection.py` (detection: threshold on/off,
  provenance, variable ndim, pad-awareness, worst-per-`(source,key,dim)`).
- Added `tests/scripts/test_outlier_logging.py` (rank-0 logger: cross-rank merge,
  global dedup, re-warn-only-on-larger, verbatim message, empty-records no-op).
- `pytest -m "not gpu" tests/policies tests/scripts` → 632 passed, 2 skipped
  (includes the 23 new tests). Pre-commit clean.
- Determinism (CLAUDE.md rule 3): verified the detector is batch-immutable,
  RNG-neutral, and deterministic, and the logger runs after `optimizer.step()` and
  touches no params/grads/optimizer-state/RNG — so the per-step loss series is
  unchanged by construction.
- GPU pytests (`pytest -m "gpu"`) and the nightly regression tests were dispatched
  on this branch via `workflow_dispatch`.

```bash
pytest -sx tests/policies/test_outlier_detection.py
pytest -sx tests/scripts/test_outlier_logging.py
```
```bash
pytest -m "gpu" tests/policies/test_pi07_low_level.py tests/policies/test_pi07_paligemma_low_level.py
```

## Checklist


shuheng-liu:
Thanks for the review. On the two nits:

**Fixed (53f7ba1)** — the per-step count now uses `accelerator.reduce(..., reduction="sum")` instead of `gather_for_metrics`. The all-reduce returns the same value to every rank (so the gather branch stays rank-aligned) while dropping the end-of-dataloader padding-trim semantics, so the last-validation-batch edge case can't shave the count. Added a comment explaining the choice so it doesn't get "consistency-fixed" back to `gather_for_metrics`.

**Keeping as-is (intentional)** — `warn_outlier_threshold=32.0` on by default for pi07. This matches the existing pi07_paligemma default, so the two π0.7 variants behave consistently. The threshold is deliberately permissive (only ~1e8-scale blow-ups from bad normalization stats trip it), the added per-step cost is one device→host sync plus one scalar all-reduce (negligible next to the forward/backward and the three existing per-step gathers), and it's a documented one-line off-switch (`--policy.warn_outlier_threshold=0`). Defaulting it off would leave the debug aid silent exactly when someone hits the failure it's meant to catch.

PR #371, opened 2026-06-02
feat(pi07): per-(robot_type, control_mode) state/action projections

shuheng-liu:
## What this does

Ports the `per_group_projection` feature (added to `pi07_paligemma` in #370) to the **primary `pi07`** low-level policy, whose state/action projections were still plain `nn.Linear`.

When `policy.per_group_projection=true`, the three continuous projections (`state_proj`, `action_in_proj`, `action_out_proj`) each keep an independent `(weight, bias)` per `(robot_type, control_mode)` group — the same grouping the stacked `Normalize`/`Unnormalize` heads use — selected per-sample by `group_index` (the value `_resolve_dataset_index` returns). Because the training `forward` and the inference denoise loop both resolve that index the same way and thread it through `embed_prefix` / `embed_suffix` / the denoise loop and all three projections, a given `(robot_type, control_mode)` always routes to the **same projection head at train and eval time**, in lockstep with the norm head. `time_mlp_*` stay shared.

Off by default → byte-identical to the pre-flag model (same params, state-dict keys, numerics). The feature reuses the shared `PerGroupLinear` (`policies/layers.py`) and the checkpoint projection-remap / `ProjectionRemapError` load path (`policies/pretrained.py`): a legacy single-head checkpoint tiles into all groups, a multi-head checkpoint name-remaps onto the new ordering, and loading a per-group checkpoint into a flag-off policy raises rather than silently collapsing the group axis. No changes to the shared `layers.py` / `pretrained.py` machinery — this PR only wires `pi07` into it.

Label: 🗃️ Feature

## How it was tested

- Added `tests/policies/test_pi07_per_group_projection.py` (7 CPU tests): `_apply_proj` dispatch, the state-dict reconciliation through pi07's `model.*` projection prefixes (tile legacy single head → all groups; name-remap reorder + add a new group; downgrade into a flag-off policy raises), per-sample routing through the real `embed_suffix`, and a **train==eval head-consistency lock-in** — a given `(robot_type, control_mode)` resolves to the same head whether tagged by the training `dataset_index` or the eval `(robot_type, control_mode)` pair, and that index selects the same projection row.
- `pytest -m "not gpu" tests/policies/` → **481 passed, 2 skipped** (includes the new file). The flag-off path is byte-identical — existing pi07 tests are unchanged apart from a 1-line test-harness fix in `test_pi07_cpu.py` (its `embed_prefix` fake now carries `_apply_proj`).
- GPU pytests on the GPU dev box: `pytest -m "gpu" -n 0 tests/policies/test_pi07_low_level.py` passes on the real Gemma 3 backbone — the edited `embed_suffix` / `embed_prefix` and the FSDP-wrap forward/backward.
- Determinism (CLAUDE.md rule 3): the new `PerGroupLinear` op is bit-identical across two same-seed CUDA runs — forward and the `index_add` backward, including a same-group collision — and the real `embed_suffix` per-group routing is likewise bit-identical + correct on GPU. A full-model training determinism smoke and a LIBERO sim rollout require multi-GPU (full pi07 does not fit a single dev GPU's optimizer state), so they are left to the nightly GPU + regression CI.

```bash
pytest -sx tests/policies/test_pi07_per_group_projection.py
```

```bash
pytest -m "gpu" -n 0 tests/policies/test_pi07_low_level.py
```

Enable the feature by setting `--policy.per_group_projection=true` on a `pi07_low_level` config whose dataset mixture spans more than one `(robot_type, control_mode)` group (so `len(dataset_names) > 1`); leave it unset for the existing single-head behaviour.

## Checklist

### DIFF
--- a/src/opentau/policies/pi07/low_level/configuration_pi07_low_level.py
+++ b/src/opentau/policies/pi07/low_level/configuration_pi07_low_level.py
@@ -94,6 +94,9 @@ class PI07LowLevelConfig(PreTrainedConfig):
             Gemma-v1 action expert's ``hidden_size`` so that suffix embeddings
             are compatible with the expert's transformer layers.  Defaults to
             1280 (the expert hidden size in ``Gemma3WithExpertConfig``).
+        per_group_projection: When True, use an independent state/action
+            projection per (robot_type, control_mode) group (shares the
+            normalization-head grouping). Defaults to False.
         dropout: Dropout rate. Defaults to 0.1.
         num_steps: Number of flow-matching denoising steps. Defaults to 5.
         max_delay: Maximum number of prefix action steps for real-time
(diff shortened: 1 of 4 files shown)


shuheng-liu:
Thanks for the review. On the one non-blocking suggestion — the `from_pretrained` seeding glue (source-config read, `BLE001` fallback, continue-training superset path) having no direct test — I'm leaving it untested directly, on purpose:

- The substantive logic is covered: `_reconcile_projection_tensor` / `_remap_per_group_projection_weights_in_state_dict` and how they consume `old_dataset_names` (legacy single-head tile, name-remap reorder, add-new-group, downgrade-raises) are exercised by `tests/policies/test_pi07_per_group_projection.py` plus the shared `tests/policies/test_per_group_projection.py`.
- The remaining glue is the thin wrapper that *obtains* `old_dataset_names` (read the source `PreTrainedConfig`, fall back to the policy's current `dataset_names` on a read failure). Exercising it directly goes through `PI07LowLevelPolicy.from_pretrained`, which instantiates the full ~6B Gemma 3 model — not runnable in the CPU test suite. It is also byte-for-byte the same glue already merged in #370 for `pi07_paligemma`.

Happy to add a `gpu`-marked `from_pretrained` round-trip test if you'd rather have it covered there.

PR #374, opened 2026-06-02
feat(train): aggregate val loss per (dataset, control_mode), parallelize

### LINKED ISSUE #373: Aggregate val loss per (dataset, control_mode) outside forward + parallelize validation
## Background

Now that every sample carries dataset provenance in our standard data format — `_TaggedDataset` injects `dataset_index` (the norm-head row, dropout-immune) and `dataset_repo_id` (the deduplicated mixture-level name) into each item (`src/opentau/datasets/dataset_mixture.py:83-135`), and optional keys `robot_type` / `control_mode` are emitted by `_emit_optional_keys` (`src/opentau/datasets/lerobot_dataset.py`) — we can compute per-(dataset, control mode) loss breakdowns from the batch itself, instead of relying on how the validation loop happens to be batched.

Two related shortcomings in the current validation path (`src/opentau/scripts/train.py:844-950`) motivate this issue.

### Problem 1: per-dataset breakdown is an artifact of dataloader structure, not provenance

Today the only per-dataset granularity we get comes from iterating one **separate** dataloader per underlying dataset:

```python
for ds_name, ds_loader in per_dataset_val_dataloaders.items():
    ds_tracker = per_dataset_trackers[ds_name]
    for batch in ds_loader:
        losses = policy.forward(batch)
        ...
```

(`src/opentau/scripts/train.py:868-918`, dataloaders built by `WeightedDatasetMixture.get_per_dataset_dataloaders()` at `src/opentau/datasets/dataset_mixture.py:944-977`.)

Consequences:

- The breakdown granularity is **per source dataset**, never per **(dataset, control mode)** — even though `control_mode` (`{"joint", "ee", "mixed"}`) is exactly the axis we want to slice MSE/CE along, and it is already present in the batch.
- The loss itself (`policy.forward`) returns a single scalar `MSE`/`CE` reduced over the whole batch (`modeling_pi05.py:740`, `modeling_pi06.py:623`, `pi07/low_level/modeling_pi07_low_level.py:931`). There is no per-sample / per-group disaggregation, so we cannot break a *mixed* batch down by provenance at all — the only reason we get any per-dataset numbers today is that each dataloader is homogeneous by construction.
- Crucially this disaggregation does **not** belong inside `forward`. The model shouldn't know about dataset taxonomy or own metric bookkeeping; grouping by `(dataset_index, control_mode)` should happen in the training/eval loop using the provenance keys already in the batch.

### Problem 2: validation badly under-parallelized across ranks

`get_per_dataset_dataloaders()` creates one DataLoader per dataset, each iterated separately and then sharded across ranks by `accelerator.prepare()`. With a heterogeneous mixture of many **small** validation subsets (some with only a single frame after `val_split_ratio`), this is pathological:

- A dataset with fewer frames than `world_size` leaves most ranks with an empty/padded shard — they do a no-op `forward` (or wait) while one rank does the real work.
- Because we loop datasets **sequentially** (`for ds_name, ds_loader in ...`), the idle time stacks: every tiny dataset is its own under-filled collective round, and `accelerator.gather_for_metrics` + the trailing `accelerator.wait_for_everyone()` (`train.py:950`) force all ranks to rendezvous on each one.
- `drop_last=False` (`dataset_mixture.py:973`) is correct for not discarding val data, but combined with tiny datasets it guarantees ragged, under-utilized batches.

Net effect: validation wall-clock scales with the *number* of datasets rather than the *total* val frames, and most ranks sit idle.

## Proposed direction (for discussion, not prescriptive)

1. **Move per-group aggregation out of `forward` and into the loop, keyed on provenance.** Have policies optionally return *unreduced* (per-sample) MSE/CE — or keep returning scalars but additionally expose a per-sample loss — and let the validation loop bucket them by `(dataset_index, control_mode)` from the batch, using `dataset_index` as the dropout-immune key (`control_mode` can be masked at train time; pair it with `dataset_index` or recover via `compute_norm_key`). Log `Validation/{dataset}/{control_mode}/{MSE,CE}` plus the existing mixture-weighted aggregate (`_mixture_weighted_aggregate`, `train.py:932-946`).
2. **Parallelize validation across all ranks regardless of per-dataset size.** Instead of one sequential under-filled dataloader per dataset, run a **single** validation pass over the *combined* val mixture (so every rank's shard is full), and rely on the per-sample provenance keys for grouping rather than on homogeneous dataloaders. This decouples the breakdown from the batching and keeps every rank busy. (Open question: whether to keep `get_per_dataset_dataloaders` for any callers, or replace it.)

## Acceptance

- Validation logs MSE/CE broken down by `(dataset, control_mode)`, computed in the loop from batch provenance — not inside any `forward`.
- The grouping is keyed on `dataset_index` (dropout-immune) so a mixed batch can be disaggregated correctly; no reliance on one-dataset-per-dataloader homogeneity.
- Validation keeps all ranks busy: a mixture of many 1-frame datasets no longer leaves most ranks idle. Quantify the wall-clock improvement on a representative multi-dataset config.
- Per-step val loss remains deterministic under a fixed seed (per CLAUDE.md hard rule #3).

## References

- Validation loop: `src/opentau/scripts/train.py:844-950`
- Per-dataset dataloaders: `src/opentau/datasets/dataset_mixture.py:944-977`
- Provenance tagging (`dataset_index`, `dataset_repo_id`): `src/opentau/datasets/dataset_mixture.py:83-135`
- Optional keys (`robot_type`, `control_mode`): `src/opentau/datasets/lerobot_dataset.py` (`_emit_optional_keys`)
- Loss returned from forward (scalar MSE/CE): `modeling_pi05.py:740`, `modeling_pi06.py:623`, `pi07/low_level/modeling_pi07_low_level.py:931`
- Metrics plumbing: `MetricsTracker` / `AverageMeter` in `src/opentau/utils/logging_utils.py`; mixture aggregate `_mixture_weighted_aggregate` in `src/opentau/scripts/train.py`

shuheng-liu:
## What this does

Closes #373.

Reworks the validation path so per-dataset loss breakdowns come from **batch provenance** rather than from how the val loop happens to be batched, and so validation stays parallel across ranks regardless of per-dataset size.

Before, validation built **one DataLoader per underlying dataset** and iterated them sequentially. That had two problems:

1. **Granularity was an artifact of batching, not provenance.** `policy.forward` only returned a single batch-reduced `MSE`/`CE` scalar, so a *mixed* batch could not be disaggregated — the only reason we got any per-dataset numbers was that each loader was homogeneous by construction. We could never slice by `(dataset, control_mode)`.
2. **Validation was badly under-parallelized.** With a heterogeneous mixture of many small val subsets (some a single frame after `val_split_ratio`), a dataset with fewer frames than `world_size` left most ranks idle, and looping datasets sequentially made that idle time stack — wall-clock scaled with the *number* of datasets, not total val frames.

This PR:

- Adds a `PerSampleLoss(sum, count)` dataclass and a `return_per_sample` path to `flow_matching_masked_mse`, plus a shared `ce_per_sample` helper, in `policies/utils.py`. Policies optionally expose **per-sample** (unreduced) MSE/CE; the scalar they return is computed exactly as before (bit-identical), so the training path is untouched.
- Threads `return_per_sample` through every action policy (`pi0`, `pi05`, `pi05_mem`, `pi06`, `pi07/low_level`, `pi07_paligemma/low_level`, `value`). Disaggregation lives in the training loop, not in `forward` — the model never learns dataset taxonomy.
- Adds `WeightedDatasetMixture.get_combined_val_dataloader()`: a single deterministic (`shuffle=False`, `drop_last=False`) sequential pass over the whole val mixture, returning `None` when empty.
- Rewrites the validation loop in `scripts/train.py` to run **one** combined pass (every rank's shard stays full), gather per-sample `(sum, count)` + provenance keys via `gather_for_metrics`, and bucket on rank 0 keyed on the **dropout-immune** `dataset_index` (norm head) and a `dataset_repo_id`-derived per-source index. Per-group means are `Σsum / Σcount`. Logs `Validation/{norm_key}/{MSE,CE,Loss}` (headline) and `Validation/by_dataset/{name}/{MSE,CE,Loss}` (per-source), plus the existing mixture-weighted aggregate (`Validation/{Loss,MSE Loss,CE Loss}`). Collective counts stay aligned across ranks (the per-sample flag is rank-uniform; bucketing is rank-0-only arithmetic). Policies whose `forward` doesn't accept `return_per_sample` (the high-level planners) fall back to an aggregate-only pass — they still get the single-pass parallelization.

**Behavior notes for reviewers:**
- Per-group **MSE** is exactly consistent with the legacy normalization (both are `sum/sum`). Per-group **CE** uses a per-valid-token mean (`Σsum/Σcount`, pooling discrete-action + response components), which shifts CE magnitudes modestly vs the old per-dataset numbers — the regrouping changes them regardless.
- The per-subset wandb namespace moves from `Validation/{dataset_name}/…` to `Validation/{norm_key}/…` plus the new `Validation/by_dataset/{name}/…`. The aggregate keys (`Validation/Loss`, `Validation/MSE Loss`, `Validation/CE Loss`) are unchanged.

🗃️ Feature

## How it was tested

- **Pre-commit**: clean (`ruff`, `ruff-format`, `pyupgrade`, `typos`, `bandit`, …).
- **CPU (`pytest -m "not gpu"`)**: the full policy CPU subset is green (487 passed / 2 skipped), plus new unit tests:
  - `tests/policies/test_pi06.py` — `flow_matching_masked_mse(return_per_sample=True)` returns a scalar **bit-identical** to the default path, `Σsum/Σcount` reproduces the masked mean, padded steps/dims and fully-padded samples are excluded from the count; `ce_per_sample` and `PerSampleLoss.__add__` arithmetic.
  - `tests/scripts/test_train.py` — `_bucket_per_sample` disaggregates a **mixed** batch by integer group key with no cross-contamination, group mean is `Σsum/Σcount` (not mean-of-means), and zero-count groups stay NaN-free.
  - `tests/datasets/test_dataset_mixture.py` — `get_combined_val_dataloader` is a single sequential loader over the concatenated mixture, covering every sample exactly once.
- **GPU (`pytest -m "gpu"`)** on the GPU dev box, for every touched policy (`test_pi05`, `test_pi05_mem_gpu`, `test_pi06`, `test_pi07_low_level`, `test_pi07_paligemma_low_level`, `test_value`): 14 passed, 6 skipped (single-GPU skips), 0 failed.
- **Determinism (CLAUDE.md rule 3)**: the per-step validation series should be verified **bit-identical** across two same-seed `configs/dev/dev_config.json` runs — being run separately. By construction the training path is unaffected: `return_per_sample` defaults `False` and the train step never sets it, so the train forward graph and reduction order are byte-for-byte unchanged (a CPU test asserts the scalar is `torch.equal` to the pre-change value). The combined val pass is `shuffle=False` over a fixed order with rank-0-only bucketing arithmetic, so the val series is deterministic.
- Nightly `gpu_test.yml` + regression tests will exercise the full GPU/regression suite.

```bash
pytest -sx tests/scripts/test_train.py::TestBucketPerSample
pytest -sx tests/policies/test_pi06.py::TestFlowMatchingMaskedMsePerSample
pytest -sx "tests/datasets/test_dataset_mixture.py::TestWeightedDatasetMixture::test_get_combined_val_dataloader"
```

The new validation breakdown runs automatically whenever `val_freq > 0`; a smoke run logs `Validation/{norm_key}/…` and `Validation/by_dataset/…` every `val_freq` steps:

```bash
accelerate launch --num_processes 1 src/opentau/scripts/train.py --config_path=configs/dev/dev_config.json
```

## Checklist

### DIFF
--- a/src/opentau/datasets/dataset_mixture.py
+++ b/src/opentau/datasets/dataset_mixture.py
@@ -975,3 +975,46 @@ class WeightedDatasetMixture:
                 worker_init_fn=worker_init_fn,
             )
         return loaders
+
+    def get_combined_val_dataloader(self) -> DataLoader | None:
+        """Create one deterministic sequential DataLoader over the whole mixture.
+
+        Unlike :meth:`get_per_dataset_dataloaders` (one loader per dataset), this
+        returns a single loader over the concatenated mixture so that, under
+        ``accelerator.prepare``, every rank's shard is full even when individual
+        validation subsets have fewer frames than ``world_size`` — the per-dataset
+        loaders leave most ranks idle on tiny subsets and stack that idle time
+        across datasets. Each sample still carries its ``dataset_index`` /
+        ``dataset_repo_id`` provenance (injected by ``_TaggedDataset``), so the
+        validation loop can disaggregate metrics per ``(dataset, control_mode)``
+        from the batch rather than relying on homogeneous per-dataset loaders.
+        ``shuffle=False`` + ``drop_last=False`` make the pass order-deterministic
+        (seed-independent) and score every sample exactly once.
+
+        Returns:
+            A single ``DataLoader`` over the mixture, or ``None`` when the mixture
+            is empty (mirrors the empty-dataset skip in
+            :meth:`get_per_dataset_dataloaders`).
+        """
+        if len(self.concatenated_dataset) == 0:
+            logging.info("Combined validation DataLoader skipped: the mixture is empty.")
+            return None
+
+        worker_name_mapping_overrides = self._get_worker_name_mapping_overrides()
+        worker_init_fn = None
+        if worker_name_mapping_overrides:
+            worker_init_fn = functools.partial(
+                _apply_data_feature_name_mapping_overrides,
+                mapping_overrides=worker_name_mapping_overrides,
+            )
+
+        return DataLoader(
+            self.concatenated_dataset,
+            batch_size=self.cfg.dataloader_batch_size,
+            shuffle=False,
+            num_workers=self.cfg.num_workers,
+            pin_memory=torch.cuda.is_available(),
+            drop_last=False,
+            prefetch_factor=self.cfg.prefetch_factor,
+            worker_init_fn=worker_init_fn,
+        )
--- a/src/opentau/policies/pi0/modeling_pi0.py
+++ b/src/opentau/policies/pi0/modeling_pi0.py
@@ -25,7 +25,7 @@ from collections import deque
 
 import torch
 import torch.nn.functional as F  # noqa: N812
-from einops import rearrange
+from einops import rearrange, reduce
 from torch import Tensor, nn
 from transformers import AutoTokenizer
 
@@ -37,7 +37,7 @@ from opentau.policies.pi0.paligemma_with_expert import (
     PaliGemmaWithExpertModel,
 )
 from opentau.policies.pretrained import PreTrainedPolicy
-from opentau.policies.utils import log_model_loading_keys, make_action_dim_mask
+from opentau.policies.utils import PerSampleLoss, log_model_loading_keys, make_action_dim_mask
 from opentau.utils.accelerate_utils import get_proc_accelerator
 from opentau.utils.utils import get_safe_dtype
 
@@ -466,14 +466,22 @@ class PI0Policy(PreTrainedPolicy):
         return actions
 
     def forward(
-        self, batch: dict[str, Tensor], noise: Tensor | None = None, time: Tensor | None = None
-    ) -> dict[str, Tensor]:
+        self,
+        batch: dict[str, Tensor],
+        noise: Tensor | None = None,
+        time: Tensor | None = None,
+        return_per_sample: bool = False,
+    ) -> dict[str, Tensor | PerSampleLoss]:
         """Do a full training forward pass to compute the loss.
 
         Args:
             batch: Batch of data containing environment observations, actions, and targets.
             noise: Optional noise tensor.
             time: Optional time tensor.
+            return_per_sample: When True, also returns per-sample
+                ``MSE_per_sample``/``CE_per_sample`` (:class:`PerSampleLoss`) for the
+                validation per-(dataset, control_mode) breakdown. ``CE`` is a zero
+                stub for pi0, so ``CE_per_sample`` carries zero sum and count.
 
         Returns:
             A dictionary containing the loss components ("MSE" and "CE").
@@ -537,7 +545,21 @@ class PI0Policy(PreTrainedPolicy):
 
         loss = losses.sum() / (full_mask.sum() + 1e-8)
 
-        return {"MSE": loss, "CE": torch.zeros_like(loss, requires_grad=True)}
+        out: dict[str, Tensor | PerSampleLoss] = {
+            "MSE": loss,
+            "CE": torch.zeros_like(loss, requires_grad=True),
+        }
+        if return_per_sample:
+            # ``losses`` is already masked (and AWR-weighted, if enabled), matching
+            # the scalar reduction; reduce over (chunk, dim) keeping the batch axis.
+            out["MSE_per_sample"] = PerSampleLoss(
+                sum=reduce(losses, "b c d -> b", "sum"),
+                count=reduce(full_mask.float(), "b c d -> b", "sum"),
+            )
+            # pi0 has no CE term; emit zero sum/count so it forms no CE groups.
+            zeros = torch.zeros(losses.shape[0], device=losses.device)
+            out["CE_per_sample"] = PerSampleLoss(sum=zeros, count=zeros.clone())
+        return out
 
     def prepare_images(self, batch: dict[str, Tensor]) -> tuple[list[Tensor], list[Tensor]]:
         """Apply Pi0 preprocessing to the images.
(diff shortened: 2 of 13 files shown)


shuheng-liu:
Thanks for the review. Addressing the two suggestions:

**`train.py` — per-sample gather alignment (fixed in 8559365).** Switched the six separate `gather_for_metrics` calls to a single `gather_for_metrics({...})` over a dict, so accelerate applies one identical ragged-last-batch de-pad to every entry — the per-sample losses and their `norm_index` / `source_index` provenance now stay row-aligned **by construction** rather than relying on six independent de-pads landing on the same trim. A true >1-rank run (plus the same-seed determinism check) is being exercised separately on multi-GPU hardware, since the CPU suite can't cover the multi-rank de-pad path.

**`train.py` — per-group `L1`/`Accuracy` (intentional; keeping aggregate-only).** `L1`/`Accuracy` are emitted only by the value head, which is a single-norm-head policy (`num_datasets=1`) — so its one norm-key group already equals the aggregate, making per-norm-key `L1`/`Accuracy` redundant with `Validation/L1 Loss` / `Validation/Accuracy`. Issue #373 scopes the per-(dataset, control_mode) breakdown to MSE/CE, and `L1`/`Accuracy` aren't per-sample quantities today (they're batch-reduced scalars), so bucketing them would require adding a separate per-sample path to the value head. I'm happy to add per-source `L1`/`Accuracy` as a follow-up if value-function multi-source training ever needs it, but it's out of scope for this PR.

PR #376, opened 2026-06-02
perf(flash_cuda): head-parallel dK/dV + vectorized loads + cp.async

WilliamYue37:
## What this does

(⚡️ Performance) FlashAttention-2 follow-up to #358. The custom block-causal
`flash_cuda` kernel landed in #358 with a real memory win but ran much slower
than the `eager`/`sdpa` backends at the pi07_paligemma shape (head_dim=256, MQA).
This PR applies FA2 techniques to close most of that gap, **benchmarked on a
local RTX 3090** (B2 S1024 D256 bf16, MQA):

| | baseline `flash_cuda` (main) | this PR | eager / sdpa |
|---|---|---|---|
| per-op fwd | 3.14 ms | **2.19 ms** (1.4×) | 0.81 / 0.41 |
| per-op fwd+bwd | 32.4 ms | **7.13 ms** (4.5×) | 2.18 / 2.72 |
| stacked 18-layer **+ckpt** | 842.7 ms | **295.8 ms** (2.85×) | ~97 / ~93 |
| +ckpt peak mem | 0.21 GB | 0.24 GB | 0.50 / 0.28 |

**Three changes, all in `flash_blockmask.cu`:**

1. **dK/dV WMMA kernel parallelized over query heads** (`grid.y = H`, was `Hkv`).
   It now writes per-head fp32 partials `(H, B, Sk, D)` that a new
   `dkv_reduce_kernel` sums over each GQA/MQA head group in a fixed order. This
   removes the serial `for h in group` loop that, under MQA (`Hkv=1`), launched a
   single key-head block looping all 8 query heads — starving the GPU. **No
   atomics**, so bit-identical determinism is preserved (CLAUDE.md rule #3).
2. **Vectorized 128-bit (`uint4`) tile loads** (`load_tile_vec`) across all WMMA
   kernels, replacing the scalar per-element loads that round-tripped every
   element bf16→float→bf16. The kernels were memory-bound, so this was the single
   biggest win.
3. **`cp.async` double-buffered streaming** of Q/dO in the dominant dK/dV kernel
   (`cpasync_tile`, `__pipeline_*`), iterating a contiguous query-tile suffix
   derived from block-id monotonicity (skip-free pipeline).

**Honest verdict:** this is a 2.85–4.5× speedup over the previous `flash_cuda`
and the memory win is preserved (~2× under eager), but it **still does not beat
`eager`/`sdpa`** (~3× off) at head_dim=256. After vectorization the kernels are
compute/smem-bound, not load-bound, so `cp.async` only bought ~9% on the dK/dV
kernel; I deliberately did not force it into the dQ/forward kernels where it
would cost an occupancy drop against the 99 KB sm_86 shared-memory cap. Fully
closing the gap to cuDNN would require a register-resident `mma.sync` rewrite
(tracked as future work). The memory cost rises slightly (per-head dK/dV
partials are `H×` the gradient, transient during backward) but stays well below
eager.

## How it was tested

- All **134 GPU tests** in `tests/policies/test_flash_attn_cuda.py` pass on a
  local RTX 3090 (fp32 vs dense ref ~1e-4; bf16/fp16 WMMA vs `sdpa` ~2e-2; WMMA
  backward vs fp32 autograd ~5e-2; non-square cross-attention; GQA/MQA
  `Hkv ∈ {1,2}`; padding; every mask pattern).
- **Determinism** verified: same-input fwd+bwd is bit-identical across 8 repeats
  on 4 configs (output + dQ/dK/dV), confirming no `cp.async` race and no
  atomic nondeterminism.
- **Benchmarks** above via `python -m opentau.scripts.benchmark_flash_attn`
  (extended to report checkpointed fwd+bwd latency for eager/sdpa/flash).
- **A100 (8×A100-SXM4-80GB): validated.** 134/134 GPU tests + bit-identical determinism on sm_80; kernel-level 2.7–4.1× faster than main's flash_cuda; end-to-end ZeRO-2 max-batch workflow benchmark in the PR comments. Nightly regression suite still pending.

```bash
pytest -m "gpu" -n 0 tests/policies/test_flash_attn_cuda.py
```
```bash
python -m opentau.scripts.benchmark_flash_attn
```

## Checklist

    - Note: ran the `flash_attn_cuda` GPU tests (134 pass) + determinism on a local RTX 3090 **and on A100 (sm_80)**, plus an 8×A100 DeepSpeed ZeRO-2 end-to-end benchmark (see PR comment). Full nightly regression is still pending.

### DIFF
--- a/src/opentau/policies/flash_attn_cuda/_csrc/flash_blockmask.cu
+++ b/src/opentau/policies/flash_attn_cuda/_csrc/flash_blockmask.cu
@@ -33,6 +33,7 @@
 #include <cuda_runtime.h>
 #include <cuda_fp16.h>
 #include <cuda_bf16.h>
+#include <cuda_pipeline.h>
 #include <mma.h>
 #include <c10/cuda/CUDAGuard.h>
 #include <limits.h>
@@ -72,6 +73,57 @@ __device__ __forceinline__ long idx_qkv(long b, long s, long h, long d, long S, 
 }
 __device__ __forceinline__ long idx_lse(long b, long h, long s, long H, long S) {
   return (b * H + h) * S + s;
+}
+
+// Vectorized (128-bit / 8-element) cooperative tile load: copy `rows` rows of D
+// contiguous elements each from a strided global source into a packed smem tile,
+// 16 bytes per transaction. Rows >= max_row are zero-filled. Both src and dst are
+// the WMMA element type (bf16/fp16), so this is a raw byte copy — no per-element
+// dtype round-trip. Threads [tid, tid+nthreads, ...) of the calling group
+// participate. Alignment: D is a multiple of 32 (checked host-side) and every row
+// base is a multiple of D, so all uint4 accesses are 16-byte aligned.
+//   global row r lives at  src_base + (first_row + r) * row_stride
+//   smem   row r lives at  dst      + r * D
+template <typename wt>
+__device__ __forceinline__ void load_tile_vec(
+    wt* __restrict__ dst, const wt* __restrict__ src_base, long row_stride,
+    int rows, int D, long first_row, long max_row, int tid, int nthreads) {
+  constexpr int VEC = 16 / sizeof(wt);  // 8 for bf16/fp16
+  const int dvec = D / VEC;
+  const uint4 zero = make_uint4(0u, 0u, 0u, 0u);
+  for (int c = tid; c < rows * dvec; c += nthreads) {
+    const int r = c / dvec, dv = c % dvec;
+    const long gr = first_row + r;
+    uint4* d4 = reinterpret_cast<uint4*>(dst + (long)r * D) + dv;
+    if (gr < max_row) {
+      *d4 = *(reinterpret_cast<const uint4*>(src_base + gr * row_stride) + dv);
+    } else {
+      *d4 = zero;
+    }
+  }
+}
+
+// Asynchronous (cp.async) variant of load_tile_vec: issues 16-byte global->smem
+// copies that the warp can overlap with compute. The caller must __pipeline_commit()
+// after issuing a tile's copies and __pipeline_wait_prior(...) before reading them.
+// Out-of-range rows are zero-filled synchronously (rare; only the last tile).
+template <typename wt>
+__device__ __forceinline__ void cpasync_tile(
+    wt* __restrict__ dst, const wt* __restrict__ src_base, long row_stride,
+    int rows, int D, long first_row, long max_row, int tid, int nthreads) {
+  constexpr int VEC = 16 / sizeof(wt);  // 8 for bf16/fp16
+  const int dvec = D / VEC;
+  for (int c = tid; c < rows * dvec; c += nthreads) {
+    const int r = c / dvec, dv = c % dvec;
+    const long gr = first_row + r;
+    uint4* d4 = reinterpret_cast<uint4*>(dst + (long)r * D) + dv;
+    if (gr < max_row) {
+      const uint4* s4 = reinterpret_cast<const uint4*>(src_base + gr * row_stride) + dv;
+      __pipeline_memcpy_async(d4, s4, sizeof(uint4));
+    } else {
+      *d4 = make_uint4(0u, 0u, 0u, 0u);
+    }
+  }
 }
 
 // =========================================================================
@@ -475,6 +527,10 @@ __global__ void flash_fwd_wmma_kernel(
   using namespace nvcuda;
   using wt = typename WmmaTraits<scalar_t>::wt;
   auto f2w = WmmaTraits<scalar_t>::from_float;
+  // Raw (bit-compatible) views for vectorized 16-byte tile loads.
+  const wt* Qw = reinterpret_cast<const wt*>(Q);
+  const wt* Kw = reinterpret_cast<const wt*>(K);
+  const wt* Vw = reinterpret_cast<const wt*>(V);
 
   // Shared layout: per-warp slabs for Q/O/P/S/stats, block-shared K/V tile.
   extern __shared__ char smem_raw[];
@@ -509,14 +565,10 @@ __global__ void flash_fwd_wmma_kernel(
   int* qblkW = sQblk + warp * BR_W;
   char* qvalidW = sQvalid + warp * BR_W;
 
-  // Init this warp's accumulator + load its query slab.
+  // Init this warp's accumulator + load its query slab (vectorized).
   for (int idx = lane; idx < BR_W * D; idx += WARP) oW[idx] = 0.f;
   for (int i = lane; i < BR_W; i += WARP) { mW[i] = -INFINITY; lW[i] = 0.f; }
-  for (int idx = lane; idx < BR_W * D; idx += WARP) {
-    const int i = idx / D, d = idx % D;
-    const long qi = q0 + i;
-    qW[idx] = (qi < Sq) ? f2w(to_f(Q[idx_qkv(b, qi, h, d, Sq, H, D)])) : f2w(0.f);
-  }
+  load_tile_vec<wt>(qW, Qw + ((long)b * Sq * H + h) * D, (long)H * D, BR_W, D, q0, Sq, lane, WARP);
   for (int i = lane; i < BR_W; i += WARP) {
     const long qi = q0 + i;
     qblkW[i] = (qi < Sq) ? q_blk[b * Sq + qi] : INT_MIN;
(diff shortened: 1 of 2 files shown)


shuheng-liu (approved):
**Approving.** Traced the rewrite statically and it holds up; the speedup is real and the PR is honest that it's still ~3× off eager/sdpa at `head_dim=256`. GPU-suite/A100 caveat at the bottom.

### What I verified
- **Head-parallel dK/dV is equivalent** to the old in-block group accumulation — per-head fp32 partials `(H,B,Sk,D)` written once each (no atomics), summed in fixed order by `dkv_reduce_kernel`. Partial write/read indexing and the reduce's output decomposition line up with `idx_qkv`, and the `*scale`-on-dK (not dV) placement is preserved.
- **cp.async double-buffer is race-free** — ≤2 batches outstanding (matches `DKV_NBUF=2`), the prefetch slot `(t+1)&1` always differs from the consumed slot `t&1`, and `__pipeline_wait_prior(has_next ? 1 : 0)` + `__syncthreads()` gate each tile's landing and slot reuse correctly. `n_proc ∈ {0,1}` edge cases hold, and the contiguous-suffix `qt_start` processes exactly the old per-tile-skip set given `q_blk`/`k_blk` monotonicity.
- **Shared-memory accounting matches the pointer layout** field-for-field; at `D=256`, `nw=1` ≈ 83.5 KB and `nw=2` ≈ 134.6 KB, so `pick_nw` lands on nw=1 (sm_86) / nw=2 (A100) and both fit.
- **Determinism preserved** (no atomics, fixed-order reduce, same-stream ordering between the dkv kernel and the reduce); **fp32 path untouched**; vectorized-load 16-byte alignment is guaranteed by the host-side `D % 32` check.

### Non-blocking suggestions (for a follow-up, not gating this merge)
1. **(memory scaling)** `dKp`/`dVp` are `O(H·B·Sk·D)` fp32 globals — `2·(H/Hkv)×` the bytes of the final dK/dV per tensor (16× each under MQA `Hkv=1, H=8`). Tiny at the benchmarked shape (~33 MB, still the memory winner with +ckpt) but grows linearly with `B·Sk`; the prior kernel had no global temp here, so a one-line note in the comment about the scaling would set expectations.
2. **(robustness)** The backward dkv/dq WMMA launches use `pick_nw` + `cudaFuncSetAttribute` but lack the actionable `TORCH_CHECK(smem <= max_smem, ...)` the forward path has. If even `nw=1` exceeds the device opt-in cap, `pick_nw` still returns 1 and you get an opaque launch failure instead of a clear message. Pre-existing, and the targeted arches fit `nw=1` at `D=256` — but since this PR raises the dkv footprint, mirroring the forward's check is a cheap win.
3. **(test)** Consider committing a bitwise-determinism regression test (run `flash_bwd` twice, assert dQ/dK/dV bit-identical). The correctness story for the rewrite rests on determinism; you verified it ad hoc but it isn't locked in. Existing tests already exercise the new paths well (GQA `hkv=2` / MQA `hkv=1` group reductions, multi-tile pipelines via `sq=40/50`, non-square cross-attention), so this is purely additive.
4. **(nit)** Backward correctness tests use `H=4`; the production pi07_paligemma shape and the benchmark use `H=8, Hkv=1` (group=8). The reduce is generic over group size so correctness is unaffected, but a group=8 case would be marginally more representative.
5. **(nit, style)** `DKV_PREFETCH` is a large multi-statement macro — justified by the ~15 captured locals, but a `__device__ __forceinline__` helper (or a small pointer-bundle struct) would be more type-safe and debugger-friendly. Optional.
6. **(numerics note)** dK/dV will differ in the last bits from #358's kernel — the group sum is now per-head-partial-then-reduce rather than one running in-block accumulator (a reassociation), and `scale` is applied to fp32 partials before the sum. Both are fine and within your test tolerances; just worth knowing if anyone bisects numerics against #358.

### Caveat
GPU tests / A100 numbers — I could not run the GPU suite locally (CPU-only box), so the 134-pass / determinism claims rest on your reported runs. A100 benchmarking + full nightly regression are still pending per the PR, matching the #358 process. CPU CI is unaffected (the `.cu` isn't built on CPU runners; the benchmark script is import-only).

WilliamYue37:
## A100 validation + benchmarks (8×A100-SXM4-80GB)

A100 results are in (tested as current `main` + this PR's two files, since only these files differ).

### Correctness (single A100, sm_80)
- **134/134 GPU tests pass** (`tests/policies/test_flash_attn_cuda.py`)
- **Bit-identical determinism**: fwd+bwd outputs + dQ/dK/dV identical across 8 repeats × 4 configs
- In-model parity: same-seed ZeRO-2 training losses match eager to bf16 noise (3.25891 vs 3.25984 at step 29)

### Kernel-level (per-op, B1 S1024 D256 bf16 MQA)

| | `main` flash_cuda | this PR | eager | PR speedup vs main |
|---|---|---|---|---|
| fwd | 3.50 ms | **2.63 ms** | 1.02 ms | 1.33× |
| fwd+bwd | 23.09 ms | **5.67 ms** | 2.58 ms | **4.07×** |
| fwd+bwd S2048 | 57.78 ms | **18.31 ms** | 8.05 ms | 3.16× |
| stacked 18L +ckpt | 649 ms / 0.21 GB | **240 ms** / 0.24 GB | 92 ms / 0.50 GB | 2.71× |

### End-to-end standard workflow (DeepSpeed ZeRO-2, 8×A100, max batch)

`configs/benchmarks/pi07_paligemma_low_level.json` (6-obs-step × 3-cam history → S≈4.8k tokens), `gradient_checkpointing=true`, grad-accum 2, 30-step runs, per-rank micro-batch maxed per backend (probed to the OOM boundary):

| backend | max micro-batch/rank | step time | peak GPU mem | global throughput |
|---|---|---|---|---|
| eager @ its max | 40 | **24.0 s** | 79.2 GB | **~26.7 samples/s** |
| flash_cuda @ its max | **48 (+20%)** | 34.0 s | 76.2 GB | ~22.6 samples/s |
| flash_cuda @ B=40 | — | 29.0 s | **67.0 GB** | ~22.1 samples/s |

(sdpa excluded: the block-causal mask forces sdpa onto the math backend, which performs the same as eager for this model.)

### Verdict
- This PR makes `flash_cuda` **2.7–4.1× faster on A100** than the kernel on `main`, with correctness and determinism preserved.
- It now fits **+20% more batch** than eager (67 vs 79 GB at equal batch), but **eager still wins end-to-end throughput by ~18%** at each backend's max batch (24.0 vs 29.0 s/step at equal batch).
- So `flash_cuda` remains the option for memory-constrained / longer-sequence regimes, not a default speed win at this shape. Closing the remaining gap needs the register-resident `mma.sync` rewrite noted in the PR body.

Nightly regression suite: still pending (separate AWS workflow).

WilliamYue37:
## Long-sequence (multi-camera) ZeRO-2 benchmark — where flash_cuda wins

Follow-up to the standard-workflow benchmark above, per request: same 8×A100 DeepSpeed ZeRO-2 setup (grad ckpt on, GA=2), but with the camera count raised to push sequence length past 10k tokens.

**Correction to the previous comment:** the video encoder (`SpaceTimeSiglipVideoEncoder`) drops past-frame tokens — each camera contributes 256 current-frame tokens + 256 subgoal-slot tokens to the prefix, so `S = num_cams × 512 + ~215`. The 3-camera standard workflow above therefore runs at **S ≈ 1.75k** (not "≈4.8k" as I wrote — the step-time/memory numbers are unaffected). The configs here: 8 cameras → **S ≈ 4.3k**, 20 cameras → **S ≈ 10.5k**.

### Max per-rank micro-batch (probed to OOM, midpoint-refined)

| S | flash_cuda max B | eager max B | flash batch advantage |
|---|---|---|---|
| ≈1.75k (3 cam) | 48 | 40 | +20% |
| ≈4.3k (8 cam) | 16 | 12 | +33% |
| **≈10.5k (20 cam)** | **8** (75.3 GB) | **2** (58.4 GB; B=3 OOMs) | **4×** |

At equal batch (B=2, S≈10.5k): eager peaks at **58.4 GB** vs flash **32.8 GB** — the ~26 GB gap is eager's recomputed fp32 (B,H,S,S) score/prob transient, which flash never materializes.

### Throughput at S ≈ 10.5k (20-step timing runs)

| backend | B/rank | step time | global samples/s |
|---|---|---|---|
| flash_cuda @ its max | 8 | 94.0 s | **1.36** |
| eager @ its max | 2 | 19.0 s | **1.68** |

(equal-batch flash@B=2 timing pending — node shut down overnight; will add when it's back up)

### Verdict across regimes (honest)
- **On pure throughput, eager wins in every regime tested** — even at S≈10.5k it delivers ~1.24× the samples/s of flash at each backend's own max batch (1.68 vs 1.36). Per-sample: eager ~0.59 s vs flash ~0.73 s.
- **flash_cuda's win is capability, not speed**: 4× the per-rank batch at S≈10.5k, 26 GB less per-sample memory, and it keeps running where eager hits the OOM wall (eager B=2 is already at 58/80 GB at S≈10.5k — a modestly longer S or larger model leaves eager no feasible batch at all).
- So for this model family: use eager (≈sdpa) for today's S≲2k training; flash_cuda becomes necessary, not just preferable, once S grows past eager's memory wall (~S≈13k+ at B=1, or large-batch long-S regimes).

WilliamYue37:
## Addendum: same long-sequence benchmark on A100-**40GB** (8×A100-SXM4-40GB)

Re-ran the S≈10.5k (20-camera) ZeRO-2 benchmark on a 40GB node (80GB nodes unavailable). Same setup: grad ckpt on, GA=2, per-rank micro-batch probed to OOM, 20-step timings. 134/134 GPU tests pass on this node as well; same-seed equal-batch losses match eager to bf16 noise (3.6853 vs 3.6882).

| backend (40GB) | max B/rank | step time | peak mem | global samples/s |
|---|---|---|---|---|
| flash_cuda @ its max | **3** | 38.0 s | 39.8 GB | 1.26 |
| eager @ its max | 1 (B=2 OOMs) | 12.0 s | 38.4 GB | **1.33** |
| flash_cuda @ B=1 (equal) | — | 16.0 s | **26.7 GB** | 1.00 |

With yesterday's 80GB numbers, the cross-VRAM picture at S≈10.5k:

| | flash max B | eager max B | throughput winner |
|---|---|---|---|
| A100-80GB | 8 | 2 | eager, by ~24% (1.68 vs 1.36) |
| A100-40GB | 3 | 1 | eager, by ~5% (1.33 vs 1.26) |

### Final verdict
- **Throughput: eager edges out flash_cuda in every regime tested** — but the margin shrinks as VRAM tightens (24% → 5% going 80GB → 40GB).
- **Robustness/capability: flash_cuda is the safer long-sequence choice on smaller cards.** On 40GB, eager trains only at B=1 with 1.6 GB of headroom — right on the OOM cliff (any longer S, e.g. ≥24 cameras, leaves eager with no feasible batch), while flash at the same batch keeps 13 GB free and scales to 3× the batch.
- Net: ship `flash_cuda` as an opt-in `attention_implementation` for memory-constrained / long-sequence training; keep eager (≈sdpa) as the throughput default at today's S≲2k workloads.

Issue #385, opened 2026-06-03
RoboCasa: add GPU/regression CI coverage (+ control_freq wiring test)


shuheng-liu:
## Problem

RoboCasa currently has only CPU mock tests (`tests/envs/test_robocasa.py`): registration, factory dispatch, per-rank sharding, the version shim — none of which touch the real sim. LIBERO, by contrast, is exercised in `gpu_test.yml` and the nightly `regression_test.yml`, and has a config-wiring test (`tests/envs/test_libero_control_freq.py`).

So there's no sim-level regression protection for RoboCasa: a break in the wrapper / robosuite-master / robocasa-fork stack wouldn't be caught by CI.

## Why it matters

The RoboCasa stack has several moving parts (robosuite pinned to a master commit, a packaging fork, a version shim, EGL rendering). Without GPU/regression coverage, regressions in any of those land silently.

## Suggested approach

- A `gpu`-marked smoke test (construct / reset / step / render on a couple of tasks), gated on RoboCasa asset availability (skip cleanly when assets aren't present, since they're a ~5-10 GB runtime download).
- Decide how/whether to cache the assets for the GPU CI runners.
- A `control_freq` wiring unit test once #380 lands (config `fps` -> `gym_kwargs["control_freq"]` -> sim), mirroring `test_libero_control_freq.py`.

## References

- `tests/envs/test_robocasa.py`, `.github/workflows/gpu_test.yml`, `.github/workflows/regression_test.yml`
- related: #380 (fps wiring)

Issue #384, opened 2026-06-03
RoboCasa: subgoal-image conditioning (pi07) is LIBERO-only


shuheng-liu:
## Problem

Eval-time subgoal-image conditioning (the `subgoal{k}` inputs consumed by pi07 low-level / pi07-paligemma policies) is wired only for LIBERO. `make_subgoal_generator` early-returns for anything else:

```python
if not isinstance(cfg.env, LiberoEnv) or cfg.env.subgoal_source is None:
    return None
```

So evaluating a subgoal-conditioned policy on RoboCasa silently falls back to the policy's missing-key default (zero subgoals, mask off) — the subgoal channel is effectively dead for RoboCasa.

## Why it matters

pi07 low-level / pi07-paligemma can't be evaluated faithfully on RoboCasa, since they're trained with subgoal conditioning and would see none at eval.

## Suggested approach

- Generalize the subgoal generator beyond LIBERO — either a `RoboCasaLastFrameSubgoalGenerator`, or make the generator env-agnostic over any `subgoal_source` v2.1 LeRobot dataset and add a `subgoal_source` field to `RoboCasaEnv`.
- Update `make_subgoal_generator` to dispatch per env type.

## References

- `src/opentau/scripts/eval.py:620` (`make_subgoal_generator`)
- `src/opentau/envs/subgoal.py` (`LiberoLastFrameSubgoalGenerator`, `SubgoalImageGenerator`)

Issue #383, opened 2026-06-03
RoboCasa: verify eval seed/split reproduces the benchmark eval protocol


shuheng-liu:
## Problem

LIBERO pins canonical per-task initial states (`init_states` + `set_init_state`, sourced from the benchmark demos) so eval runs the same conditions as the benchmark and success rates are directly comparable to published numbers.

RoboCasa instead derives a per-worker seed (`worker_seed = seed + episode_index` in `RoboCasaEnv.reset`) and lets `RoboCasaGymEnv` sample layout/style from it, plus the `split` (`pretrain`/`target`) selection. Seed-based held-out scenes may well be RoboCasa's intended protocol (it's how upstream LeRobot's wrapper works) — but we have **not verified** that our seeding + split reproduces RoboCasa's official eval configurations.

## Why it matters

If our seed/split scheme doesn't match the benchmark harness, RoboCasa success rates aren't comparable to published results, and runs aren't reproducible against the reference.

## Suggested approach

- Confirm the `split` semantics and seed protocol against RoboCasa's eval harness; document the mapping.
- Decide whether to pin specific eval scene configurations (the `init_states` analogue) for reproducibility, and whether the eval-time per-rank seed offset interacts correctly.

## References

- `RoboCasaEnv.reset` (`worker_seed`) in `src/opentau/envs/robocasa.py`; seeding in `src/opentau/scripts/eval.py`
- cf. LIBERO `init_states` (`src/opentau/envs/libero.py`, `src/opentau/envs/factory.py`)

Issue #382, opened 2026-06-03
RoboCasa: per-task episode horizons (TASK_SUITE_MAX_STEPS equivalent)


shuheng-liu:
## Problem

`RoboCasaEnv` uses a single `episode_length` (default `1000`) as `_max_episode_steps` for every task. RoboCasa spans ~65 atomic tasks (e.g. `CloseFridge`, single-skill, short) and ~300 composite tasks (multi-step, long), whose natural horizons differ by an order of magnitude. A flat 1000 is too long for atomic tasks (wastes rollout time, and "any-step success" can be inflated) and potentially too short for long composites.

LIBERO tunes this per suite via `TASK_SUITE_MAX_STEPS` (`libero_spatial=280`, `libero_10=520`, ...).

## Suggested approach

- A per-task or per-group (atomic / composite / pretrain) max-steps mapping for RoboCasa, defaulting sensibly and overridable from the config.
- Consider sourcing horizons from RoboCasa's own task metadata if available.

## References

- `_max_episode_steps` in `src/opentau/envs/robocasa.py`
- cf. `TASK_SUITE_MAX_STEPS` in `src/opentau/envs/libero.py`

PR #386, opened 2026-06-03
feat(envs): add RoboCasa365 simulated-eval environment alongside LIBERO

### LINKED ISSUE #381: RoboCasa: eval.recording_root uses the LIBERO-hardcoded dataset recorder
## Problem

`eval_policy_all` records rollouts to a dataset unconditionally via the LIBERO recorder:

```python
if cfg.eval.recording_root is not None:
    ...
    consolidate_task_result(aggregate_task_results(task_results), ...)
```

`aggregate_task_results` / `consolidate_task_result` come from `src/opentau/utils/libero_dataset_recorder.py`, which hardcodes a `LIBERO_TASKS` list (40 LIBERO task names) and LIBERO-shaped metadata. Setting `eval.recording_root` with a **RoboCasa** env therefore mislabels / misbuilds the recorded dataset (RoboCasa tasks aren't in `LIBERO_TASKS`, and the obs/action layout differs).

The shipped RoboCasa example config deliberately omits `recording_root`, so the default path is safe — but it's a latent footgun and blocks RECAP-style rollout capture for RoboCasa.

## Suggested approach

- Make the recorder env-aware: dispatch on `cfg.env` type, or guard the recorder to LIBERO and raise a clear error for other envs.
- Longer term, add a RoboCasa rollout recorder analogous to `libero_dataset_recorder` if RoboCasa self-training is wanted.

## References

- `src/opentau/scripts/eval.py:893` (the `recording_root` branch in `eval_policy_all`)
- `src/opentau/utils/libero_dataset_recorder.py` (`LIBERO_TASKS`, `aggregate_task_results`, `consolidate_task_result`)

shuheng-liu:
## What this does

Adds **RoboCasa365** (kitchen manipulation, robosuite/MuJoCo) as a second simulated-eval environment alongside LIBERO, and makes it a first-class `uv` extra. RoboCasa plugs into the same environment-agnostic eval pipeline as LIBERO — parallel vectorized envs, per-task success rates, and `grid_summary` eval videos to wandb — with no changes to `eval.py` / `train.py`.

Highlights:
- **`src/opentau/envs/robocasa.py`** — `RoboCasaEnv(gym.Env)` wrapper + `create_robocasa_envs()`, ported from upstream LeRobot's RoboCasa and reshaped to OpenTau's LIBERO multi-task pattern: the 3 PandaOmron cameras map to `camera0/1/2`, `info["success"]` is bridged to `is_success`, tasks shard round-robin across accelerator ranks, and the `robocasa`/`robosuite` import is deferred so the module (and the CPU suite) loads without the sim installed. 12-D action / 16-D state / 3 cameras.
- **Config + factory** — `RoboCasaEnv` draccus config (`env.type=robocasa`, task-group shortcuts + `split`) and `make_envs` dispatch. Both LIBERO and RoboCasa return `dict[group][task_id] -> VectorEnv`, so the eval pipeline gives per-task success + grid videos for free.
- **Packaging — `uv sync --extra robocasa`** co-installs with `libero` on a shared robosuite stack. Two non-obvious things make it resolve: robocasa needs `MujocoEnv(load_model_on_init=...)`, added on robosuite **master** *after* the 1.5.2 PyPI release, so `[tool.uv.sources]` repins `robosuite` to a master commit that still self-reports `1.5.2` (LIBERO re-validated on it); and robocasa is pulled from the `shuheng-liu/robocasa` packaging fork that drops upstream's `lerobot==0.3.3` / `tianshou` / `opencv-python` / `hidapi` deps and loosens its `numpy`/`numba`/`scipy`/`mujoco` pins + import-time version asserts (uv can't `--no-deps` a single package in a lock). Kitchen assets remain a separate runtime download.
- **`fix(eval)`** — `eval_policy_all` now raises a clear error if `eval.recording_root` (the LIBERO-only dataset recorder) is set with a non-LIBERO env, instead of silently mislabeling rollouts. Fixes #381.

RoboCasa is eval-validated, but the train→eval loop and a few fidelity/parity features aren't there yet — tracked as follow-ups: #379 (training data + 12-DoF projection for meaningful eval), #380 (`fps`→`control_freq` wiring), #382 (per-task horizons), #383 (eval seed/split protocol parity), #384 (subgoal conditioning), #385 (GPU/regression CI).

## How it was tested

- **CPU suite (CI-equivalent, `-m "not gpu"`):** `tests/envs/` + `tests/scripts/test_eval.py` → 142 passed. New: `tests/envs/test_robocasa.py` (config registration, factory dispatch shape, per-rank task sharding, version shim, pure helpers — all mock-based, no sim) and `tests/scripts/test_eval.py::test_eval_policy_all_rejects_recording_root_for_non_libero_env`.
- **Packaging:** regenerated `uv.lock` (resolved 274 packages, no conflicts); verified a fresh `uv sync --extra robocasa` installs and imports `robosuite` + `robocasa` + `opentau.envs.robocasa` on Linux.
- **Real sim (a Linux GPU box, headless EGL):** end-to-end against the actual robosuite/MuJoCo sim — `RoboCasaEnv` construct/reset/step/render (`CloseFridge`: language, `camera0/1/2` 256x256x3, 16-D `agent_pos`, success bridge), parallel vectorized envs via `make_envs`, and `grid_summary.mp4` + the `Eval Videos/CloseFridge_0` wandb key produced via eval's own `create_grid_summary_video` / `collect_grid_summary_videos`. Also confirmed LIBERO still constructs+steps on the shared robosuite-master commit.

CPU, no sim needed:
```bash
uv run --extra dev pytest -sx tests/envs/test_robocasa.py
uv run --extra dev pytest -sx tests/scripts/test_eval.py::test_eval_policy_all_rejects_recording_root_for_non_libero_env
```
Full sim eval (Linux + GPU; the kitchen assets are a separate ~5-10GB download):
```bash
uv sync --extra robocasa
python -m robocasa.scripts.download_kitchen_assets
MUJOCO_GL=egl opentau-eval --accelerate-config configs/examples/accelerate_ddp_config.yaml --config_path=configs/examples/pi05_robocasa_eval_config.json
```

## Checklist

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -43,13 +43,15 @@ Dependency management is **`uv` (>= 0.8.4) only** — `pyproject.toml`/`uv.lock`
 
 ```bash
 uv sync --extra dev --extra libero          # standard dev setup (matches CI)
-uv sync --all-extras                         # everything (libero + urdf now co-install on numpy 2.x)
+uv sync --all-extras                         # everything (libero + urdf co-install on the shared robosuite 1.5.2 / numpy 2.x stack)
 source .venv/bin/activate
 ```
 
 Re-run `uv sync` whenever `pyproject.toml`/`uv.lock` change. Add deps with `uv add <pkg>`; lock with `uv lock`.
 
-Installable extras: `dev` (pre-commit, sphinx, pytest), `libero` (sim env — pulls a forked LIBERO from `shuheng-liu/LIBERO`, runs on numpy 2.x + gymnasium), `urdf` (rerun ≥0.28, numpy 2.x), `trt` (TensorRT, Linux/Win x86_64 only).
+Installable extras: `dev` (pre-commit, sphinx, pytest), `libero` (sim env — pulls a forked LIBERO from `shuheng-liu/LIBERO`, on robosuite 1.5.2 + numpy 2.x + gymnasium), `urdf` (rerun ≥0.28, numpy 2.x), `trt` (TensorRT, Linux/Win x86_64 only).
+
+**RoboCasa365** (`envs/robocasa.py`) shares the `libero` extra's robosuite 1.5.2 stack but has **no first-class uv extra yet**: upstream `robocasa`'s `setup.py` hard-pins `lerobot==0.3.3` (would drag a second, conflicting LeRobot into the shared lock) plus over-strict `numba`/`numpy`/`mujoco`/`tianshou`. A packaging fork (`shuheng-liu/robocasa`) that drops those pins — mirroring the `shuheng-liu/LIBERO` / `egl_probe` forks — is the intended home for a `robocasa` extra. Until then install it manually into the libero venv: `uv pip install "git+https://github.com/robocasa/robocasa" --no-deps`, then download kitchen assets (`python -m robocasa.scripts.download_kitchen_assets`, ~10GB) and run headless with `MUJOCO_GL=egl`.
 
 ## Common commands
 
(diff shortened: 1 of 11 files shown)


shuheng-liu:
Thanks for the review! Addressed the two nits:

- **`final_info` (`robocasa.py:394`)** — dropped in 983f795. It was a vestige of the upstream LeRobot port: never read anywhere (eval consumes `info["is_success"]`/`"task"`/`"done"`, which are already set just above) and divergent from `LiberoEnv.step`. Kept the terminal auto-reset.
- **`visualization_width`/`visualization_height` (`robocasa.py:235`)** — keeping as-is, for parity with `LiberoEnv` (same inert fields) and the upstream LeRobot `RoboCasaEnv` config. They're genuinely inert today because `render()` returns the obs-resolution camera; wiring a separate higher-res render path is worth doing for *both* envs but is out of scope here (good follow-up candidate).

PR #387, opened 2026-06-03
feat(envs): auto-download RoboCasa assets to a relocatable external store

shuheng-liu:
## What this does

Running a RoboCasa env (`env.type=robocasa`) crashed on the first scene build with `FileNotFoundError` / `ValueError: Probabilities contain NaN`: robocasa's multi-GB kitchen asset packs are not installed by `uv sync` and robocasa ships no auto-download — and even when downloaded, they land *inside* the venv and are lost whenever it is recreated.

This makes the assets **download automatically, once**, into a **venv-external, relocatable** store:

- `_ensure_robocasa_assets` (rank-0 gated, with a cross-rank barrier so distributed eval stays in NCCL lockstep) seeds the wheel-shipped stubs, then downloads exactly the packs an env's `obj_registries` need — `textures`, `tex_generative`, `fixtures_lw`, and the per-registry object packs (e.g. `objs_lw`) — into the store, guarded by per-pack marker files.
- It then **symlinks** `<venv>/robocasa/models/assets` → the store. A symlink (rather than patching `robocasa.models.assets_root`) is required because robocasa builds its object registry by scanning `models/assets` **at import time**, so the relocation must be in place *before* robocasa is imported — done in the version shim, in every process including `AsyncVectorEnv` spawn workers.
- The store defaults to `$HF_OPENTAU_HOME/robocasa/assets`, overridable via the `ROBOCASA_ASSETS_ROOT` env var or the new `env.assets_root` config field. Because it lives outside the venv, ephemeral `uv` envs no longer lose or re-download assets.

Adds `RoboCasaEnv` config fields `assets_root` and `auto_download_assets` (default on), forwarded through the env factory, plus a `opentau.scripts.download_robocasa_assets` warm-up script for pre-fetching on a login node before a multi-rank eval.

Label: 🗃️ Feature

## How it was tested

- Added 14 CPU tests + 1 `@pytest.mark.gpu` end-to-end test in `tests/envs/test_robocasa.py` (env-var resolver, symlink/relink, needed-packs mapping, rank-0-gated download with barrier, factory wiring). `pytest -m "not gpu"` → 45 passed locally; 50 with `robocasa` + `libero` installed (including `test_factory`).
- End-to-end on a CUDA box (sim + EGL): the `@pytest.mark.gpu` test auto-downloads the lightwheel pack set (~4.4G) into the external store and runs a real `CloseFridge` reset+step — **passed**. Confirmed the venv's `models/assets` is a symlink into the external store (outside `site-packages`).
- Ephemeral / idempotent: a second run did **0 re-downloads** (marker-guarded) and passed in ~12s.
- `pre-commit run --all-files` clean (ruff, ruff-format, bandit, …).

```bash
pytest -sx tests/envs/test_robocasa.py -m "not gpu"
```

On a CUDA box with the sim installed (downloads ~4.4G of assets to `$HF_OPENTAU_HOME/robocasa/assets` on first run, cached thereafter):

```bash
MUJOCO_GL=egl pytest -sx tests/envs/test_robocasa.py -m gpu
```

Or pre-fetch the assets without running an eval:

```bash
python -m opentau.scripts.download_robocasa_assets
```

## Checklist

### DIFF
--- a/src/opentau/envs/configs.py
+++ b/src/opentau/envs/configs.py
@@ -345,6 +345,11 @@ class RoboCasaEnv(EnvConfig):
         obj_registries: Object-mesh registries to sample assets from. Defaults to
             ``["lightwheel"]`` (the pack the asset downloader ships by default);
             add ``"objaverse"`` only after downloading that ~30GB pack.
+        assets_root: Directory to store/read RoboCasa kitchen assets, kept outside the
+            (ephemeral) uv venv. ``None`` resolves to the ``ROBOCASA_ASSETS_ROOT`` env
+            var, else ``HF_OPENTAU_HOME/robocasa/assets``.
+        auto_download_assets: If ``True`` (default), the asset packs ``obj_registries``
+            needs are downloaded automatically (once) on first env build.
         features: Mapping from logical feature names to ``PolicyFeature`` definitions.
         features_map: Mapping from environment keys to standardized OpenTau keys.
     """
@@ -361,6 +366,8 @@ class RoboCasaEnv(EnvConfig):
     visualization_width: int = 512
     split: str | None = None
     obj_registries: list[str] = field(default_factory=lambda: ["lightwheel"])
+    assets_root: str | None = None
+    auto_download_assets: bool = True
     features: dict[str, PolicyFeature] = field(
         default_factory=lambda: {
             "action": PolicyFeature(type=FeatureType.ACTION, shape=(12,)),
--- a/src/opentau/envs/factory.py
+++ b/src/opentau/envs/factory.py
@@ -95,6 +95,8 @@ def make_envs(
             env_cls=env_cls,
             episode_length=cfg.episode_length,
             obj_registries=tuple(cfg.obj_registries),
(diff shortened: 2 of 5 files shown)


shuheng-liu:
Thanks for the review — addressed all three in `6de15ac`:

- **`:339` (shared-FS assumption)** — documented in `_ensure_robocasa_assets` that the global-main download assumes `assets_root` is on storage shared by all nodes; point `ROBOCASA_ASSETS_ROOT` at a shared path when `HF_OPENTAU_HOME` is node-local. Kept the global-`is_main_process` gating to match the `lerobot_dataset` download precedent (single-node eval is the common case here).
- **`:151` (url-helper coverage)** — added CPU unit tests for `_direct_download_url` (`TestDirectDownloadUrl`: base rewrite + trailing-slash).
- **`:173` (retry / atomicity)** — `_download_and_extract_zip` now retries 3× (matching robocasa's original downloader) and raises with the last error if all attempts fail. The pack done-marker is still only written after a clean extract, so a partial/failed extract is simply retried and overwritten.

PR #391, opened 2026-06-04
feat(datasets): add read-only support for LeRobot v3.0 datasets

shuheng-liu:
## What this does

Adds native **read-only** support for **LeRobot v3.0** ("file-consolidation") datasets, which OpenTau could not load before (`CODEBASE_VERSION = "v2.1"`; `get_safe_version` raised `ForwardCompatibilityError` on a v3.0-only repo). v3.0 packs many episodes into a single consolidated parquet/mp4 and moves metadata to parquet (`meta/episodes/**/*.parquet`, `meta/tasks.parquet`) instead of one file per episode.

The change is deliberately isolated: v3.0 metadata is loaded into the **same in-memory shapes** the v2.1 path already produces (`meta.episodes`, `meta.tasks`, `meta.episodes_stats`), so everything downstream (sampler, mixture, `episode_data_index`, normalization) is unchanged. Only these surfaces branch on the new format:

- **Version gating** — a `READ_CODEBASE_VERSION = "v3.0"` read ceiling in `get_safe_version` (additive; v2.x/v1.x resolution unchanged) plus a same-major guard so the v2.0 global-stats warning doesn't mis-fire on v3.0.
- **Metadata loaders** (`datasets/utils.py`) — `load_tasks_v30` / `load_episodes_v30` / `load_episodes_stats_v30`, reading the parquet metadata and reconstructing per-episode stats from the flattened `stats/*` columns.
- **Path accessors** — `get_data_file_path` / `get_video_file_path` resolve the consolidated file from each episode's `data/chunk_index`/`data/file_index` and `videos/{key}/...`.
- **Loading** — `load_hf_dataset` reads the deduped consolidated data files and, for an episode subset, compacts via an `episode_index` mask + `Dataset.select` (memory-mapping preserved), reusing the existing `episode_data_index`.
- **Video** — a per-episode `from_timestamp` offset so frames decode correctly from the consolidated mp4 (0.0 for v2.1, so that path is byte-for-byte unchanged).

The **write path is unchanged** — `CODEBASE_VERSION` stays `v2.1`; creating/saving datasets in the v3.0 layout is out of scope.

Label: 🗃️ Feature

## How it was tested

- Added `tests/datasets/test_datasets_v30.py` (13 CPU tests) over a synthetic on-disk v3.0 dataset: metadata loading, full- and subset-load `episode_data_index` alignment, `__getitem__` row correctness, the consolidated-mp4 timestamp offset, no-video and multi-task variants, version gating, and per-episode image-stat `(3,1,1)` reconstruction.
- Full `pytest -m "not gpu" tests/datasets/` → **526 passed, 7 skipped** (no v2.x regression).
- Loaded real v3.0 Hub datasets end-to-end on a GPU dev box:
  - `lerobot/pusht_keypoints` (`revision="v3.0"`, no video) — metadata + data + subset compaction (`episodes=[0,1,2]` → aligned `episode_data_index`).
  - `lerobot/droid_1.0.1` (`revision="v3.0"`, ~412 GB, 95,658 episodes, 3 cameras) — a single episode (`episodes=[1]`) pulled only its consolidated chunk (~2.2 GB / 4 files, not the full 412 GB), decoded all three camera videos, and queried the consolidated mp4 at the correct per-episode `from_timestamp` offset (11.133 s). This also surfaced a real bug — per-episode image/video stats round-trip from the parquet as deeply-nested object arrays, breaking `aggregate_stats` — now fixed (`_deep_float_array`) with a dedicated regression test.

Determinism (CLAUDE.md rule 3): not triggered — no changes to the training loop / modeling / optim / sampler, and the v2.x load path is unchanged (covered by the 526 passing dataset tests).

```bash
pytest -sx tests/datasets/test_datasets_v30.py
```

```python
# Load any v3.0 dataset directly. Request revision="v3.0" explicitly when the
# repo also has a v2.1 tag (the default target resolves to v2.1 when present).
from opentau.datasets.lerobot_dataset import LeRobotDatasetMetadata

meta = LeRobotDatasetMetadata("lerobot/pusht_keypoints", revision="v3.0")
print(meta._is_v30, meta.total_episodes)  # True 206
```

## Checklist

### DIFF
--- a/src/opentau/datasets/lerobot_dataset.py
+++ b/src/opentau/datasets/lerobot_dataset.py
@@ -141,9 +141,8 @@ from opentau.datasets.utils import (
     is_valid_version,
     load_advantages,
     load_episodes,
+    load_episodes_and_stats_v30,
     load_episodes_stats,
-    load_episodes_stats_v30,
-    load_episodes_v30,
     load_info,
     load_stats,
     load_tasks,
@@ -421,8 +420,8 @@ class LeRobotDatasetMetadata(DatasetMetadata):
             # carries the extra file-mapping columns the v3.0 path accessors and
             # video offset read; everything else downstream is unchanged.
             self.tasks, self.task_to_task_index = load_tasks_v30(self.root)
-            self.episodes = load_episodes_v30(self.root)
-            self.episodes_stats = load_episodes_stats_v30(self.root)
+            # Parse the episode-metadata shards once for both episodes + stats.
+            self.episodes, self.episodes_stats = load_episodes_and_stats_v30(self.root)
             self.stats = load_stats(self.root)
         else:
             self.tasks, self.task_to_task_index = load_tasks(self.root)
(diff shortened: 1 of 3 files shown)


shuheng-liu:
Thanks for the review — addressed in 0bb89f3:

- **Double metadata read** (`utils.py`) — added `load_episodes_and_stats_v30`, which parses the `meta/episodes/**` shards once and splits the result into episodes + per-episode stats (the individual loaders keep an optional pre-read-table param for standalone use). Halves the metadata-load cost on the DROID-scale path.
- **Row-order safety** (`lerobot_dataset.py`) — `_load_hf_dataset_v30` now scans the loaded `episode_index` column and asserts it is non-decreasing (ascending-by-episode) before building `episode_data_index`, so a malformed consolidated layout fails loudly instead of silently desyncing per-dataset metrics (CLAUDE.md §5). This replaces the previous whole-load scan-skip.
- **`readable_newer` nit** — added a comment clarifying that in practice only v3.0 and the v2.x minors fall in that window, and a hypothetical newer v2.x minor would be read via the v2.1-compatible path (consistent with how `check_version_compatibility` already treats all v2.x).
- **Lint nits** — single import style (string-based `monkeypatch.setattr` targets) and explicit `_ =` discards in the tests.

PR #396, opened 2026-06-04
fix(train): broadcast output_dir so all ranks share one checkpoint tree

shuheng-liu:
## What this does

🐛 **Bug fix** — makes `output_dir` rank-consistent so every rank of a multi-rank run writes its checkpoint shard into the **same** tree.

**Symptom.** A DeepSpeed ZeRO-2 run on 8 GPUs produced checkpoints that were silently missing one rank's optimizer shard (`bf16_zero_pp_rank_<r>_mp_rank_00_optim_states.pt`) **and** that rank's `random_states_<r>.pkl` — every save, the same rank. `zero_to_fp32` then refused to consolidate (`Expected 8 of '*_optim_states.pt' ... but found 7`), and the checkpoint could not be resumed. There was no error and no hang at save time; the run trained normally.

**Root cause.** `TrainPipelineConfig.validate()` assigns the default `output_dir` from a per-process wall-clock timestamp:

```python
now = dt.datetime.now()
self.output_dir = Path("outputs/train") / f"{now:%Y-%m-%d}/{now:%H-%M-%S}_{job_name}"
```

`validate()` runs independently in every rank (each process parses the config), so a launch whose process startup straddles a one-second boundary leaves ranks disagreeing on `output_dir` — e.g. one rank computes `…-49` while the other seven compute `…-50`. Each rank then writes its optimizer shard and RNG state into its **own sibling directory**. The NCCL collectives inside DeepSpeed's `save_checkpoint` are path-independent, so nothing errors or hangs; the per-rank files are simply split across two trees, and each tree looks like it is missing a rank. (Confirmed in practice: the two sibling trees were exactly complementary — 7 ranks in one, the 8th in the other — so no data was ever lost, only split.)

**Fix.** Broadcast rank 0's `output_dir` to all ranks immediately after the `Accelerator` is constructed (it does not consume `output_dir`, so doing it there is safe) and before `output_dir` is first used, so every rank shares one output tree. Also adds a cheap, backend-agnostic post-save tripwire: rank 0 verifies that the number of `random_states_*.pkl` files equals `world_size` and logs a loud error otherwise — so any future partial save surfaces immediately instead of silently at resume/consolidation time.

## How it was tested

- `pre-commit run --files src/opentau/scripts/train.py` — clean (ruff, ruff-format, pyupgrade, bandit, …).
- `pytest -m "not gpu" -n auto tests/scripts/test_train.py tests/scripts/test_outlier_logging.py` — **29 passed** (the CPU suite `cpu_test.yml` gates on).
- **8-GPU ZeRO-2 forced-divergence smoke** (reproducing the exact failure mode): with a temporary patch that forced each rank's `output_dir` to diverge **before** the broadcast (`…_ODTEST0` … `…_ODTEST7`), all 8 ranks converged to rank 0's directory **after** the broadcast, exactly one output tree was created, and its checkpoint contained **all 8** optimizer shards + **all 8** RNG-state files; the new integrity tripwire stayed silent. A control run without the forced divergence likewise produced a complete 8-shard checkpoint.

```bash
pytest -sx tests/scripts/test_train.py
```

The fix is exercised by any multi-rank run — every rank now resolves and writes to the same `output_dir`. To observe the failure it prevents, temporarily force divergence right before the broadcast in `train()`:

```python
cfg.output_dir = Path(f"{cfg.output_dir}_{accelerator.process_index}")  # simulate the timestamp race
```

Without the fix this yields `world_size` sibling directories each holding one shard; with the fix all ranks collapse onto rank 0's directory and the checkpoint is complete.

## Checklist

> Note: `train.py` is touched, but the change is checkpoint-path coordination plus a logging tripwire — no model math / optimizer / sampler / numerics impact, so the policy box is left unchecked (determinism is unaffected). An 8-GPU ZeRO-2 smoke was nonetheless run to verify the fix end-to-end (see *How it was tested*).

### DIFF
--- a/src/opentau/scripts/train.py
+++ b/src/opentau/scripts/train.py
@@ -56,6 +56,7 @@ from opentau.utils.accelerate_utils import set_proc_accelerator
 from opentau.utils.logging_utils import AverageMeter, MetricsTracker
 from opentau.utils.random_utils import set_seed
 from opentau.utils.train_utils import (
+    find_missing_rng_state_ranks,
     get_step_checkpoint_dir,
     get_step_identifier,
     load_training_state,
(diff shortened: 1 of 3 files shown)


shuheng-liu:
Addressed the two non-blocking review nits in 4dd9b1e:

- **`train.py` (no unit test)** — factored the tripwire's completeness check into `find_missing_rng_state_ranks(checkpoint_dir, world_size)` in `utils/train_utils.py` (it now returns the exact missing rank indices, not just a count) and added a CPU unit test in `tests/utils/test_train_utils.py` mirroring the failure case (write ranks `0–5,7` → asserts `[6]`; also covers the complete, unrelated-file, and missing-directory cases).

- **tripwire logs vs. raises** — kept it as an error log (now more prominent: `CHECKPOINT INCOMPLETE: … rank(s) [...]`) rather than raising, by design. An incomplete checkpoint already hard-fails loudly at **resume** — accelerate `load_state` errors on the missing optim shard and `zero_to_fp32` refuses to consolidate — so the true gate exists. Raising mid-run off a rank-0 directory listing would risk false positives from NFS `readdir` lag (especially multi-node), which would kill an otherwise-healthy long run. The prominent error log plus the hard resume-time failure is the intended defense-in-depth balance.

PR #397, opened 2026-06-04
feat(datasets): default to main/master when no codebase-version tag

shuheng-liu:
## What this does
`get_safe_version` resolves a dataset's Hub revision from its codebase-version tags. For a repo that carries **no version tags at all** (only a `main` branch — common for community LeRobot / RoboCasa conversions), it raised `RevisionNotFoundError`, so the dataset could not be loaded directly from the Hub.

This PR makes the resolver **fall back to the repo's default branch — `main`, then `master`, in that order — when no `revision` is explicitly requested and the repo has no codebase-version tag available**. An explicitly pinned `revision` keeps the strict behaviour (still raises), so nothing silently changes for callers that pass a revision.

Details:
- New `allow_branch_fallback` flag on `get_safe_version` (default `False`), set to `True` by the two read-path call sites *only when `revision` is unset* (`LeRobotDatasetMetadata.__init__` and `LeRobotDataset.__init__`).
- New `get_repo_branches(repo_id)` helper to look up the repo's branch refs; the first of `("main", "master")` present is returned.
- When fallback is allowed but neither branch exists (and there are no version tags), the original `RevisionNotFoundError` is still raised.

🗃️ Feature

## How it was tested
- Added 7 tests in `tests/datasets/test_datasets_v30.py`:
  - Unit (`get_safe_version`): falls back to `main`; falls back to `master` when `main` is absent; prefers `main` over `master`; raises without `allow_branch_fallback` (explicit revision); raises when neither default branch exists.
  - End-to-end (`LeRobotDatasetMetadata`): an untagged repo with no local cache resolves its revision to `main` when no revision is pinned, and still raises for an explicitly pinned revision.
- `uv run pytest -m "not gpu" tests/datasets/test_datasets_v30.py` → 20 passed.
- `pre-commit run --files ...` clean (ruff, ruff-format, pyupgrade, bandit, …).
- Not a policy change (touches `datasets/` only), so the GPU/regression suites were not required.

```bash
pytest -sx tests/datasets/test_datasets_v30.py -k "untagged or main_over_master or no_default_branch"
```

## Checklist

### DIFF
--- a/src/opentau/datasets/lerobot_dataset.py
+++ b/src/opentau/datasets/lerobot_dataset.py
@@ -388,7 +388,9 @@ class LeRobotDatasetMetadata(DatasetMetadata):
             self.load_metadata()
         except (FileNotFoundError, NotADirectoryError):
             if is_valid_version(self.revision):
-                self.revision = get_safe_version(self.repo_id, self.revision)
+                self.revision = get_safe_version(
+                    self.repo_id, self.revision, allow_branch_fallback=not revision
+                )
 
             # In distributed training, only rank 0 downloads to avoid race conditions
             # where other ranks read metadata before the download has finished.
@@ -1597,7 +1599,9 @@ class LeRobotDataset(BaseDataset):
             self.hf_dataset = self.load_hf_dataset()
         except (AssertionError, FileNotFoundError, NotADirectoryError):
             if is_valid_version(self.revision):
-                self.revision = get_safe_version(self.repo_id, self.revision)
+                self.revision = get_safe_version(
+                    self.repo_id, self.revision, allow_branch_fallback=not revision
+                )
             self.download_episodes(download_videos)
             self.hf_dataset = self.load_hf_dataset()
 
--- a/src/opentau/datasets/lerobot_dataset.py
+++ b/src/opentau/datasets/lerobot_dataset.py
@@ -1502,9 +1502,12 @@ class LeRobotDataset(BaseDataset):
 
         self.root.mkdir(exist_ok=True, parents=True)
 
-        # Load metadata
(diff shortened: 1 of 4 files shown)


shuheng-liu:
Thanks — the blocking catch on the `LeRobotDataset` path is spot on. Addressed in e275ee7.

**1. Dead fallback on the full `LeRobotDataset` path (blocking) — fixed.**
`LeRobotDataset.__init__` now forwards the *original* `revision` (which may be `None`) to `LeRobotDatasetMetadata`, instead of the already-`CODEBASE_VERSION`-coerced `self.revision`. The metadata constructor applies the default itself, so it again sees an unset revision and enables `allow_branch_fallback`. Untagged repos now resolve to `main`/`master` at metadata construction rather than raising.

**2. Missing E2E coverage on `LeRobotDataset` (blocking) — added.**
`test_dataset_untagged_repo_loads_via_branch_fallback` exercises the full `LeRobotDataset` path on an untagged repo. I verified it **fails without the fix** (raises `RevisionNotFoundError` at metadata construction) and passes with it.

**3. Second `list_repo_refs` call in `get_repo_branches` (suggestion) — keeping as is.**
The extra refs fetch only happens on the cold fallback path (a repo with zero version tags), once per dataset load. Collapsing `get_repo_versions` + `get_repo_branches` into a single refs fetch inside `get_safe_version` would remove the two independently-monkeypatchable seams the unit tests rely on (`get_repo_versions` is patched by the existing v2.1/v3.0 tests; `get_repo_branches` by the new fallback tests). The clarity/testability seems worth more than saving one round-trip on a rare path — happy to revisit if you'd prefer otherwise.

PR #398, opened 2026-06-04
feat(envs): render all cameras in eval rollout videos

shuheng-liu:
## What this does

Eval rollout videos (and the `grid_summary.mp4`) showed only a single camera, even though the policy is fed multiple cameras. The sole video frame source is `env.render()`, which returned one camera. This makes both sim envs' `render()` return **all configured cameras concatenated side-by-side**, so per-episode rollout videos — and the grid summary and `wandb.Video` built from them — show every camera the policy sees, making rollouts actually debuggable.

- `RoboCasaEnv.render()` / `LiberoEnv.render()` now concatenate every `camera_name` view (left→right in config order) via `einops.rearrange("n h w c -> h (n w) c")`. LIBERO keeps its `render_cam` override as a single-camera escape hatch; RoboCasa caches the per-camera frames in `_format_raw_obs` since its underlying `render()` is single-view.
- `create_grid_summary_video` now loads episode clips with `imageio.mimread(..., memtest=False)`: the wider multi-camera frames exceed imageio's default 256 MB per-read cap, which otherwise aborts grid creation.

No change to the rollout/record path in `eval.py`, the grid tiler's layout, or the wandb logging — they read frame dimensions dynamically and adapt to the wider frames.

🗃️ Feature

## How it was tested

- pre-commit (ruff lint + format, pyupgrade, bandit, …) passes on the changed files; `pytest -m "not gpu" tests/scripts/test_eval.py` green (3 passed).
- Ran a RoboCasa **CloseFridge** sim eval on a GPU dev box with the change: per-episode rollout mp4s are now `(256, 768, 3)` — 3 cameras × 256 px wide — vs `(256, 256, 3)` before, and `grid_summary.mp4` renders each episode row with all three camera views. The eval completed normally; rendering is visualization-only and does not touch the rollout, policy, or RNG, so success rates are unaffected.

```bash
```

Then any sim eval with rendering enabled records all cameras — an eval config with `env.type=robocasa` (or `libero`) and `eval.max_episodes_rendered>0` produces rollout videos whose width is `n_cameras × per_camera_width` (set `env.render_cam` on LIBERO to keep a single camera).

## Checklist

### DIFF
--- a/src/opentau/envs/libero.py
+++ b/src/opentau/envs/libero.py
@@ -24,6 +24,7 @@ from functools import partial
 from pathlib import Path
 from typing import Any
 
+import einops
 import gymnasium as gym
 import numpy as np
 import torch
@@ -225,13 +226,15 @@ class LiberoEnv(gym.Env):
         )
 
     def render(self) -> np.ndarray:
-        r"""Render the environment and return a numpy array representing the RGB camera.
-        If `self.render_cam` is set, use that camera; otherwise, use the first camera."""
+        r"""Render an RGB array for video recording. If ``self.render_cam`` is set,
+        use that single camera; otherwise concatenate all configured cameras
+        side-by-side so eval rollout videos show every camera the policy sees."""
         raw_obs = self._env.env._get_observations()
         cams: dict[str, np.ndarray] = self._format_raw_obs(raw_obs)["pixels"]
-        # if `self.render_cam` is not set, use the first camera
-        render_cam = self.render_cam or next(iter(cams))
-        return cams[render_cam]
+        if self.render_cam:
+            return cams[self.render_cam]
+        frames = list(cams.values())  # each (H, W, 3), camera_name order
+        return einops.rearrange(np.stack(frames), "n h w c -> h (n w) c")
 
     def _make_envs_task(self, task_suite: Any, task_id: int = 0):
         task = task_suite.get_task(task_id)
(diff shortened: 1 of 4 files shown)


shuheng-liu:
Thanks for the review. Addressed each item:

- **Add a test (suggestion):** done in ce2578a — `TestRenderMultiCamera` in `tests/envs/test_robocasa.py`, CPU-only. It asserts the `(H, n·W, 3)` all-cameras concatenation + left→right ordering, single-camera passthrough (n=1 isn't widened), and the pre-first-observation `_env.render()` fallback.

- **`.values()` duplicates a frame under a custom 1→many `camera_name_mapping` (nit):** keeping as-is, by design. The strip renders one panel per *policy-input* slot (`camera{i}`). When a mapping deliberately feeds one physical camera into multiple slots, the policy genuinely receives it in each, so showing it in each panel is faithful to the model's input rather than a rendering bug. It's a no-op for the default 1:1 mapping.

- **Grid-builder holds all clips in memory (suggestion):** real, but pre-existing — `create_grid_summary_video` already loaded every episode clip at once before this change; multi-camera frames just widen each clip ~3×. For the default eval (≤8 rendered episodes × 3 cams × 256 px ≈ 2 GB) it's comfortably fine. Streaming the tiler frame-by-frame (one frame per clip at a time) would remove the accumulation entirely, but that's an orthogonal optimization — I'd rather keep this PR focused on the visualization change and do the streaming refactor as a separate follow-up.

PR #402, opened 2026-06-05
fix(envs): pin MuJoCo EGL render to per-rank GPU for multi-GPU sim eval

shuheng-liu:
## What this does

Fixes a multi-GPU sim-eval bug where every rank rendered on the same GPU.

MuJoCo's EGL backend (`mujoco.egl.create_initialized_egl_device_display`) selects the render GPU from the `MUJOCO_EGL_DEVICE_ID` env var and falls back to **EGL device 0** when it is unset. robosuite forwards a `device_id` that `mujoco.egl` ignores, and OpenTau set `MUJOCO_EGL_DEVICE_ID` nowhere — so under multi-GPU eval (e.g. an 8-GPU job evaluating during training), **every rank rendered its sim cameras on GPU 0**. That overloads GPU 0 — which also holds rank 0's resident training state — risking OOM at the first eval, while the other GPUs render nothing. CPU (`osmesa`) rendering masked this; it only surfaces once EGL/GPU rendering is used.

Fix: a new helper `_pin_egl_render_device()`, called at the top of `make_envs`, sets `MUJOCO_EGL_DEVICE_ID` to the accelerator's `local_process_index` **before** the vec env's `AsyncVectorEnv` spawn workers are created (they inherit `os.environ`, so the worker that creates the EGL context picks up the right device). Each rank then renders on its own GPU.

It is a no-op unless `MUJOCO_GL=egl`, respects an explicit `MUJOCO_EGL_DEVICE_ID` override, and leaves the single-GPU / no-accelerator case unchanged (device 0 is correct there). This also matches what robosuite's own `binding_utils.py` assert expects (`MUJOCO_EGL_DEVICE_ID` being one of `CUDA_VISIBLE_DEVICES`).

## How it was tested

- Added `TestPinEglRenderDevice` (4 cases) in `tests/envs/test_factory.py`: sets the device id under `MUJOCO_GL=egl`, no-op under `osmesa`, respects an explicit `MUJOCO_EGL_DEVICE_ID`, and no-op when no accelerator is present. Pure env-var checks — no GPU or sim import needed.
- Full CPU env suite green: `pytest -m "not gpu" tests/envs/` → **174 passed**. The existing `make_envs` tests confirm the new call is a no-op in non-egl contexts.
- Only `src/opentau/envs/factory.py` is touched (not a policy change). The end-to-end per-GPU render spread under a real multi-GPU eval is to be confirmed on hardware; EGL already enumerates one device per GPU there.

```bash
pytest -sx tests/envs/test_factory.py::TestPinEglRenderDevice
```

```bash
pytest -m "not gpu" tests/envs/
```

## Checklist

### DIFF
--- a/src/opentau/envs/factory.py
+++ b/src/opentau/envs/factory.py
@@ -17,12 +17,14 @@
 r"""This module contains factory methods to create environments based on their configuration."""
 
 import importlib
+import os
 from functools import partial
 
 import gymnasium as gym
 
 from opentau.configs.train import TrainPipelineConfig
 from opentau.envs.configs import EnvConfig, LiberoEnv, RoboCasaEnv
+from opentau.utils.accelerate_utils import get_proc_accelerator
 
 
 def make_env_config(env_type: str, **kwargs) -> EnvConfig:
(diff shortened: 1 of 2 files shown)


shuheng-liu:
Thanks — addressed in 6e0cb5b:

- **EGL device id vs masked `CUDA_VISIBLE_DEVICES`** (inline suggestion): `_pin_egl_render_device` now indexes the local rank *through* `CUDA_VISIBLE_DEVICES`, so a masked/reordered subset like `"4,5,6,7"` maps rank 1 → physical GPU `5` — co-located with that rank's train GPU and satisfying robosuite's `MUJOCO_EGL_DEVICE_ID in CUDA_VISIBLE_DEVICES` assert. Falls back to the raw local index when `CUDA_VISIBLE_DEVICES` is unset. Covered by `test_maps_through_masked_cuda_visible_devices`. The deeper EGL-index==CUDA-index assumption (when CVD is unset) is inherent to `mujoco.egl`'s `eglQueryDevicesEXT` indexing and shared with upstream robosuite, so left as-is.
- **No assertion that `make_envs` pins before worker spawn** (nit): added `test_make_env_pins_egl_device_before_spawning_workers`, which asserts the pin runs before the vec env is constructed.

PR #404, opened 2026-06-05
fix(envs): register nvidia EGL ICD so RoboCasa sim eval renders on GPU

shuheng-liu:
## What this does

🐛 **Bug fix** — makes RoboCasa sim eval render on **GPU** (egl) in the training container, instead of failing to create a headless display (or silently falling back to CPU software rendering).

**Symptom.** Multi-GPU RoboCasa sim eval under `MUJOCO_GL=egl` crashed in the `AsyncVectorEnv` spawn workers with `robosuite/.../egl_context.py: ImportError: Cannot initialize a EGL device display ... PLATFORM_DEVICE extension`. (MuJoCo's own renderer didn't crash earlier because it silently used the Mesa **llvmpipe software** device — eval "ran" but never touched the GPU.)

**Root cause.** The training container image injects `libEGL_nvidia.so` (via `NVIDIA_DRIVER_CAPABILITIES` *graphics*) but ships only Mesa's glvnd ICD descriptor (`/usr/share/glvnd/egl_vendor.d/50_mesa.json`) — there is **no nvidia ICD descriptor**. glvnd therefore loads Mesa, whose GPU path needs `/dev/dri` render nodes the container doesn't expose (`/dev/dri/* Permission denied`). So `eglQueryDevicesEXT()` enumerates only unusable Mesa/software devices, and robosuite's per-device EGL context — which indexes that list by `MUJOCO_EGL_DEVICE_ID` (set per-rank by #402) — fails to initialize a display.

**Fix.** `_ensure_nvidia_egl_icd()`, called in `make_envs` alongside #402's `_pin_egl_render_device` before the spawn workers build their render context: when `MUJOCO_GL=egl` and no nvidia ICD is registered, synthesize a minimal glvnd descriptor pointing at the injected `libEGL_nvidia.so` and set `__EGL_VENDOR_LIBRARY_FILENAMES` so glvnd uses only nvidia EGL. `eglQueryDevicesEXT()` then returns the GPUs, and each rank renders on its own. No-op unless `MUJOCO_GL=egl` **and** the descriptor is genuinely missing — so a host with a normal nvidia EGL install (its own `*nvidia*.json`) is untouched.

## How it was tested

- `pre-commit run --files src/opentau/envs/factory.py tests/envs/test_factory.py` — clean.
- `pytest -m "not gpu" tests/envs/test_factory.py` — the new `TestEnsureNvidiaEglIcd` (5 cases: not-egl, vendor-already-set, nvidia-ICD-already-registered, no-nvidia-lib, synthesize + idempotent) and the existing `_pin_egl` tests pass. The 3 `TestMakeEnv` cases require the `libero` extra (they pass on CI `cpu_test.yml`; they fail only in a `--extra dev`-only local env — confirmed identical on `main`).
- **8-GPU in-training RoboCasa CloseFridge eval** under `MUJOCO_GL=egl` + graphics caps: with this fix all 8 ranks create **GPU** EGL contexts (`eglQueryDevicesEXT` returns the 8 GPUs), the eval renders and reports per-task success, and the run exits cleanly. Without it, the spawn workers fail with "Cannot initialize a EGL device display." Also verified the isolated robosuite `EGLGLContext` succeeds per device (0–7) under the synthesized ICD.

```bash
pytest -sx tests/envs/test_factory.py -k Egl
```

On a GPU node running the container with `MUJOCO_GL=egl` and `NVIDIA_DRIVER_CAPABILITIES=all` (graphics), any RoboCasa eval now auto-registers the nvidia EGL ICD inside `make_envs`, so robosuite renders on the GPU instead of failing or falling back to llvmpipe.

## Checklist

> Note: `envs/factory.py` is environment / render plumbing — no model math, optimizer, sampler, or numerics — so the policy box is left unchecked. GPU sim eval was nonetheless validated end-to-end (see *How it was tested*).

### DIFF
--- a/src/opentau/envs/factory.py
+++ b/src/opentau/envs/factory.py
@@ -76,19 +76,28 @@ def _ensure_nvidia_egl_icd() -> str | None:
     for vendor_dir in ("/usr/share/glvnd/egl_vendor.d", "/etc/glvnd/egl_vendor.d"):
         if glob.glob(os.path.join(vendor_dir, "*nvidia*.json")):
             return None
-    libs = sorted(glob.glob("/usr/lib/x86_64-linux-gnu/libEGL_nvidia.so.*")) or sorted(
-        glob.glob("/usr/lib/*/libEGL_nvidia.so.*")
+    # Prefer the version-stable soname symlink (``libEGL_nvidia.so.0``) over a
+    # specific versioned file, so the choice doesn't depend on lexicographic-vs-
+    # numeric ordering of multiple injected libs.
+    libs = (
+        glob.glob("/usr/lib/x86_64-linux-gnu/libEGL_nvidia.so.0")
+        or glob.glob("/usr/lib/*/libEGL_nvidia.so.0")
+        or sorted(glob.glob("/usr/lib/x86_64-linux-gnu/libEGL_nvidia.so.*"))
+        or sorted(glob.glob("/usr/lib/*/libEGL_nvidia.so.*"))
     )
     if not libs:
         return None
-    icd_dir = os.path.join(tempfile.gettempdir(), "opentau_egl")
+    # Namespace the descriptor dir by uid: on a shared multi-tenant node a second
+    # user must not hit ``PermissionError`` writing into the first user's dir (which
+    # would crash ``make_envs``). Same-user ranks share it; the write below is atomic
+    # (mkstemp + ``os.replace``) so a partial file is never visible to a spawn worker.
+    uid = getattr(os, "getuid", os.getpid)()
+    icd_dir = os.path.join(tempfile.gettempdir(), f"opentau_egl_{uid}")
     os.makedirs(icd_dir, exist_ok=True)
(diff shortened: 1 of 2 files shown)


shuheng-liu:
Addressed both comments in 150dd63:

- **`/tmp/opentau_egl` not uid-namespaced** — the descriptor dir is now `opentau_egl_<uid>` (`getattr(os, "getuid", os.getpid)()`), so a second user on a shared node writes to their own dir instead of hitting `PermissionError` in the first user's.
- **`libs[-1]` version ordering** — now prefers the version-stable `libEGL_nvidia.so.0` soname (falling back to the sorted versioned glob only if the symlink is absent), so the choice no longer depends on lexicographic-vs-numeric ordering of multiple injected libs.

shuheng-liu:
Addressed in aa54b80 — switched to `tempfile.mkdtemp()` (a unique, unpredictable, 0700 dir owned by the process), so there's no predictable path to pre-squat and no cross-user collision. Simpler too (no shared file / no atomic-write dance).

PR #405, opened 2026-06-05
feat(config): add wandb option to skip eval video logging

shuheng-liu:
## What this does
Adds an optional `wandb.disable_video` flag to `WandBConfig` that skips logging the eval grid-summary videos to Weights & Biases. It defaults to `False`, so the default behavior is unchanged — videos are still logged whenever wandb is enabled and rendering is on.

This is useful for runs that don't need the eval videos in wandb, or that bump into wandb media-storage limits, without having to disable wandb or rendering entirely.

The implementation mirrors the existing `disable_artifact` flag:
- New field `disable_video: bool = False` on `WandBConfig` (`src/opentau/configs/default.py`), with a Google-style docstring entry.
- Added to `excluded_keys` in `to_wandb_kwargs`, so it never reaches `wandb.init()` as an unknown kwarg.
- Gates the `wandb.Video` log in the eval block of `src/opentau/scripts/train.py`:
  `if cfg.wandb.enable and not cfg.wandb.disable_video and cfg.eval.max_episodes_rendered > 0`.

Label: 🗃️ Feature

## How it was tested
- `pre-commit run --files <changed files>` — clean (ruff lint + format, pyupgrade, etc.).
- `pytest -m "not gpu" tests/configs/test_refs.py tests/configs/test_train.py tests/configs/test_utils_policies.py` — 54 passed. This includes `test_train_config_round_trip_matches_baseline`, which re-serializes the `wandb` block, so the new field round-trips cleanly.
- Direct check: `WandBConfig().disable_video` is `False` by default, and `to_wandb_kwargs()` does not contain `disable_video` (so it is never passed to `wandb.init()`).
- Not a policy change — no model / training-loop / optimizer / sampler math is touched, only the wandb video-logging gate in the eval block — so the GPU pytests and regression tests are not applicable here.

```bash
pytest -sx tests/configs/test_refs.py::test_train_config_round_trip_matches_baseline
```
```python
from opentau.configs.default import WandBConfig

# Default: videos are still logged.
assert WandBConfig().disable_video is False

# Opt out, and confirm the flag never reaches wandb.init().
cfg = WandBConfig(enable=True, notes="x", disable_video=True)
assert "disable_video" not in cfg.to_wandb_kwargs()
```
To skip eval video logging on a real run, set the flag on the CLI:
```bash
opentau-train --config_path=configs/examples/pi05_training_config.json --wandb.enable=true --wandb.disable_video=true
```

## Checklist

### DIFF
--- a/src/opentau/configs/default.py
+++ b/src/opentau/configs/default.py
@@ -433,6 +433,8 @@ class WandBConfig:
             `run_id` is provided. Defaults to True.
         disable_artifact: Set to True to disable saving an artifact despite
             `training.save_checkpoint=True`. Defaults to False.
+        disable_video: Set to True to skip logging eval grid-summary videos to
+            wandb. Defaults to False (videos are logged).
     """
 
     enable: bool = False  # Enable Weights & Biases logging.
@@ -450,6 +452,8 @@ class WandBConfig:
     allow_resume: bool | None = True  # If True, resume the run from the last checkpoint.
     # Set to true to disable saving an artifact despite training.save_checkpoint=True
     disable_artifact: bool = False
+    # Set to true to skip logging eval grid-summary videos to wandb.
+    disable_video: bool = False
 
     def __post_init__(self):
         """Prompt user for wandb notes if enabled and notes are not provided."""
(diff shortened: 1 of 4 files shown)


shuheng-liu:
Addressed the review nit in ba5cab6: added `TestWandBConfig` to `tests/configs/test_default.py` with two asserts — `disable_video` defaults to `False`, and `to_wandb_kwargs()` excludes it so it never reaches `wandb.init()`.

The two `note` items are confirmations of correctness, so no change needed there.

PR #409, opened 2026-06-05
fix(train,pi07): stabilize multi-rank in-training sim eval (OOM + desync)

shuheng-liu:
## What this does

Multi-rank **in-training simulation eval** (e.g. RoboCasa under DeepSpeed ZeRO-2 with `MUJOCO_GL=egl`) crashed at *every* eval boundary. Two independent root causes, both fixed here (🐛 Bug):

**1. CUDA OOM right after eval.** Eval envs were built once and reused, so the sim's `AsyncVectorEnv` EGL render contexts (~GiB per env — *non-PyTorch* device memory the caching allocator can't reclaim) persisted into training and OOM'd the first post-eval backward. Eval envs are now built **per eval** and closed in a `finally` (`_eval_with_fresh_envs` in `scripts/train.py`), so that memory is released before training resumes. `close_envs` runs even if the eval raises.

**2. Cross-rank collective deadlock (and pathologically slow evals).** Each rank rolls out independent, variable-length episodes, but the policy forward fired a per-step cross-rank all-reduce — `_global_or_branch_decisions` (the branch-decision OR-reduce in `pi07` / `pi07_paligemma`). When the fastest rank finished its rollout and reached the post-eval `gather` while slower ranks were still mid-rollout, the mismatched collectives (ALLREDUCE vs ALLGATHER at the same sequence number) deadlocked at NCCL. The all-reduce is now gated by an explicit `sync_across_ranks` flag threaded from the **independent-rollout entry point** (`sample_actions`), defaulting `True`: only the sim-eval rollout passes `False`, while **training *and* lockstep in-training validation keep the collective** (so the FSDP/ZeRO-3 param-all-gather desync protection of CLAUDE.md rule 5 is preserved). The training forward stays byte-identical.

**3. Fail-fast guard.** In-training sim eval under parameter sharding (DeepSpeed ZeRO-3 / FSDP) now raises a clear error at startup instead of hanging — the eval forward's param all-gathers would desync across the independent rollouts regardless of (2). Use ZeRO-1/2, or run eval out-of-process.

## How it was tested

- `pre-commit` clean; CPU suite (`pytest -m "not gpu"`): `tests/policies/` (693 passed) + `tests/scripts/` (198 passed). New coverage: the `sync_across_ranks` guard truth table + default-`True` tripwires (both `pi07` and `pi07_paligemma`), and `_eval_with_fresh_envs` env-lifecycle pairing (build↔close each eval, incl. `close_envs` still called when the eval raises).
- **Targeted GPU pytests** (`pytest -m "gpu"` for the two changed policy files) on A100: **5 passed, 4 skipped** (the skips are pre-existing memory-gated tests).
- **Determinism** (CLAUDE.md rule 3): two same-seed 8×A100 ZeRO-2 runs produced **bit-identical** per-step training losses (total / MSE / CE) across all pre-eval steps — the fix leaves the training forward byte-identical.
- **End-to-end 8×A100 multi-eval run** (cold start, `eval_freq=10`, `steps=50` → **5 consecutive** in-training RoboCasa CloseFridge evals, ZeRO-2, `MUJOCO_GL=egl`, 8 envs/GPU = the OOM-trigger config): **all 5 evals completed, training resumed past every eval to step 50 — zero OOM, zero NCCL-watchdog deadlock, zero requeue.** `eval_per_gpu_s` is now a consistent **~104–110 s** (vs. **597 s → 1481 s** plus a crash pre-fix). This directly reproduces, then resolves, the conditions that crashed the prior run at every eval boundary.
- The nightly `gpu_test.yml` / `regression_test.yml` provide the full GPU + regression coverage.

```bash
pytest -sx tests/policies/test_pi07_low_level.py::TestGlobalOrBranchDecisionsSyncGuard
pytest -sx tests/scripts/test_train.py::TestEvalWithFreshEnvs
pytest -m "not gpu" tests/policies/ tests/scripts/
```

Or exercise the path directly: run training with `eval_freq > 0` on a RoboCasa config under DeepSpeed ZeRO-2 (`MUJOCO_GL=egl`); consecutive in-training evals no longer OOM the next step or deadlock at NCCL.

## Checklist

### DIFF
--- a/src/opentau/policies/pi07/low_level/modeling_pi07_low_level.py
+++ b/src/opentau/policies/pi07/low_level/modeling_pi07_low_level.py
@@ -1472,6 +1472,7 @@ class PI07LowLevelFlowMatching(nn.Module):
         subgoal_images: list[Tensor] | None = None,
         subgoal_img_masks: list[Tensor] | None = None,
         group_index: Tensor | None = None,
+        sync_across_ranks: bool = True,
     ) -> tuple[Tensor, Tensor, Tensor]:
         """Embed all prefix modalities and build the 1-D attention pattern.
 
@@ -1586,7 +1587,7 @@ class PI07LowLevelFlowMatching(nn.Module):
             any_locals=(has_response_local, has_subgoal_local, has_metadata_local),
             field_names=("response", "subgoal", "metadata"),
             device=device,
-            sync_across_ranks=self.training,
+            sync_across_ranks=sync_across_ranks,
         )
         has_any_optional = bool(has_response or has_subgoal or has_metadata)
 
@@ -2188,6 +2189,9 @@ class PI07LowLevelFlowMatching(nn.Module):
             subgoal_img_masks=subgoal_img_masks,
             obs_history_is_pad=obs_history_is_pad,
             group_index=group_index,
+            # Independent per-rank rollout (sim eval): skip the cross-rank branch
+            # all-reduce so variable-length rollouts don't desync at NCCL.
+            sync_across_ranks=False,
         )
         prefix_att_2d_masks = make_att_2d_masks(prefix_pad_masks, prefix_att_masks)
         prefix_position_ids = torch.cumsum(prefix_pad_masks, dim=1) - 1
--- a/src/opentau/policies/pi07/low_level/modeling_pi07_low_level.py
+++ b/src/opentau/policies/pi07/low_level/modeling_pi07_low_level.py
@@ -119,9 +119,10 @@ def _global_or_branch_decisions(
         device: Device to allocate the all-reduce buffer on; must match the
             distributed backend (typically the policy's CUDA device).
         sync_across_ranks: When ``True`` (training default), OR-reduce the
-            decisions across ranks. When ``False`` (eval/inference, passed from
-            ``self.training``), skip the all-reduce and return the local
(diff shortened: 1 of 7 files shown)


shuheng-liu:
Addressed the blocking finding (good catch) at `dfc7921`:

- Replaced the `self.training` gate with an explicit `sync_across_ranks` flag threaded from the **independent-rollout entry point** (`sample_actions → embed_prefix → _global_run_image_tower / _global_or_branch_decisions`), defaulting `True`. Only the sim-eval rollout passes `sync_across_ranks=False`; the in-training **validation** forward (lockstep, eval mode) now keeps the cross-rank OR-reduce, so the FSDP/ZeRO-3 param-all-gather desync protection (CLAUDE.md rule 5) is preserved.
- The fail-fast guard stays sim-eval-only (`eval_freq > 0 and cfg.env is not None`) by design — validation keeps the collective, so it doesn't need the guard.
- Reverted the now-unneeded mock `training` attrs (the explicit flag is read, not `self.training`).
- Fixed the `github-code-quality` single-import-style nit in `test_train.py`.
- Added default-`True` tripwires for `embed_prefix` / `_global_run_image_tower` so training + validation can't be silently switched off sync.

CPU suite green (`tests/policies` + `tests/scripts`, 693 passed). GPU determinism + the 8-GPU multi-eval validation are next.

PR #410, opened 2026-06-05
test(ci): quarantine live-Hub tests off the gating CPU run

shuheng-liu:
## What this does

The gating CPU test suite (`cpu_test.yml`) goes red constantly on transient Hugging Face Hub failures rather than real bugs. Across the latest 20 failed `cpu_test` runs, ~11 were **pure** Hub rate-limit / connectivity errors (`429 Too Many Requests`, `OSError: couldn't connect to huggingface.co`, `404` on gated repos), and the most recent run was **43 failed + 6 errors, 100% `huggingface.co` connectivity** — zero real failures. This reds the required `CPU Tests Status` check for reasons unrelated to the PR, and worse, buries any genuine regression under ~40 flake-reds every run.

Root cause: a handful of integration tests download from the live Hub on every run while sitting on the **gating** path. This PR quarantines them instead of deleting them:

- Adds a `network` pytest marker (`pyproject.toml`).
- Tags the live-Hub tests (~57) `@pytest.mark.network` across 7 files — module-level on the loc-token and `test_hub` files, class-level on the 3 PaliGemma embedding test classes, per-test on the `lerobot/*` dataset integration tests.
- Excludes them from the gating run: `-m "not gpu"` → `-m "not gpu and not network"` (`cpu_test.yml`).
- Adds a **non-gating** nightly `network_test.yml` that runs `pytest -m "network" -n 0` (serial, so parallel workers don't trigger 429 storms) — the tests still execute regularly, they just can't block PRs.

No assertions are removed; dataset-path logic stays covered on the gate via the existing fixture-mocked tests (which patch `snapshot_download` / `hf_hub_download`). Only test decorators, the marker registration, and CI workflows change — no policy or modeling code.

Label: `test`.

## How it was tested

- `pytest --markers` shows the registered `network` marker.
- Collection partition verified: `-m network` selects exactly 57 tests across the 7 files; `-m "not gpu and not network"` drops all 57 (whole-file-marked files → 0 on the gate; `test_datasets.py` keeps 62 local tests, the PaliGemma file keeps 25). Confirmed no live-Hub test is left on the CPU gate — the only other Hub-loading test, `test_pi06.py::test_pi06_loc_tokens_extend_vocab_and_resize_embeddings`, is already `@pytest.mark.gpu`.
- `pre-commit run --files <changed>` clean (ruff, ruff-format, zizmor, bandit, secret scan, license header).
- Ran the kept-on-gate local tests in the four mixed-content files I edited: **167 passed**, 50 network/slow deselected, fully offline.

```bash
# the quarantined set (now runs nightly, off the gate):
pytest -m "network" --collect-only -q
```
```bash
# the new gating expression — no live-Hub tests, stays green offline:
pytest -m "not gpu and not network" --collect-only -q
```

## Checklist

### DIFF
--- a/.github/workflows/cpu_test.yml
+++ b/.github/workflows/cpu_test.yml
@@ -167,7 +167,7 @@ jobs:
           python3 -c "import sys; print(sys.path)"
           python3 -c "import libero.libero" && echo "LIBERO config set successfully."
           echo "Running cpu based pytest and generating coverage report..."
-          pytest -m "not gpu" -n auto -v --cov=lerobot/ --cov-report=xml:cpu_test/cpu_test.xml --ignore=tests/planner/test_planner.py --ignore tests/utils/test_libero_utils.py --deselect=tests/envs/test_factory.py::TestMakeEnv::test_make_env_async_vector_env --deselect=tests/envs/test_factory.py::TestMakeEnv::test_make_env_sync_vector_env tests/
+          pytest -m "not gpu and not network" -n auto -v --cov=lerobot/ --cov-report=xml:cpu_test/cpu_test.xml --ignore=tests/planner/test_planner.py --ignore tests/utils/test_libero_utils.py --deselect=tests/envs/test_factory.py::TestMakeEnv::test_make_env_async_vector_env --deselect=tests/envs/test_factory.py::TestMakeEnv::test_make_env_sync_vector_env tests/
           echo "Pytest execution and coverage report generation completed."
 
       - name: Upload coverage reports
--- a/.github/workflows/network_test.yml
+++ b/.github/workflows/network_test.yml
@@ -42,7 +42,9 @@ jobs:
   network-tests:
     name: Network Tests
     runs-on: ubuntu-latest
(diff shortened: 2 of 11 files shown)


shuheng-liu:
Addressed the review in `36a17f7`:

- **suggestion — `test_from_pretrained_path_does_not_exits` reaching the live Hub:** good catch. My completeness pass missed it because the repo id is a bare `"bert123"`, not a `google/`-`lerobot/` pattern. Rather than quarantine a fast, useful assertion, I mocked `hf_hub_download` to raise `RepositoryNotFoundError`, so `from_pretrained` still translates it into `FileNotFoundError`. The test now runs **offline and deterministically while staying on the gate** (verified under `HF_HUB_OFFLINE=1`), mirroring the sibling `test_from_pretrained_model_exits`. This closes the exact gap you flagged — a raw `OSError` connectivity failure can no longer escape the `pytest.raises`.
- **nit — `network_test.yml` timeout:** bumped `timeout-minutes` 30 → 45 for the serial run.

PR #413, opened 2026-06-06
feat(train): add running-best checkpoint retention

shuheng-liu:
## What this does
Adds an opt-in "running best" checkpoint retention mode (🗃️ Feature).

When enabled, training keeps the checkpoint(s) that achieved the best metric so far, in addition to the regular `save_freq` checkpoints:

- The feature is controlled by a single knob, `running_best_count`: `0` disables it (default), `>= 1` keeps that many running bests — consistent with the repo's existing `eval_freq`/`val_freq` "0 == off" idiom.
- After an eval/validation step yields a new best, the current weights are retained as a "running best". A best landing on a regular `save_freq` step reuses that directory (no duplicate save); an off-cadence best is saved on demand.
- The driving metric is chosen by `running_best_metric` (`"auto"` | `"eval_success"` | `"val_loss"`). `"auto"` (default) uses the eval success rate when sim eval is configured (`env` + `eval_freq > 0`), otherwise validation loss. Success rate (0–100, higher-better) and validation loss (lower-better) are non-comparable, so a single metric drives the pool per run.
- When the pool overflows, the oldest is evicted — and its directory is deleted only if it was **not** also a regular `save_freq` checkpoint (those are left to the normal retention logic).
- Running bests are protected from `last_checkpoint_only` pruning, and the high-water mark + retained pool are persisted to `checkpoints/running_best.json` so they survive a resume (self-healing if a referenced dir was later pruned; the high-water mark is discarded if a resume flips the metric direction).

The new-best decision is made on rank 0 (the only rank holding the metric) and broadcast so every rank agrees before the collective `accelerator.save_state`, avoiding a per-rank-divergent save that would hang NCCL. The feature is inert when off (the default): no change to training numerics or RNG consumption.

New config fields on `TrainPipelineConfig`:
- `running_best_count: int = 0` (0 disables; `>= 1` keeps N)
- `running_best_metric: str = "auto"`

## How it was tested
- Added config-validation tests in `tests/configs/test_train.py`: count bounds (negative raises, 0 is a valid disabled state), metric literal, missing-metric-source errors when enabled, `auto` resolution to `eval_success`/`val_loss`, `save_checkpoint=False` validity, serialization round-trip, and backward-compat defaults for configs that predate the feature.
- Added bookkeeping tests in `tests/utils/test_train_utils.py`: `running_best_is_improvement` (strict / NaN / both directions); `RunningBestTracker` eviction, regular-checkpoint-not-deleted, top-N retention, and `protected_dirs`; state round-trip + missing-file + self-heal against pruned dirs + direction-mismatch reset; and `prune_old_checkpoints` `protected_paths` skipping with `None` backward-compatibility.
- CPU suite green for the affected areas (`tests/configs`, `tests/utils`, `tests/scripts`) under `pytest -m "not gpu and not network"`; pre-commit (ruff lint + format, etc.) clean.

The feature is inert when `running_best_count == 0`, so existing-run training numerics are unchanged by construction. The multi-rank distributed-save path (the collective `accelerator.save_state` + cross-rank broadcast) cannot be exercised on CPU and has no automated test; it should be validated with a short multi-GPU smoke (eval/validation enabled) before relying on it in a real run.

```bash
pytest -sx tests/utils/test_train_utils.py -k "running_best or protected"
pytest -sx tests/configs/test_train.py -k running_best
```
To enable it on a real run, set the count (and let the metric follow whichever source is configured), e.g. with a validation split:
```bash
opentau-train --config_path=<config.json> --running_best_count=2 --val_freq=1000
```

## Checklist


shuheng-liu:
Addressed the review in `171336a`:

- **suggestion (resume direction mismatch):** fixed — `load_running_best_state` now discards the persisted high-water mark (keeping the pool) when a resume flips the metric optimization direction (`higher_is_better` mismatch), with a warning. Added `test_running_best_state_direction_mismatch_resets_best`.
- **nit (evicted regular-coincident dir lingers until the next regular-save prune):** intentional and left as-is — the spec is to delete an evicted running best *only if it was not also a regular `save_freq` checkpoint*, leaving regular checkpoints to the normal retention logic. It's disk-only with no correctness impact.

Also, per maintainer feedback, collapsed the redundant `save_running_best` boolean into `running_best_count` (`0` disables, `>= 1` keeps N), matching the existing `eval_freq`/`val_freq` "0 == off" idiom.

shuheng-liu:
On the self-heal nit (`load_running_best_state` keeps the high-water mark when a pooled dir is gone): leaving this as-is intentionally.

The regular prune (`prune_old_checkpoints`) is passed the live pool via `protected_paths`, so a dir that is currently a running best is never deleted by normal retention. A pooled dir only goes missing through deletion outside this code path (manual `rm`, or a partial save), so the "best preserved, empty pool" state isn't reachable through normal operation. Even if it occurs, `best` is a genuine high-water mark — declining to re-save a *worse* checkpoint under it is the conservative behavior, and it's disk/UX-only with no correctness impact. Resetting it on a flipped metric direction (the previous fix) is different: there the persisted value is meaningless, so it's discarded.

PR #418, opened 2026-06-08
fix(datasets): val-split RAM + v3.0 video frame-index clamp + load retry

shuheng-liu:
## What this does

Three robustness fixes that surfaced co-training a large mixture of **v3.0**
(file-consolidated) robot datasets, each independent:

**1. Bound host RAM on the train/val split (`_subset_meta`).**
`make_dataset` did `copy.deepcopy(dataset.meta)` for *each* of the train and val
subsets when `val_freq > 0`. For a dataset with **per-episode** stats this clones
the large read-only `episodes` / `episodes_stats` dicts; when the same big repo
is listed under many mixture configs that is tens of GB of duplicated metadata
**per rank**, and — worse — copy-on-write surface that explodes host RAM once the
DataLoader forks its workers (it OOM-killed an 8-rank run). The split is by frame
index, so those dicts are identical between halves and read-only during training;
share them by reference and deep-copy only the small aggregated `stats`.

**2. Clamp the v3.0 video frame index (torchcodec).**
`decode_video_frames_torchcodec` passed `round(ts * average_fps)` straight to
`get_frames_at` with no bounds check. A query timestamp can land at/just past a
consolidated video's final frame (a minority of videos are encoded with
marginally fewer frames than their metadata implies, or `average_fps` rounds up
near the end of a long video). torchcodec raises a hard `IndexError` there —
unlike the pyav path, which returns the nearest decoded frame. Clamp the indices
to `[0, num_frames-1]` so the two backends behave the same; the existing
tolerance check still rejects a frame genuinely too far from its query.

**3. Enable the load-retry fallback (`DEFAULT_RAND_RETRY_ATTEMPTS`).**
`retry_random_on_failure` read `_total_rand_attempts` via `getattr(..., 0)`, and
the attribute was never set anywhere — so the documented retry-on-load-failure
feature was inert (exactly one attempt), and a single unrecoverable item aborted
the whole run. Add a default of 8 so a transient per-item failure (e.g. a video
whose decode raises despite the clamp) is skipped and resampled instead of fatal.

⚡️ Performance / 🐛 Bug.

## How it was tested

- **New unit tests** (CPU):
  - `test_subset_meta_shares_episode_metadata_but_isolates_stats` — train/val
    subsets share `episodes`/`episodes_stats` by reference, each gets an
    independent `stats`.
  - `test_torchcodec_clamps_out_of_range_frame_index` — a past-the-end query is
    clamped to `num_frames-1` (mocked decoder) instead of raising.
  - `test_retry_random_on_failure_uses_default_attempts` /
    `..._recovers_from_transient_failure` — the retry default drives N retries and
    recovers from a transient failure.
  - Existing val-split (`test_make_dataset_*_val_split_ratio_*`) and v3.0/mixture
    suites still pass.
- **Validated against real data**: the clamp was exercised in-environment on a
  real consolidated video (a one-frame-past query returns the last frame; a
  far-past query raises a tolerance error that the retry then skips).
- **End-to-end**: an 8-GPU run of the mixture trained cleanly through the
  step-1000 checkpoint. Host RAM stayed bounded (~0.2-0.3 TB; the prior baseline
  was ~0.9 TB and climbed to an OOM kill), the load-retry recovered transient
  bad-video samples without aborting, total loss fell from ~13.7 (step 50) to
  ~2.0 (step 1000) with MSE ~10.5 -> ~1.2, and per-task sim-eval success rose
  from 0 to non-zero by step 750 — i.e. the fixes produce sane, converging
  training rather than masking a problem.

```bash
pytest -sx tests/datasets/test_video_utils.py::test_torchcodec_clamps_out_of_range_frame_index
pytest -sx tests/datasets/test_datasets.py -k "subset_meta_shares or retry_random_on_failure or val_split_ratio"
```

## Checklist

### DIFF
--- a/src/opentau/datasets/factory.py
+++ b/src/opentau/datasets/factory.py
@@ -169,6 +169,33 @@ def resolve_delta_timestamps(
     return dt_mean, {}, {}, {}
 
 
+def _subset_meta(meta: LeRobotDatasetMetadata) -> LeRobotDatasetMetadata:
+    """Per-subset metadata copy that shares read-only per-episode structures.
+
+    Returns a shallow copy of ``meta`` that shares ``episodes`` /
+    ``episodes_stats`` (and every other attribute) by reference, deep-copying
+    only the small aggregated ``stats`` dict so the train/val halves can hold
+    independent normalization stats without cross-contamination.
+
+    ``random_split`` divides a dataset by frame index, so the per-episode
+    ``episodes`` / ``episodes_stats`` are identical across the two subsets and
+    are read-only on the training path (only the dataset-*creation*
+    ``save_episode`` / stats-conversion paths write them). A blanket
+    ``deepcopy(meta)`` cloned them per subset: for a large per-episode-stats
+    dataset listed under many mixture configs that is tens of GB of needless
+    copies per rank, and — worse — copy-on-write surface that blows up host RAM
+    once forked across dataloader workers (every worker touches the clones).
+    Sharing them by reference removes both costs. Mirrors the share-by-reference
+    contract of :meth:`BaseDataset.shallow_copy_with_dropout`.
+
+    Args:
+        meta: The source dataset metadata to copy for a train/val subset.
+    """
+    subset_meta = copy.copy(meta)
+    subset_meta.stats = copy.deepcopy(meta.stats)
+    return subset_meta
+
+
 def make_dataset(
     cfg: DatasetConfig,
     train_cfg: TrainPipelineConfig,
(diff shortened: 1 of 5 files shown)


shuheng-liu:
github-code-quality flagged `__getitem__` in the two retry-test doubles (`_AlwaysFails`, `_FailsOnce`) for raising a non-standard exception. This is intentional: they are minimal fakes that exist only to exercise `retry_random_on_failure`, which catches `Exception` broadly — a real per-item load failure here is typically a `RuntimeError` from video decode, not `IndexError`/`KeyError`. The "standard exception in a special method" guidance targets production iteration/indexing semantics, not test doubles, so resolving these as non-actionable rather than changing the test exception types.

PR #420, opened 2026-06-09
feat(eval): shrink wandb eval-video uploads via CRF + frame subsampling

shuheng-liu:
## What this does

In-training eval renders one `grid_summary.mp4` montage per task and uploads it to wandb (`Eval Videos/{task_name}`) — that grid is the **only** video uploaded. It was written with imageio's default encoder (which exposes no size knob), every frame, while the per-episode source clips (`eval_episode_*.mp4`) were never deleted. This steadily consumes the wandb video-storage quota.

This adds configurable, storage-lean H.264 encoding for the uploaded grid, via four new `EvalConfig` fields (all individually overridable on the CLI / config):

| Field | Default | Effect |
|---|---|---|
| `video_crf` | `30` | H.264 constant-rate-factor (higher = smaller / lower quality, 0–51) |
| `video_preset` | `"veryfast"` | x264 encode-speed vs ratio preset; does not change the quality target |
| `video_frame_stride` | `2` | write only every k-th composed grid frame (~linear size cut; plays k× faster) |
| `keep_per_episode_videos` | `False` | delete the per-episode clips once the grid is built (they are never uploaded) |

`create_grid_summary_video` now writes `libx264` + `yuv420p` with `quality=None` (so `-crf` governs the rate instead of imageio's default `-qscale`), keeping the file browser/wandb HTML5-playable. The per-episode clips and `write_video` are **deliberately left uncompressed** — they aren't uploaded, and keeping them clean avoids double-compressing the grid that is re-encoded from them. Knobs are read with `getattr(..., default)` so a resume from an older persisted config still works.

Measured magnitude (apples-to-apples, identical composed frames): imageio's default is already a reasonable VBR, so CRF 30 alone is ~1.2–1.9× and CRF 30 + stride 2 is ~1.5–3.9× smaller depending on content (~1.5–2.5× for typical eval renders). Resolution downscale — the big quadratic lever — was intentionally left out of scope.

## How it was tested

- `pre-commit run --all-files` clean; CPU suite green.
- Added CPU tests in `tests/scripts/test_eval.py`:
  - `test_create_grid_summary_video_crf_shrinks_file` — higher CRF yields a smaller file.
  - `test_create_grid_summary_video_frame_stride_subsamples` — stride k writes ~1/k of the body frames.
  - `test_create_grid_summary_video_default_params_keep_shape` — codec/pixel-format change preserves grid dimensions.
  - `test_cleanup_episode_clips_deletes_unless_kept` — clips removed when `keep=False`, retained when `True`, missing paths tolerated.
- Existing grid/io tests stay green (function-level defaults preserve prior behavior).
- Not a policy change (no `modeling_*` / `train.py` / `optim` / `sampler` touched), so no GPU/regression run.

```bash
```

Quantify the size win on synthetic frames:

```bash
python - <<'PY'
import os, tempfile, imageio.v2 as imageio, numpy as np
from opentau.scripts.eval import create_grid_summary_video
d = tempfile.mkdtemp(); rng = np.random.default_rng(0)
clips = []
for k in range(4):
    p = os.path.join(d, f"c{k}.mp4")
    imageio.mimsave(p, [rng.integers(0, 256, (256, 256, 3), np.uint8) for _ in range(60)], fps=20)
    clips.append(p)
for crf, stride, name in [(0, 1, "crf0"), (30, 1, "crf30"), (30, 2, "crf30_s2")]:
    out = os.path.join(d, name + ".mp4")
    create_grid_summary_video(clips, [True, False, True, False], out, fps=20,
                              highlight_duration=0.0, crf=crf, frame_stride=stride)
    print(f"{name:10s} {os.path.getsize(out) / 1024:8.1f} KiB")
PY
```

## Checklist

### DIFF
--- a/src/opentau/configs/default.py
+++ b/src/opentau/configs/default.py
@@ -513,11 +513,23 @@ class EvalConfig:
             Defaults to 16.
         grid_size: Grid dimensions for video summary (rows, cols). If None, will
             be auto-calculated as a square grid. Defaults to None.
+        video_crf: H.264 constant-rate-factor for the uploaded grid-summary video
+            (higher = smaller file / lower quality, 0-51). Defaults to 30.
+        video_preset: x264 encode preset (ultrafast..veryslow); an encode-speed
+            vs compression-ratio knob that does not change the quality target.
+            Defaults to "veryfast".
+        video_frame_stride: Keep only every k-th frame of the grid-summary video
+            (k>1 shrinks the upload ~linearly and speeds playback up k x).
+            Defaults to 2.
+        keep_per_episode_videos: If False, delete the per-episode
+            eval_episode_*.mp4 clips after the grid summary is built (they are
+            never uploaded to wandb). Defaults to False.
         recording_root: Root directory for saving evaluation recordings.
             Defaults to None.
 
     Raises:
-        ValueError: If `batch_size` is greater than `n_episodes`.
+        ValueError: If `batch_size` is greater than `n_episodes`, or if any of
+            `video_crf`, `video_preset`, `video_frame_stride` is out of range.
     """
 
     n_episodes: int = 16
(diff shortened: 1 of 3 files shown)


shuheng-liu:
Thanks for the review — addressed in ac5fccc:

1. **`info["video_paths"]` dangling refs** (eval.py:494) — fixed. `video_paths` is now filtered to clips that still exist before going into `info`, so a `keep_per_episode_videos=False` run no longer serializes deleted paths into `eval_info.json`.
2. **`__post_init__` validation had no test** — added `test_eval_config_rejects_out_of_range_video_params`, covering the `video_crf` range, `video_frame_stride >= 1`, and unknown-`video_preset` branches.
3. **`video_frame_stride=2` default → 2× faster playback** — intentional, and "flagged for awareness" is the right call. Frame subsampling was explicitly chosen as a default-on storage lever; the field doc notes the k× speedup, and anyone wanting real-time playback sets `eval.video_frame_stride=1`. Leaving as-is.

shuheng-liu:
Re-review on ac5fccc confirms both actionable items resolved. On the remaining awareness nit (`keep_per_episode_videos=False` also affects standalone `opentau-eval`, not just in-training eval): intentional. The default is storage-lean across the board; the clips are local-only in both paths, and `eval.keep_per_episode_videos=true` restores them for per-episode inspection. Leaving as-is.

PR #425, opened 2026-06-13
fix(policies): respect skip_normalization_weights on checkpoint load

shuheng-liu:
## What this does

Fixes a bug where `skip_normalization_weights` was silently ignored when fine-tuning a policy from a checkpoint via `make_policy(..., ds_meta=...)`.

`make_policy` calls `policy._inject_stats(per_norm_key_stats, ...)` **after** `from_pretrained`, with the comment that it repopulates buffers that were saved as `+inf` (the `save_normalization_stats=False` round-trip). But it called it **unconditionally** whenever the run trains from a dataset mixture, so it always overwrote the loaded checkpoint's `normalize_*` / `unnormalize_*` buffers with the current mixture's recomputed stats. The effect:

- `skip_normalization_weights=False` (the default, meaning "keep the checkpoint's normalization") was a **no-op for any mixture fine-tune** — the saved stats were always clobbered by recomputed ones.
- When the fine-tune mixture differs from the checkpoint's training mixture (e.g. a dataset is dropped), the recomputed per-group action std can shift substantially. In a real fine-tune this pushed some sources' normalized actions past **1000 sigma** (vs ~50 during pretraining), which spiked the regression MSE (~0.2 to ~40 at the first step) and destabilized gradients — even though the model and its checkpointed normalization were both fine.

The fix gates the post-load injection on a new `PreTrainedPolicy._norm_buffers_have_inf()`: only repopulate when buffers are still the `+inf` sentinel. Otherwise keep the buffers that were deliberately loaded (`skip_normalization_weights=False`) or freshly built from the current mixture (`skip_normalization_weights=True`, which already strips the saved buffers and initializes from `per_dataset_stats`). This makes the flag behave as documented in all three cases.

🐛 Bug

## How it was tested

- Added `TestNormBufferInfGate` in `tests/policies/test_pretrained_skip_normalization.py` (3 tests):
  - `_norm_buffers_have_inf()` returns `False` for valid buffers, `True` once a buffer is set to the sentinel.
  - A clean `skip_normalization_weights=False` load keeps the checkpoint's stats (the gate skips `_inject_stats`, so the current-mixture stats do **not** overwrite the loaded ones).
  - `+inf` buffers (the `save_normalization_stats=False` case) are still repopulated by the gated `_inject_stats`.
- Ran the full CPU policies suite: `pytest tests/policies -m "not gpu and not network"` -> **467 passed, 2 skipped**.
- End-to-end: a fine-tune from a checkpoint onto a reduced mixture that previously started at MSE ~40 with normalized-action outliers ~1000 now starts at MSE ~0.24 with outliers ~45 (the pretraining-tolerable range) and stable gradients, confirming the loaded normalization is preserved.

This change is in policy construction / checkpoint-loading code (`make_policy`, `PreTrainedPolicy`), not in model math, the training loop, the optimizer, or the sampler, and is fully covered by CPU tests — so the policy / GPU-pytest checklist boxes are left unchecked.

```bash
pytest -sx tests/policies/test_pretrained_skip_normalization.py::TestNormBufferInfGate
```

```bash
pytest tests/policies -m "not gpu and not network"
```

## Checklist

### DIFF
--- a/src/opentau/policies/factory.py
+++ b/src/opentau/policies/factory.py
@@ -298,7 +298,20 @@ def make_policy(
     # have them, otherwise raise a clear error rather than letting the first
     # forward fail mid-step.
     if cfg.pretrained_path and per_norm_key_stats is not None:
-        policy._inject_stats(per_norm_key_stats, dataset_names=norm_keys)
+        # Repopulate normalization buffers ONLY when the checkpoint carried no
+        # stats (buffers are still the +inf sentinel — the
+        # `save_normalization_stats=False` round-trip). When the checkpoint's
+        # stats loaded cleanly (`skip_normalization_weights=False`) or were
+        # freshly built from the current mixture
+        # (`skip_normalization_weights=True`, which strips the saved buffers and
+        # keeps the per_dataset_stats-initialised ones), keep them as-is.
+        # Unconditionally injecting here clobbered a deliberately-loaded
+        # checkpoint normalization, which made `skip_normalization_weights=False`
+        # a silent no-op for mixture fine-tunes.
+        if policy._norm_buffers_have_inf():
+            policy._inject_stats(per_norm_key_stats, dataset_names=norm_keys)
+        else:
+            policy._check_norm_stats_loaded()
     elif cfg.pretrained_path:
         policy._check_norm_stats_loaded()
 
--- a/src/opentau/policies/pretrained.py
+++ b/src/opentau/policies/pretrained.py
@@ -24,6 +24,7 @@ and safetensors for efficient serialization.
 import abc
 import logging
 import os
+from collections.abc import Iterator
 from pathlib import Path
 from typing import Type, TypeVar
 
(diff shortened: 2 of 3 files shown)


shuheng-liu:
Thanks for the review. Addressed the **nit** in a285070: extracted a shared `_iter_norm_buffer_params` staticmethod so `_norm_buffers_have_inf`, `_check_norm_stats_loaded`, and `_assert_normalize_buffers_initialized` no longer duplicate the buffer walk (each keeps its own `isinf` predicate; `_assert` still applies `_materialize` for the FSDP2 DTensor case).

On the **suggestion** (factory gate not driven end-to-end via `make_policy`): leaving as-is by design. This test module deliberately exercises the load path without a multi-GB backbone (see the module docstring near the `TestSkipNormalizationMultiHeadRoundTrip` section) and replicates the normalization-relevant steps inline — `TestNormBufferInfGate` follows that same convention. A true `make_policy` end-to-end test would need a real policy class + checkpoint round-trip, which is out of scope for this fix. The gate itself is one `if policy._norm_buffers_have_inf():`, and both branches of the helper it calls are covered.

PR #428, opened 2026-06-17
feat(train): optional torch.compile of flow-matching module for pi05/pi07

WilliamYue37:
## What this does

Adds a **`torch.compile` path for training** (⚡️ Performance), wired for **π₀.₅ and π₀.₇** and **on by default** for them.

Two new fields on `PreTrainedConfig`:

- `use_torch_compile: bool` — base default `False`; **overridden to `True` on `PI05Config` and `PI07LowLevelConfig`** (the two policies wired for it), so pi05/pi07 training compiles by default. Other policies stay eager (no behaviour change for them). Set `--policy.use_torch_compile=false` to opt out.
- `torch_compile_mode: str = "default"`

When enabled, `PreTrainedPolicy.maybe_compile_for_training()` compiles the inner **flow-matching module** (`self.model` — `PI05FlowMatching` / `PI07LowLevelFlowMatching`), not the policy wrapper. The wrapper's `forward` does tokenization / normalization / `.cpu().tolist()` / numpy / string work that would only graph-break, while `self.model.forward` is the heavy two-pass transformer (backbone prefix + action-expert) that actually benefits.

Key implementation choices:

- **In-place `nn.Module.compile()`, not `model = torch.compile(model)`.** The in-place form swaps only the submodule's `__call__` dispatch — it does **not** wrap the module in an `OptimizedModule` and does **not** prefix `state_dict` keys with `_orig_mod.`. So `policy.parameters()` and the on-disk checkpoint layout are unchanged: the optimizer built afterwards sees the same params, and resume / `from_pretrained` stay compatible with existing (uncompiled) checkpoints.
- Because `nn.Module.compile()` only intercepts `__call__`, the two policies' training forward was switched from `self.model.forward(...)` to `self.model(...)` (a direct `.forward()` call bypasses the compiled dispatch and would silently no-op the compile). No hooks are registered on `self.model`, so this is numerically identical when compile is off (verified below).
- **pi07's `_global_or_branch_decisions` is decorated `@torch.compiler.disable`** so its cross-rank `all_reduce` graph-breaks cleanly and stays in eager — keeping the OR-reduce rank-consistent instead of being traced into the compiled region.
- **`train.py` backend handling**: under DeepSpeed ZeRO-3 / FSDP (both re-shard parameters during forward, invalidating the traced guards), it **warns and falls back to eager** rather than raising — since compile is now a default, an incompatible backend should still train (uncompiled), not crash. DDP / single-process / DeepSpeed ZeRO-1/2 compile normally. The compile is applied after the bf16 cast and before `accelerator.prepare`, while `policy` is still the raw module.
- `profile_step.py` honors the flag and adds a `TORCH_COMPILE=true/false` env override (mirrors `GRAD_CHECKPOINT`) for A/B benchmarking.
- Because the flag lives on the shared `PreTrainedConfig` but only pi05 / pi07-low-level dispatch `self.model` via `__call__`, a `supports_torch_compile` class marker gates the compile: it's `True` only on the two wired (and validated) policies. Enabling the flag on any other policy is a **loud warn-and-skip**, not a silent compile-but-never-dispatch.

## How it was tested

**CPU unit tests** — added `tests/policies/test_torch_compile.py` (26 cases): compile defaults ON for the wired pi05 / pi07-low-level configs and OFF elsewhere; both fields keyword-settable; `maybe_compile_for_training` is a no-op when off, installs the compiled dispatch when on, **leaves `state_dict` keys unchanged** (no `_orig_mod.`), and warn-skips on an unwired policy; the `supports_torch_compile` marker is opt-in (base/pi0 `False`, pi05/pi07 `True`). Full CPU gate (`pytest -m "not gpu and not network"`) green locally.

**Real training, 8×A100-80GB, DeepSpeed ZeRO-2 + accelerate** (`profile_step.py`, π₀.₅ `configs/examples/pi05_training_config.json`, eager attention, 20 warmup + 30 measured steps, median — **0 recompiles, 0 errors**). Two regimes:

*Expert-only (`train_expert_only=true`, batch 8/rank):* compile speeds up the whole step because backward is light —

| | forward | total step | throughput |
|---|---|---|---|
| eager | 389 ms | 726 ms | 74.8 samples/s |
| **compiled** | **267 ms (1.46×)** | **574 ms (1.26×)** | **91.7 samples/s (+23%)** |

*Full fine-tune (`train_expert_only=false`, `freeze_vision_encoder=false`), batched to the 80 GB limit:* the step is dominated by ZeRO-2 gradient reduce-scatter over all 3B params (backward ≈ 93% of the step), so at a **matched** batch compile's forward win barely moves the wall-clock — but it **cuts peak memory ~21%**, which lets you fit a bigger batch:

At matched batch 12/rank (= global 96, the eager max):

| | forward | bwd | total step | peak mem |
|---|---|---|---|---|
| eager | 447 ms | 6639 ms | 7141 ms | 74.3 GB |
| **compiled** | **341 ms (1.31×)** | 6714 ms | 7074 ms (≈flat) | **58.7 GB (−21%)** |

Maxed out per backend (largest batch that fits 80 GB, full throughput at that batch):

| | max batch/rank | global batch | throughput |
|---|---|---|---|
| eager | 12 | 96 | 13.9 samples/s |
| **compiled** | **16** | **128** | **18.8 samples/s (1.35×)** |

Takeaways: compile reliably makes the **forward 1.3–1.5× faster**, is checkpoint- and determinism-safe, and **frees ~21% peak memory**. End-to-end wall-clock at a *fixed* batch depends on the forward/backward balance — large for forward-heavy regimes (expert-only, inference, grad-accumulation > 1), ≈flat for comms-bound full-finetune-under-ZeRO-2-at-GA=1. But when you *max out the batch* (what you'd actually do), the freed memory lets full-finetune go from batch 12 → 16, lifting throughput **1.35×** (13.9 → 18.8 samples/s) on the 8×A100 node. All runs above had **0 recompiles**.

**Determinism (CLAUDE.md rule 3)** — `seed=1000`, two runs each, per-step loss series (`:.6f`) on DeepSpeed ZeRO-2 (2 ranks):
- compile **off**: **bit-identical** across runs (confirms the `.forward()`→`__call__` refactor is numerically inert).
- compile **on**: initially diverged (inductor functionalized RNG); **bit-identical after `fallback_random=True`** — e.g. both runs: `5.700914, 4.987164, 5.013531, 5.779664, 5.702332, 5.995920`. Re-confirmed bit-identical with **compile-on as the new default** on a second 8×A100 node (same loss series), so the default training path is reproducible.

**π₀.₇ low-level** (`configs/examples/pi07_low_level_libero.json`, Gemma 3 backbone, `sdpa`, `gradient_checkpointing=true`, full-finetune, 6-frame video, compile on, ZeRO-2): **runs cleanly (rc=0)** — exercises the `@torch.compiler.disable`'d `all_reduce`. Two caveats worth noting: (1) the SDPA-flash + grad-checkpointing inductor crash is resolved only by `fallback_random=True` (above); (2) π₀.₇'s variable-length text embedding (`embed_language_tokens`) triggers `torch.compile` recompiles (24 observed) unless sequence lengths are stable — so the compile *speedup* on π₀.₇ is data-dependent (bucket/pad lengths to realize it), even though correctness holds. π₀.₅ (fixed shapes) saw **0 recompiles**.

**GPU pytests** — ran the `gpu`-marked subset that directly exercises the changed forward paths on the A100 node: `tests/policies/test_pi05.py` + `tests/policies/test_pi07_low_level.py` → **8 passed, 0 failed** (3 pre-existing resource-gated skips). Notably `TestPI05Integration::test_complete_pi05_pipeline_integration` passes — that's the end-to-end pi05 forward/backward that the `self.model.forward(...)`→`self.model(...)` change runs through. The full nightly `gpu_test.yml` + `regression_test.yml` (g6.12xlarge) will run on merge.

```bash
pytest -sx tests/policies/test_torch_compile.py
```

```bash
# A/B the compiled vs eager step time under DeepSpeed ZeRO-2:
TORCH_COMPILE=true PROFILE_STEPS=30 \
  accelerate launch --config_file <zero2_ga1.yaml> \
  --config_path=configs/examples/pi05_training_config.json --val_freq=0
```

```bash
# Real training with compile enabled:
opentau-train --config_path=configs/examples/pi05_training_config.json --policy.use_torch_compile=true
```

## Checklist

### DIFF
--- a/src/opentau/configs/policies.py
+++ b/src/opentau/configs/policies.py
@@ -306,6 +306,25 @@ class PreTrainedConfig(draccus.ChoiceRegistry, HubMixin, abc.ABC):
     # `use_amp` determines whether to use Automatic Mixed Precision (AMP) for training and evaluation. With AMP,
     # automatic gradient scaling is used.
     use_amp: bool = False
+
+    # When True, training `torch.compile`s the policy's heavy compute submodule
+    # (the flow-matching `self.model` for pi05 / pi07) *in place* via
+    # `torch.nn.Module.compile`, leaving the host-side preprocessing in the
+    # policy wrapper's own forward uncompiled. The compile is applied in
+    # `PreTrainedPolicy.maybe_compile_for_training`, called from train.py after
+    # the bf16 cast and before `accelerator.prepare`. Compatible with DDP /
+    # single-process / DeepSpeed ZeRO-1/2; train.py rejects ZeRO-3 and FSDP
+    # (both re-shard parameters during forward, which invalidates the traced
+    # graphs). Defaults to False (no behaviour change; the default eager path
+    # is byte-for-byte unchanged).
+    use_torch_compile: bool = False
+    # `torch.compile` mode forwarded to `torch.nn.Module.compile(mode=...)` when
+    # `use_torch_compile` is True. One of "default", "reduce-overhead",
+    # "max-autotune", "max-autotune-no-cudagraphs". Stick with "default" under
+    # DeepSpeed — the cudagraph-based modes ("reduce-overhead", "max-autotune")
+    # conflict with DeepSpeed's activation/parameter memory reuse.
+    torch_compile_mode: str = "default"
+
     pretrained_path: str | None = None
     skip_normalization_weights: bool = False
 
--- a/src/opentau/policies/pi05/configuration_pi05.py
+++ b/src/opentau/policies/pi05/configuration_pi05.py
@@ -155,6 +155,13 @@ class PI05Config(PreTrainedConfig):
     # require. Defaults to False (no ckpt, lowest risk).
     gradient_checkpointing: bool = False
 
+    # torch.compile defaults ON for pi05: PI05Policy is wired for it
+    # (forward dispatches self.model via __call__, supports_torch_compile=True)
+    # and it gives a measured ~1.3-1.5x faster forward with 0 recompiles under
+    # DeepSpeed ZeRO-1/2 / DDP. Set False to train in eager (train.py also
+    # auto-falls back to eager under FSDP / ZeRO-3, which compile can't support).
+    use_torch_compile: bool = True
+
     # Training presets
     optimizer_lr: float = 2.5e-5
     optimizer_betas: tuple[float, float] = (0.9, 0.95)
--- a/src/opentau/policies/pi05/modeling_pi05.py
+++ b/src/opentau/policies/pi05/modeling_pi05.py
@@ -238,6 +238,8 @@ class PI05Policy(PreTrainedPolicy):
 
     config_class = PI05Config
     name = "pi05"
+    # forward() dispatches self.model via __call__, so nn.Module.compile() takes effect.
+    supports_torch_compile = True
 
     def __init__(
         self,
--- a/src/opentau/policies/pi05/modeling_pi05.py
+++ b/src/opentau/policies/pi05/modeling_pi05.py
@@ -726,7 +726,10 @@ class PI05Policy(PreTrainedPolicy):
         )  # in actions_is_pad we have False for real actions and True for padded actions
 
         state = self.prepare_state(batch) if self.config.state_type == "continuous" else None
-        losses = self.model.forward(
+        # Call via __call__ (not .forward) so an in-place torch.compile of
+        # self.model (see PreTrainedPolicy.maybe_compile_for_training) is
+        # actually dispatched; a direct .forward() bypasses the compiled call.
+        losses = self.model(
             images,
             img_masks,
             lang_tokens,
--- a/src/opentau/policies/pi07/low_level/configuration_pi07_low_level.py
+++ b/src/opentau/policies/pi07/low_level/configuration_pi07_low_level.py
@@ -228,6 +228,13 @@ class PI07LowLevelConfig(PreTrainedConfig):
     spacetime_layer_stride: int = 4
     gradient_checkpointing: bool = False
 
+    # torch.compile defaults ON for pi07 low-level: PI07LowLevelPolicy is wired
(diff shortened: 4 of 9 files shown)


WilliamYue37:
Addressed the `maybe_compile_for_training` footgun in de69cd6: added a `supports_torch_compile` class marker (default `False`, set `True` only on `PI05Policy` / `PI07LowLevelPolicy` — the two whose forward dispatches `self.model` via `__call__`). Enabling `use_torch_compile=True` on any other policy now logs a loud warn-and-skip instead of silently compiling-but-never-dispatching. Added `test_unsupported_policy_warns_and_skips` + `test_support_marker_is_opt_in` to pin it.

shuheng-liu (approved):

WilliamYue37:
Updated for the two requested changes:

**1. torch.compile is now ON by default** for pi05 / pi07-low-level (their config `use_torch_compile` defaults to `True`; other policies unchanged). `train.py` now warns-and-falls-back-to-eager under FSDP / ZeRO-3 instead of raising, so the new default never crashes an incompatible backend.

**2. Re-benchmarked the full model (`train_expert_only=false`, `freeze_vision_encoder=false`) at max batch on a fresh 8×A100 node**, DeepSpeed ZeRO-2:
- Matched batch 12/rank: forward **1.31×** (447→341 ms), total ≈flat (full-finetune is gradient-comms-bound, bwd 93%), peak mem **−21%** (74.3→58.7 GB).
- **Maxed out: eager fits batch 12 (13.9 samples/s); compile's freed memory fits batch 16 (18.8 samples/s) → 1.35× throughput.**
- 0 recompiles; determinism re-confirmed bit-identical with compile-on as the default.

PR description updated with the full tables.

PR #431, opened 2026-06-19
feat(robocasa): per-task official horizon for sim eval

shuheng-liu:
## What this does
RoboCasa simulation eval now defaults each task to its **official per-task horizon** from the RoboCasa dataset registry (e.g. `OpenCabinet`=1050, `TurnOnMicrowave`=450) instead of one global `episode_length` cap, so a multi-task eval list evaluates every task for its intended length.

- `envs/robocasa.py`: new `_official_task_horizon()` reads the per-task horizon from `robocasa.utils.dataset_registry` (`ATOMIC_TASK_DATASETS` / `COMPOSITE_TASK_DATASETS`); `create_robocasa_envs` resolves it per task when `episode_length` is `None`, and logs the chosen horizon per built vec env. The registry-read except path now logs the failure (via `acc_print`) so a renamed/moved registry surfaces a diagnostic instead of silently reverting every task to the 1000-step cap.
- `envs/configs.py`: `RoboCasaEnv.episode_length` is now `int | None` (default `1000` unchanged for back-compat); setting it to `null` selects the per-task registry horizon.

🗃️ Feature

## How it was tested
- CPU suite: `pytest -m "not gpu and not network" tests/envs/test_robocasa.py` — 62 passed.
- New unit tests (`TestOfficialTaskHorizon` + two `TestCreateRoboCasaEnvs` cases) lock in the behavior with a faked `dataset_registry`: known atomic/composite task → its horizon, float horizon coerced to `int`, unknown task / missing-`horizon` entry → `None`, import failure → `None` (1000 fallback); and `create_robocasa_envs(episode_length=None)` threads the resolved horizon into `_make_env_fns` while an explicit int skips the registry lookup.
- Validated live on a 2-node (16×A100) DeepSpeed ZeRO-2 eval: 16 `atomic_seen` tasks sharded one-per-GPU, each built at its official per-task horizon (900/750/600/450), per-task success gathered across ranks, clean exit.

```bash
pytest -sx tests/envs/test_robocasa.py
```
Set `env.episode_length=null` in a RoboCasa eval config to use the official per-task horizons; set an int to keep a single global cap.

## Checklist

### DIFF
--- a/src/opentau/envs/configs.py
+++ b/src/opentau/envs/configs.py
@@ -329,6 +329,9 @@ class RoboCasaEnv(EnvConfig):
             auto-expands to the upstream task list and auto-sets ``split``.
         fps: RoboCasa control frequency (Hz); also the ``render_fps`` for videos.
         episode_length: Maximum steps per episode (``_max_episode_steps``).
+            Defaults to 1000; set to ``null`` (None) to use RoboCasa's official
+            per-task horizon from the dataset registry (e.g. OpenCabinet=1050,
+            TurnOnMicrowave=450) instead of a single global cap.
         obs_type: ``"pixels"`` or ``"pixels_agent_pos"``.
         render_mode: Rendering mode for the environment.
         camera_name: Comma-separated raw RoboCasa camera names to render. The
@@ -356,7 +359,7 @@ class RoboCasaEnv(EnvConfig):
 
     task: str = "CloseFridge"
     fps: int = 20
-    episode_length: int = 1000
+    episode_length: int | None = 1000
     obs_type: str = "pixels_agent_pos"
     render_mode: str = "rgb_array"
     camera_name: str = "robot0_agentview_left,robot0_eye_in_hand,robot0_agentview_right"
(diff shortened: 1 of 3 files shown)


shuheng-liu:
Addressed the two review notes in ddcba70:

- **suggestion (tests):** Added `TestOfficialTaskHorizon` (known atomic/composite task → registry horizon, float→`int` coercion, unknown task → `None`, missing-`horizon` entry → `None`, import failure → `None`) plus two `TestCreateRoboCasaEnvs` cases proving `episode_length=None` threads the resolved horizon into `_make_env_fns` and an explicit int skips the registry lookup. They fake `robocasa.utils.dataset_registry` the same way the existing tests stub `robocasa` in `sys.modules`, so they stay sim-free. `tests/envs/test_robocasa.py` is now 62 passed. Also corrected the PR body, which previously described a manual check rather than committed tests.
- **nit (silent degradation):** the `except` path in `_official_task_horizon` now logs via `acc_print` (task name + exception type/message + the 1000-step fallback), so a renamed key / moved module is diagnosable instead of invisible.

PR #447, opened 2026-06-28
feat(pi07_paligemma): AR discrete-action decode + goal-frame eval infra

shuheng-liu:
## What this does

Adds two related eval-side capabilities (🗃️ Feature):

**1. Experimental AR (autoregressive) discrete-action decode for `pi07_paligemma`.**
A new eval-only path, gated by `policy.eval_use_discrete_actions` (default `False`),
makes `sample_actions` ignore the flow-matching action expert and instead
autoregressively generate FAST discrete-action tokens with the VLM backbone:
greedy `argmax` over the `da_head` logits using the training-identical `"Action: "`
prefix layout/positions, then a relaxed FAST decode and inversion of the `MIN_MAX`
discrete-action normalization. There is no learned EOS (training masks the CE loss
at pad positions), so generation stops *structurally* — a complete FAST sequence
BPE-decodes to exactly `chunk_size * max_action_dim` DCT coefficients — and is
capped at `discrete_action_max_length` like training truncation. The default
flow-matching path is byte-for-byte unchanged. Standalone eval fails fast under
FSDP / DeepSpeed ZeRO-3 when this flag is set (the per-rank-variable decode loop
would desync the sharded-param all-gathers and hang at NCCL), mirroring the
existing in-training guard.

**2. Eval infrastructure for goal-image conditioning experiments.**
- `eval.goal_frames_dir`: harvest the final raw camera frames of every
  **successful** eval episode as PNGs + a `manifest.csv`, keyed by per-episode
  scene seed (memory-cheap; no-op when unset). Multi-rank safe: per-rank rows are
  gathered to a single rank-0 manifest writer and PNG filenames are rank-suffixed,
  so concurrent ranks never race/tear writes; only the real (non-padded) episodes
  are harvested; manifest columns are sized to the cameras actually captured.
- `eval.seed_list`: evaluate an explicit comma-separated set of scene seeds instead
  of the contiguous `start_seed` range (right-padding the final batch), for
  targeted goal-frame backfill of the exact scenes a prior checkpoint failed.
- `env.subgoal_frames_dirs` + `RoboCasaGoalFrameSubgoalGenerator`: feed a harvested
  success terminal frame as an eval-time subgoal, keyed by the exact per-episode
  scene seed (`camera{k} -> subgoal{k}`) and merged across dirs; scenes without a
  harvested success (or missing a camera) are padded (no subgoal). Per-env scene
  seeds are threaded into `SubgoalImageGenerator.start_episode`.

The eval-path `embed_prefix` calls pass `sync_across_ranks=False` so independent
per-rank sim rollouts don't desync at NCCL on the cross-rank branch all-reduce
(the default `True` used in training is unchanged). The branch was synced with
`main` (the already-merged seeding work reconciles cleanly) and temporary
profiling scaffolding was removed.

## How it was tested

- CPU suite (`pytest -m "not gpu and not network"`) green: **506 passed** across
  `tests/scripts/`, `tests/envs/`, and `tests/policies/test_pi07_paligemma_low_level.py`.
- New CPU tests:
  - `tests/policies/test_pi07_paligemma_low_level.py` — AR decode
    (`_decode_discrete_action_tokens`: shape, over-length truncation, under-length
    zero-coeff pad, out-of-alphabet snap, zero-range MIN_MAX denominator snap), the
    FAST stop rule, and the `include_action_indicator` prefix layout.
  - `tests/scripts/test_eval.py` — `seed_list` parse + final-batch right-pad, and a
    goal-frame harvest writer→reader round-trip (`eval_policy` → `manifest.csv` →
    `RoboCasaGoalFrameSubgoalGenerator`), incl. the `decoder` column toggling with
    `eval_use_discrete_actions`.
  - `tests/envs/test_subgoal.py` — `RoboCasaGoalFrameSubgoalGenerator` (scene-seed
    keying, no-harvest padding, cross-dir merge).
- The kept `sync_across_ranks=False` eval-path behavior is covered by the existing
  `test_global_run_image_tower_sync_across_ranks_false_skips_collective`.
- Policy-related changes: the targeted GPU pytests
  (`pytest -m gpu tests/policies/test_pi07_paligemma_low_level.py`) pass on the GPU
  dev box (1 passed, 1 marker-skip).

```bash
# Subgoal / goal-frame eval infra (CPU):
pytest -sx tests/envs/test_subgoal.py
```
```bash
# AR-decode + seed_list + harvest round-trip (CPU):
pytest -m "not gpu and not network" tests/policies/test_pi07_paligemma_low_level.py tests/scripts/test_eval.py
```
```bash
# Enable the experimental AR decode at eval time (otherwise default flow path):
opentau-eval --accelerate-config <yaml> --config_path=<eval.json> --policy.eval_use_discrete_actions=true
```

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,7 @@ Whether you use the official OpenPi codebase or LeRobot’s reimplementation, yo
 - Dropout in the VLM to reduce overfitting
 - Authentic $\pi_{0.6}$ policy with a Gemma 3 4B backbone, 448×448 vision, and 5-step flow matching
 - Hierarchical $\pi_{0.7}$ policy: high-level planner + low-level controller with a SpaceTime SigLIP video encoder on a Gemma 3 backbone
+- `cosmos3` policy: the $\pi_{0.5}$ flow-matching recipe on a frozen Qwen3-VL-32B (NVIDIA Cosmos3-Super reasoner) backbone with a custom sub-1B action expert
 - A reinforcement learning pipeline described in $\pi^*_{0.6}$
 - And more...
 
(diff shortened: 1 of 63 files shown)


shuheng-liu:
Fixed in 9ff4908 — the `@parser.wrap()` decorator was displaced onto the new `_eval_uses_sharded_params` helper when it was inserted above `eval_main`; moved it back onto `eval_main` so the `opentau-eval` entry point is wrapped again. Added a regression test (`TestEvalUsesShardedParams`) that exercises the helper directly and asserts `eval_main.__wrapped__` is set, so this displacement is caught by the CPU suite next time.

PR #481, opened 2026-07-23
fix(policies): openpi-faithful zero-range normalization, gated by config_version

shuheng-liu:
## What this does

Fixes OpenTau's padding-vs-normalization divergence from openpi, gated so no existing checkpoint is invalidated. (🐛 Bug)

Under MIN_MAX and QUANTILE normalization, a **zero-range band** (`max == min`: a zero-padded state/action tail dim, or a genuinely-constant real dim) mapped to **`-1.0`**. openpi normalizes only the real dims (its `Normalize` slices `stats.q01[..., : x.shape[-1]]`) and zero-pads *after* normalization, so its padded columns are a `0.0` pad constant. Verified numerically against an openpi DROID `norm_stats.json` padded to 32 dims: 24 of 32 action dims and 24 of 32 state dims differed by a full `1.0` — the model saw a constant `-1` across three-quarters of its input width where openpi sees `0`.

OpenTau keeps padding in the dataset and reproduces openpi's **output** arithmetically, so nothing about where padding happens has to move:

```
shifted = deviation + 0.5 * (denom_snapped - (max - min))   # numerator offset
out     = shifted / (denom_snapped + EPS) * 2 - 1
```

The offset is a **float-exact no-op on healthy dims** (both terms derive from the same `max - min`, so the subtraction cancels bit-for-bit) and equals `0.5` exactly where the zero-range guard fired, cancelling the `* 2 - 1` re-centering so a padded column lands on `0.0`. Healthy dims stay **bit-identical**, `Unnormalize(Normalize(x))` **round-trips exactly**, and MEAN_STD is untouched (it already emitted `0.0` on a zero-std dim). The zero-range guard is deliberately kept: on a constant *real* dim with a deviating value, openpi divides by `~1e-6` and emits `~1e6` (the "outlier normalized state" blow-up); OpenTau stays bounded and invertible at `2 * deviation`. So this is openpi-faithful where openpi is correct, and deliberately divergent only where it is numerically broken.

**This is breaking** — every pre-fix checkpoint learned a constant `-1.0` across its padded dims — so it is gated on a new `config_version` schema field that resolves **automatically**, with no manual config:

- **Fresh run**: resolves to the current version → new `0.0` behavior.
- **Loading pre-fix weights** whose config carries no tag: resolves to `0` (legacy `-1.0`), so existing checkpoints, and serving / ONNX export of them, keep their trained behavior. Resolution happens in `PreTrainedPolicy.from_pretrained` (the single chokepoint every weight load crosses), keyed off the weights — it peeks `config.json`, `train_config.json`, and `hf_config.json`.
- **Fine-tuning from a pre-fix checkpoint stays on `config_version` 0**, including the new checkpoints it writes; `--policy.config_version=1` opts a fine-tune into the new convention.
- **Resume across this change** keeps `config_version` 0 mid-run (no discontinuity).

Every saved checkpoint self-describes (`config.json` / `train_config.json` carry a concrete `config_version` plus an informational `opentau_version`). The FAST discrete-action tokenizer bakes the normalization convention into its BPE corpus, so it is versioned in lockstep: `fit_fast_tokenizer.py --config-version` writes an `opentau_action_norm.json` sidecar, and a policy raises at construction if the tokenizer's convention disagrees with its `config_version`. Upstream / pre-versioning tokenizers (e.g. `physical-intelligence/fast`) carry no sidecar and are a no-op.

Adds `CHANGELOG.md` (this is the first entry), documenting the breaking change, the migration steps (discrete-action runs at v1 need a tokenizer re-fit; older installs can't read the new config key, so mixed-version resume is unsupported), and the openpi comparison. See it for the full migration notes.

## How it was tested

- **New** `tests/policies/test_config_version_norm.py` (25 tests, CPU): the MIN_MAX/QUANTILE flip to `0.0` under v1 with healthy dims bit-identical; MEAN_STD unaffected; `Unnormalize` round-trip exact under both conventions; constant-real-dim deviation preserved (not flattened); the `config_version` method contract, save-time concretization, provenance peek (incl. `hf_config.json`), and the FAST tokenizer convention guard (match passes / mismatch raises / absent-sidecar no-op).
- **Parametrized** the existing zero-range tests in `tests/policies/test_normalize_per_dataset.py` and the inline-inverse test in `tests/policies/test_pi07_paligemma_low_level.py` over both conventions.
- **Determinism** (CLAUDE.md rule 3): the change is a config-gated deterministic branch with no new RNG. Verified the legacy (v0) `Normalize` path is **byte-identical** to the pre-change formula — so legacy loss curves are unchanged by construction — and that two `Normalize` calls are bit-identical under both conventions. The full seeded train-loop smoke was not run here (the config-gated branch introduces no stochastic op, and the shared `Normalize` path is covered by the GPU forward tests above); worth one on a box with a full CUDA toolkit before merge.
- **CPU suite** (`pytest -m "not gpu and not network" -n auto`) green — no new failures vs. baseline (only the pre-existing LIBERO env-import failures remain).
- **GPU suite** (`pytest -m "gpu"`) run on the GPU dev box for the affected policy families (pi05, pi06, pi07, pi07\_paligemma, shared policies).
- pre-commit clean.

```bash
pytest -sx tests/policies/test_config_version_norm.py
```
```bash
pytest -sx tests/policies/test_normalize_per_dataset.py -k zero_range
```
```bash
# Fit a FAST tokenizer under the new convention (writes the version sidecar):
python src/opentau/scripts/fit_fast_tokenizer.py --config-version 1 --help
```

## Checklist


shuheng-liu:
Thanks for the review — both addressed in 31a503a:

**1. Hub-hosted tokenizer guard (`pretrained.py`)** — good catch, this was a real silent gap. The sidecar read is now scoped by path type: a local tokenizer directory reads off disk (network-free, as before), while an **HF-repo** tokenizer attempts a network read for `opentau_action_norm.json`. `AutoProcessor.from_pretrained` already contacted that repo to load the tokenizer, so this adds no new dependency, and the reader still fails open — offline/air-gapped construction stays a no-op rather than crashing. The one CPU-gated test that builds a discrete-action policy with a remote path (`test_pi07_cpu.py`) now stubs the hub read too, so the suite remains network-free. New tests assert local dirs never call `hf_hub_download` and remote repos attempt it + fail open.

**2. Stale "mirrors" docstring (`diagnose_norm_distribution.py`)** — fixed the docstring rather than re-plumbing `config_version` into a diagnostic that isn't in this change's scope. It now states it implements the base MEAN_STD/MIN_MAX/QUANTILE formula on the `_slice_stat`-trimmed real dims and deliberately omits the zero-range snap / `zero_range_center` convention; since padded columns are sliced out upstream, the only divergence from the model is genuinely-constant real dims (whose diagnostic z-scores stay unbounded — a useful degeneracy signal).

PR #482, opened 2026-07-23
fix(policies): keep SigLIP patch/pos embeddings in float32 (openpi parity)

shuheng-liu:
## What this does

Fixes a checkpoint-load precision bug in the PaliGemma / SigLIP vision tower shared by `pi0`, `pi05`, `pi05_mem`, and `pi07_paligemma`, and makes the fix reach the served model. (🐛 Bug)

**Root cause.** `to_bfloat16_like_physical_intelligence` cast the **entire** SigLIP tower to bfloat16 at build time, so when `from_pretrained` loaded the float32 converted checkpoint, the **patch-embedding conv** and **position-embedding table** were rounded to bfloat16 — permanently (a later `.to(float32)` cannot recover the bits). Those two tables are the only float32-master weights in the tower (Big Vision keeps "patch extraction and posemb in float32"); every other vision weight is a bfloat16 master upcast on export, so only the embeddings lost precision. On a converted openpi `pi05_droid` checkpoint, `embed_image` diverged from openpi's own PyTorch port by up to **0.395** on the `(1, 256, 2048)` image tokens while every other vision weight loaded bit-identically. Not a SigLIP forward-structure difference, not an attention-kernel difference.

**Fix** (mirrors openpi's `to_bfloat16_for_selected_params`):

1. Pin the three SigLIP embedding tensors to float32 in `to_bfloat16_like_physical_intelligence` (`pi0` + `pi05`; `pi05_mem` / `pi07_paligemma` import `pi05`'s `PaliGemmaWithExpertModel`).
2. Bridge the resulting float32-embeddings → bfloat16-encoder boundary in **all three** hand-rolled SigLIP forwards, so bfloat16 serving doesn't hit a dtype mismatch: the patched `SiglipVisionTransformer.forward` (`utils/transformers_patch.py`), `RLDXVideoEncoder`, and `SpaceTimeSiglipVideoEncoder` (the default pi05_mem encoder + pi07_paligemma low-level's; the cast is placed before the temporal-mask build, which derives from `hidden.dtype`). No-op for single-dtype towers, so the Gemma3 `pi06`/`pi07` towers are unaffected.
3. Preserve the pinning across the blanket `policy.to(torch.bfloat16)` in the inference / serving entry points (gRPC server, `inference.py`, `eval.py`, `benchmark_inference.py`, `high_level_planner_inference.py`) via `to_dtype_preserving_siglip_float32`, which snapshots the float32 embeddings before the cast and restores them after. Only params float32 before the cast are preserved (no-op for uniform-bf16 towers).

**Out of scope / follow-up (#483):** the float32 tables are used in the patch-conv / posemb *compute* in float32 only at inference / serving / ONNX export. No *training* backend computes them in float32 — DDP and non-FSDP DeepSpeed blanket-cast to bfloat16, and FSDP keeps a float32 master but computes bfloat16 via `MixedPrecision(param_dtype=bf16)`. The gap is small (~0.14% on the image tokens, ~1e-4 on the actions; the default freezes the vision encoder) and needs backend-specific, multi-rank-validated handling — tracked in #483.

## How it was tested

Byte-identical-input parity harness (float32, CPU, eager) loading the **same** converted openpi `pi05_droid` weights into OpenTau's `PI05Policy` and openpi's `PI0Pytorch`, with the JAX model as ground truth:

- **`embed_image` vs openpi:** max|diff| **0.395 → 1.07e-4** after the fix (44/256 → 0/256 tokens > 0.05); every vision weight now loads bit-identically. OpenTau matches openpi's fidelity to the original JAX model exactly (both 1.05% vs JAX — the shared XLA-vs-Torch float32 floor).
- **bfloat16 serving:** `embed_image` runs in native bfloat16 (float32 embeddings + bfloat16 encoder) without a dtype crash — validates the forward bridges.
- **Serving helper:** validated on the real `pi05` model — after `to_dtype_preserving_siglip_float32` the patch/position embeddings are float32, the encoder is bfloat16, and `embed_image` runs.
- **Downstream:** toggling only the embedding precision through the full denoise moves the sampled actions by ~1.4e-4 max (0.005% of range).
- New `tests/policies/test_siglip_embedding_dtype.py` (CPU): the patched forward is a no-op in a single-dtype tower and bridges float32 → bfloat16; `RLDXVideoEncoder` and `SpaceTimeSiglipVideoEncoder` (T=1 and T=2) survive the mixed-dtype tower; the helper preserves the pinned embeddings and is a no-op for a uniform-bf16 tower; the pinning selectors match the transformers SigLIP parameter names.
- `pre-commit run --all-files` clean; CPU suite green (`pytest -m "not gpu and not network" tests/policies tests/scripts` → 972 passed); targeted GPU suite green on a CUDA box (`pytest -m "gpu" -n 0` over the affected policies → 9 passed, 2 skipped). Full regression runs nightly via `regression_test.yml`.

```bash
pytest -sx tests/policies/test_siglip_embedding_dtype.py
```

```bash
pytest -m "gpu" -n 0 tests/policies/test_pi05_mem_gpu.py tests/policies/test_pi07_paligemma_low_level.py
```

## Checklist

### DIFF
--- a/src/opentau/policies/pi0/paligemma_with_expert.py
+++ b/src/opentau/policies/pi0/paligemma_with_expert.py
@@ -285,14 +285,14 @@ class PaliGemmaWithExpertModel(PreTrainedModel):
         encoder-dtype hand-off is done by the patched ``SiglipVisionTransformer.forward`` in
         :mod:`opentau.utils.transformers_patch`.
 
-        Caveat: this pinning survives paths that keep the built model as-is — the gRPC
-        serving path (``scripts/grpc/server.py``), ONNX export
-        (``scripts/export_to_onnx.py``), and FSDP training (``train.py`` skips the cast
-        below under FSDP). Several other entry points (``scripts/inference.py``,
-        ``eval.py``, ``benchmark_inference.py``, ``high_level_planner_inference.py``, and
-        non-FSDP ``train.py`` / ``profile_step.py``) run a blanket
-        ``policy.to(torch.bfloat16)`` after load that rounds these tables back to bfloat16;
-        making the float32 masters survive those casts (e.g. re-establishing them from the
+        Caveat: this pinning only survives paths that keep the built model as-is — ONNX
+        export (``scripts/export_to_onnx.py``, which runs in float32) and FSDP training
+        (``train.py`` skips the cast below under FSDP). Every other entry point runs a
+        blanket ``policy.to(torch.bfloat16)`` after load that rounds these tables back to
+        bfloat16, undoing the pinning: the gRPC serving path (``scripts/grpc/server.py``),
+        ``scripts/inference.py``, ``eval.py``, ``benchmark_inference.py``,
+        ``high_level_planner_inference.py``, and non-FSDP ``train.py`` / ``profile_step.py``.
+        Making the float32 masters survive those casts (e.g. re-establishing them from the
         checkpoint after the cast, without breaking DeepSpeed/DDP param handling) is a
         follow-up.
         """
--- a/src/opentau/policies/pi0/paligemma_with_expert.py
+++ b/src/opentau/policies/pi0/paligemma_with_expert.py
@@ -285,16 +285,18 @@ class PaliGemmaWithExpertModel(PreTrainedModel):
         encoder-dtype hand-off is done by the patched ``SiglipVisionTransformer.forward`` in
         :mod:`opentau.utils.transformers_patch`.
 
-        The pinning is preserved by ONNX export (``scripts/export_to_onnx.py``, float32),
-        FSDP training (``train.py`` skips the cast below under FSDP), and the inference /
-        serving entry points (the gRPC server, ``scripts/inference.py``, ``eval.py``,
-        ``benchmark_inference.py``, ``high_level_planner_inference.py``,
-        ``actions_mse_loss.py``), which do their blanket bfloat16 cast through
-        :func:`opentau.policies.utils.to_dtype_preserving_siglip_float32` so these tables
-        stay float32. The remaining gap is the non-FSDP training cast (``train.py`` /
-        ``profile_step.py`` under DeepSpeed / DDP), which still blanket-casts to bfloat16:
-        re-introducing float32 params after the cast there would change how ZeRO / DDP
-        partition parameters and needs multi-rank validation (deferred to a follow-up).
+        These tables are used in the patch-conv / posemb *compute* in float32 only at
+        inference / serving / ONNX export: the inference entry points (the gRPC server,
+        ``scripts/inference.py``, ``eval.py``, ``benchmark_inference.py``,
+        ``high_level_planner_inference.py``, ``actions_mse_loss.py``) route their blanket
+        bfloat16 cast through :func:`opentau.policies.utils.to_dtype_preserving_siglip_float32`,
+        and ONNX export runs in float32. No *training* backend computes them in float32: DDP
+        and non-FSDP DeepSpeed blanket-cast the whole policy to bfloat16 (``train.py`` /
+        ``profile_step.py``), and FSDP keeps a float32 master but its
+        ``MixedPrecision(param_dtype=bfloat16)`` materializes bfloat16 compute params in the
+        all-gather. That gap is small (patch-conv compute differs ~0.14% in the image tokens,
+        ~1e-4 in the sampled actions; the default freezes the vision encoder) and is tracked
+        for a backend-specific, multi-rank-validated follow-up in issue #483.
         """
         self.paligemma = self.paligemma.to(dtype=torch.bfloat16)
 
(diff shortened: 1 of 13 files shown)


shuheng-liu:
Thanks for the review — all three addressed in 59cf289.

**#1 (blocking — `RLDXVideoEncoder` bypasses the bridge):** fixed. `rldx_video_encoder.py` hand-rolls the SigLIP forward (`embeddings()` then walks `encoder.layers`) instead of routing through the patched `SiglipVisionTransformer.forward`, so it now reproduces the same encoder-dtype cast right after the embeddings call. The `pi05_mem` GPU suite (`test_pi05_mem_gpu.py`) stays green with the pinned-embeddings / bf16-encoder tower.

**#3 (no test):** added `tests/policies/test_siglip_embedding_dtype.py` (CPU). It locks: the patched forward is a bit-identical no-op in a single-dtype tower; it bridges float32 embeddings into a bfloat16 encoder without a dtype mismatch; `RLDXVideoEncoder` survives the mixed-dtype tower; and the pinning selectors still match the transformers SigLIP embedding parameter names (guards a rename). The full-model pinning assertion stays on the GPU suite since a real `PaliGemmaWithExpertModel` is a ~3B random-init build.

**#2 (blanket `policy.to(torch.bfloat16)` re-rounds the pinned embeddings):** good catch — documented as a scoped caveat in the `to_bfloat16_like_physical_intelligence` docstring rather than fully fixed here. The pinning **does** take effect on the paths that keep the built model as-is: the gRPC serving path (`scripts/grpc/server.py`), ONNX export (`scripts/export_to_onnx.py`), and FSDP training (`train.py` skips the cast under FSDP). It is re-rounded by the blanket cast in `inference.py`, `eval.py`, `benchmark_inference.py`, `high_level_planner_inference.py`, and non-FSDP `train.py` / `profile_step.py`. Making the float32 masters survive those casts robustly (e.g. re-establishing them from the checkpoint after the cast) has to not break DeepSpeed/DDP param handling, which needs multi-rank validation — so I've scoped it to a follow-up rather than land an unvalidated distributed-dtype change in this PR.

shuheng-liu:
Both addressed in 7b3430a.

**Blocking (`SpaceTimeSiglipVideoEncoder` bypasses the bridge):** fixed — good catch. It's the default pi05_mem encoder (`use_motion=False`) and pi07_paligemma low-level's, and it hand-rolls the same embeddings → encoder loop. Applied the bridge right after the embeddings call and **before** the temporal-mask build, so the mask is derived from the bridged dtype. Both video encoders now key off `post_layernorm.weight.dtype` (the unambiguous SigLIP encoder dtype, since a wrapped SpaceTime layer's first parameter can be a temporal module). Added a sibling test covering T=1 (bridge only) and T=2 (bridge + temporal mask). `test_pi05_mem_gpu.py`, which builds the real pi05_mem on the default SpaceTime encoder, stays green.

**Suggestion (docstring overclaims gRPC):** corrected — you're right, `scripts/grpc/server.py` blanket-casts (`self.dtype = torch.bfloat16` then `policy.to(dtype=self.dtype)`); I'd grepped too narrowly. Moved gRPC to the "undone by a blanket cast" list. As it stands the pinning is preserved only by ONNX export (float32) and FSDP training; every other entry point rounds it back. Making it survive those casts (without breaking DeepSpeed/DDP param handling) is the follow-up noted in the caveat.

shuheng-liu:
Expanded the PR to close the serving gap (per the scope decision on the review): added `to_dtype_preserving_siglip_float32` (`policies/utils.py`) and routed the gRPC server, `inference.py`, `eval.py`, `benchmark_inference.py`, and `high_level_planner_inference.py` casts through it, so the float32-pinned SigLIP embeddings now survive the blanket bf16 cast in the served / eval paths. It snapshots the embeddings before the cast and restores them after; only params float32 before the cast are preserved, so it is a no-op for uniform-bf16 (Gemma3) towers.

Validated on the real `pi05` model: after the helper the patch/position embeddings are float32, the encoder is bfloat16, and `embed_image` runs. The non-FSDP training cast (`train.py` / `profile_step.py` under DeepSpeed / DDP) stays a documented follow-up (re-introducing float32 params there changes ZeRO / DDP partitioning). PR description updated; CPU (972 passed) + targeted GPU (9 passed) green.

shuheng-liu:
Fixed in f1815ed — good catch. Routed `actions_mse_loss.py` through `to_dtype_preserving_siglip_float32` (it measures action fidelity, so running the re-rounded model would defeat its purpose) and added it to the entry-point list in both docstrings. Thanks also for confirming the other unrouted casts are intentional — `train.py` / `profile_step.py` / `find_unused_params.py` are the deferred training-path cast, and `calculate_value.py` / `get_advantage_and_percentiles.py` use the value policy, which doesn't pin embeddings. Script + docstring only, no model-forward change.

PR #485, opened 2026-07-23
feat(policies): openpi-faithful QUANTILE norm + delta joint actions

shuheng-liu:
## What this does

Aligns π₀.₅ normalization and delta-action handling with openpi
(`Physical-Intelligence/openpi`), validated **bit-exact** against the reference
implementation. Tag: 🗃️ Feature.

Four openpi-parity fixes plus the flexibility to reproduce a teacher-faithful
DROID setup:

- **QUANTILE normalization parity.** Dropped the `q01→min` / `q99→max` fallback —
  a missing quantile now raises instead of silently rescaling one dataset's row
  by min/max while its peers use quantiles. Set the normalization epsilon to
  `1e-6` to match openpi (was `1e-8`). No clamping: values outside `[q01, q99]`
  extend past ±1 exactly as openpi does (pinned by a test).

- **Indexable state/action.** New per-dataset `state_index` / `action_index`
  select and reorder raw parquet columns (order respected); the normalization
  stats are reindexed identically so the buffers stay aligned with the emitted
  data.

- **Delta joint actions.** Opt-in per-dataset `use_delta_joint_actions` +
  `delta_action_state_map`. The single chunk-start state is subtracted from every
  action in the horizon *before* normalization, matching openpi's `DeltaActions`
  (`actions[:, :dims] -= state[:dims]`, one state broadcast across the chunk).
  Unmapped action dims stay absolute and keep absolute stats, so one action
  vector can mix relative joints with an absolute gripper. Inference re-adds the
  state (openpi's `AbsoluteActions`) — previously missing, so a delta-trained
  policy would have emitted displacements read as absolute joint targets.

- **On-the-fly delta stats.** Because `meta/stats.json` describes *absolute*
  per-frame actions, it no longer describes the training targets once the delta
  transform is on. New `datasets/delta_action_stats.py` computes the delta (and
  state) quantiles directly from parquet — **no video decode** — using a ported
  histogram estimator, and caches them under each dataset's own root. The compute
  is rank-0-only; other ranks wait by polling the filesystem rather than sitting
  in a collective, so a long compute cannot trip the NCCL watchdog. The estimator
  adds a `bounds` fast-path that removes a rebinning error the streaming histogram
  otherwise accumulates on episode-ordered data.

- **`norm_stats_override_path`** kept for teacher-faithful replay of an openpi
  checkpoint, now with a warning recommending a checkpoint with correct stats
  baked in instead.

The **normalization epsilon** also moves to openpi's `1e-6` (was `1e-8`), gated by
`config_version` — folded into the existing v1 openpi-parity behavior (#481)
alongside zero-range centering, since no v1 checkpoint has been saved yet. Legacy
checkpoints (`config_version` 0) keep `1e-8` and normalize exactly as trained;
threaded per-policy via `config.normalization_epsilon()` → `Normalize(eps=...)`.

Two index spaces are involved and easy to confuse — the config's
`delta_action_state_map` is in **original parquet** indices, while the policy's is
in **post-index** positions. A single `resolve_delta_map` helper does the
translation for both the dataset (which applies the delta) and the policy (which
inverts it), so they cannot drift.

### Relationship to recently-merged work (picked up via merge of `main`)

A parity study run for this change surfaced two pre-existing openpi divergences,
both now **fixed on `main` and merged into this branch** — so this PR builds on
them rather than re-doing them:

- **Padding vs normalization order** — #481 (`config_version` 0→1): a zero-range
  padded/constant dim now maps to `0.0` (openpi's pad-after-normalize output)
  instead of `-1.0`. This is exactly the padding divergence the study flagged.
  The epsilon gating above folds into the same v1 behavior.
- **SigLIP vision-tower divergence** — #482: the vision patch/pos embeddings are
  kept in float32 for openpi parity, closing the ~2%-median image-embedding gap
  the study measured between stock-HF and openpi's patched SigLIP.

## How it was tested

- **New tests.** `tests/datasets/test_action_indexing.py` (column select/reorder,
  the two index spaces, the delta transform + exact round-trip, config
  validation) and `tests/datasets/test_delta_action_stats.py` (histogram
  estimator vs `np.quantile`, the `bounds` rebinning fix, delta stats vs a
  brute-force reference, cache hit/miss/corrupt/stale/read-only). Updated the
  QUANTILE-fallback tests to assert the new raise-instead-of-fallback behavior and
  added a no-clamp pin. Added `test_normalization_epsilon_is_version_gated`
  (`config_version` 0 → `1e-8`, ≥1 → `1e-6`, flips together with zero-range
  centering) next to the existing `config_version` tests.
- **CPU suite** (`pytest -m "not gpu and not network" -n auto`): green.
- **GPU suite**: `pytest -m "gpu" -n 0 tests/policies/test_pi05.py` passes on the
  GPU dev box (both π₀.₅ integration tests).
- **openpi parity harness** (against genuine openpi code): `DeltaActions` /
  `AbsoluteActions`, quantile normalize/unnormalize, and the ported histogram
  estimator all match openpi to machine precision (0.0 – 3e-8). Loading the same
  converted openpi weights into both stacks, the language-embedding path is
  bit-identical end-to-end.
- **Determinism**: the new data-path code (delta transform, delta-stats compute,
  inline-vs-multiprocess) is bit-identical across repeated runs.

```bash
pytest -sx tests/datasets/test_action_indexing.py
```
```bash
pytest -sx tests/datasets/test_delta_action_stats.py
```
```bash
pytest -sx tests/policies/test_normalize_per_dataset.py
```

## Checklist


shuheng-liu:
Thanks for the review — addressed in 9c8580d.

**Blocking**
- **val-split `Subset` drops the new per-dataset attrs** — fixed. Added an `_unwrapped_attr` helper that reads `state_index` / `action_index` / `delta_action_state_map` / `delta_action_stats` one level down through the `Subset`, same as `_instance_name_map` / the `.meta` lookup already do. Regression test `test_val_split_subset_wrapper_preserves_per_dataset_settings`.
- **silent delta-map drop for non-pi05 policies** — fixed. `make_policy` now raises when the mixture publishes a delta map but the policy config has no `delta_action_state_map` field (and thus no `sample_actions` inverse), instead of training on deltas that inference would emit as absolute. Pinned the invariant that pi05 is the only config with the field.

**Suggestions**
- **read-only root + multi-rank hang** — fixed. Every rank probes cache-dir writability up front (uniform decision on the shared FS) and raises together if it's read-only, so non-main ranks never park in `_wait_for_main` waiting for a file rank 0 can't publish. Test `test_readonly_root_multi_rank_fails_fast_instead_of_hanging`.
- **VQA + delta co-training** — fixed. VQA entries (no actions, can't opt into deltas) are now exempt from the all-delta-or-all-absolute check.
- **EPS `1e-8 → 1e-6` gating** — there is no `config_version` mechanism in this repo to gate on (grep is empty), and the epsilon change is the explicit ask of this PR (openpi parity). Documented in `normalize.py` why it's immaterial rather than gated: normal dims shift ~10 ULP (below the round-trip tests' tolerance); the only behavioral change is a dim with std in `(1e-8, 1e-6)` now snapping to 1 instead of dividing by ~1e-7 — that dim is essentially constant, and dividing by ~1e-7 is the ~1e6 blow-up the snap guard exists to prevent, so snapping is if anything safer.

**Nit**
- **multi-file quantile pooling ≠ openpi single histogram** — added a docstring note bounding the claim. For a single-file dataset (the RoboLab `*_200` repos, and the common case) the merge is a passthrough and matches openpi exactly; only genuinely multi-file datasets hit the count-weighted-mean approximation, and only on the two quantile rows. Noted the exact fix (shared global bin edges) as a follow-up.

shuheng-liu:
Merged latest `main` and addressed both round-2 items in 2654997.

The branch now builds on #481/#482 rather than pre-dating them, and I went further than the round-2 suggestion on the epsilon:

- **`normalize.py` — "no config_version gate" claim (now false).** Correct — I had a stale `main`. Rather than just reword the comment, the epsilon is now **gated via `config_version`**, folded into the existing v1 openpi-parity behavior (#481) alongside zero-range centering — both-or-nothing, since no v1 checkpoint has been saved yet. `PreTrainedConfig.normalization_epsilon()` returns `1e-6` at v≥1 and legacy `1e-8` at v0; it's threaded per-policy via `Normalize(eps=...)` through all 10 modeling files, so a legacy checkpoint normalizes exactly as trained. New test `test_normalization_epsilon_is_version_gated` pins that it flips together with zero-range centering; CHANGELOG's 0→1 entry now covers both. And per the second half of that note — this branch has now been rebased-equivalent (merged) onto main's `zero_range_center` changes and re-run locally (full CPU suite + the eps-threaded policies' GPU tests), not just trusting merge-commit CI.
- **pin test only covered 3 configs.** Now iterates the whole registry: `{t for t in available_policies if "delta_action_state_map" in make_policy_config(t) fields} == {"pi05"}`, with a message that forces the guard to be revisited if any future config gains the field without the `sample_actions` inverse.

The parity study's two other findings are now closed on `main` and merged in: the padding/zero-range divergence (#481) and the SigLIP vision divergence (#482).

shuheng-liu:
Both round-3 items fixed in 2fff4ee — the two hand-rolled normalization mirrors now inherit the version-gated epsilon:

- **`pi07_paligemma` inline discrete-action inverse.** Now reads `self.normalize_discrete_actions.eps` (the version-gated epsilon of the module it inverts), mirroring how it already reads that module's `zero_range_center`. On a `config_version` 0 checkpoint the inverse now uses `1e-8` and round-trips exactly instead of being ~1e-6 off, and the snap thresholds agree. Dropped the now-unused module-`EPS` import; the test stub carries `eps` so it mirrors the real `Normalize.eps`.
- **`fit_fast_tokenizer._NORMALIZE_EPS`.** Replaced the stale hardcoded `1e-8` with an eps selected off `--config_version` (via `_normalize_eps_for`) and threaded through `_normalize_chunks` / `_normalize_chunks_per_head` / `_drain_mixture`, exactly like `zero_range_center`. A v1 fit now normalizes its BPE corpus with the same `1e-6` the v1 policy uses at runtime; a v0 fit uses `1e-8`. GPU tests for `pi07_paligemma_low_level` pass.

shuheng-liu:
Both round-4 test-coverage items added in 85a44bd — mutation-killing pins so a revert to module `EPS` no longer passes silently:

- **pi07_paligemma stub eps never exercised at a non-default value.** New `test_inline_inverse_honors_the_modules_version_gated_eps`: a MIN_MAX range of `1e-7` sits between `LEGACY_EPS` (1e-8) and `OPENPI_EPS` (1e-6), so a v0 stub (`eps=1e-8`) divides by ~1e-7 (no snap) while a v1 stub (`eps=1e-6`) snaps to denom 1 — the same tokens decode to outputs differing by ~7 orders of magnitude. Reverting the inline inverse to module `EPS` makes both use 1e-6, they become identical, and `assert not torch.allclose(...)` fails.
- **`_normalize_eps_for` untested.** New `test_normalize_eps_for_is_version_gated` pins `0 → LEGACY_EPS`, `1/2 → OPENPI_EPS` (and that the two disagree). Both fixtures assert against the `LEGACY_EPS`/`OPENPI_EPS` constants imported from `normalize.py`, so they track any future change to the values rather than re-hardcoding.

PR #488, opened 2026-07-24
feat(policies): add train_vision_encoder_only flag across all policies

shuheng-liu:
## What this does

Adds a `train_vision_encoder_only` policy-config flag (🗃️ Feature) — the mirror
image of the existing `train_expert_only`. When set, **only** the vision/video
encoder trains and everything else is frozen:

- **Trainable:** the SigLIP / Gemma 3 / Qwen3-VL vision tower **and** its
  multimodal projector (for Qwen3-VL the `merger` / `deepstack_merger_list` lives
  inside `backbone.model.visual`, so it trains with the trunk), plus the pi05_mem
  RLDX `motion_module` (part of the video encoder).
- **Frozen:** the LLM backbone (including pi07's reparented interleaved backbone
  layers), the action expert, the discrete-action heads (`da_head`,
  `discrete_action_embedding`), and all outer projections (`state_proj`,
  `action_in_proj`, `action_out_proj`, `time_mlp_*` / `action_time_mlp_*`,
  cosmos3's `adarms_proj`, optional modality embeddings).

Before this PR there was no way to express "train the encoder only": the two
existing knobs are `freeze_vision_encoder` (freeze the tower) and
`train_expert_only` (freeze the whole backbone, train only the expert) — the
exact opposite of what this adds.

Coverage across every policy: **pi0, pi05, pi05_mem, pi06, pi07 (low + high),
pi07_paligemma (low + high), cosmos3 / cosmos3_nano.** The freeze spans two
levels — the inner `*WithExpertModel` (`set_requires_grad` + `train`) and the
outer flow-matching module — wired through a shared helper
`freeze_policy_level_params_for_vision_only` in `policies/utils.py`. The helper
freezes every outer parameter that is not part of the with-expert model or a
`motion_module`, so the "only the encoder trains" contract can't be broken by
forgetting to enumerate a projection.

Notable per-backbone details:
- **cosmos3**: `run_prefix` keeps its autograd graph alive whenever
  `train_expert_only` is `False`, so gradients reach `backbone.model.visual`. The
  `no_grad` ctx stays keyed on `train_expert_only` alone (documented + tested) so
  vision-only training never becomes a silent no-op.
- **pi07 (current)** exposes the flag through `vlm_config`
  (`--policy.vlm_config.train_vision_encoder_only=true`), exactly like it already
  exposes `freeze_vision_encoder` / `train_expert_only`; its policy-level freeze
  fields are vestigial/unplumbed, so no new top-level field was added there.
- The SpaceTime video encoders (pi07, pi07_paligemma) register no parameters of
  their own — temporal attention reuses the frozen/unfrozen tower's Q/K/V/O — so
  the tower toggle covers them.

Validation: `train_vision_encoder_only` is mutually exclusive with
`train_expert_only` and requires `freeze_vision_encoder=False`.

The field defaults to `False`, so existing checkpoints and configs are byte-for-
byte unaffected — **no `config_version` bump** (this is a training-time knob, not
a checkpoint behavioral convention).

## How it was tested

- Added `tests/policies/test_freeze_vision_only_helper.py` (unit tests for the
  shared helper, including the pi05_mem `motion_module`-stays-trainable case and
  a backward pass that touches only the vision pathway).
- Added `TestTrainVisionEncoderOnly` to `tests/policies/test_pi0.py` (PaliGemma)
  and `tests/policies/test_pi07_cpu.py` (Gemma 3), and three
  `test_cosmos3_train_vision_encoder_only_*` tests to
  `tests/policies/test_cosmos3_cpu.py` (Qwen3-VL) — each asserts that exactly the
  vision-pathway params have `requires_grad=True` and everything else is frozen,
  that `train()` keeps the frozen bulk in eval, and (cosmos3) that a real
  backward populates gradients on `backbone.model.visual` but not on the expert
  or the LLM.
- Config-validation tests for the mutual-exclusion and `freeze_vision_encoder`
  guards on both the with-expert configs and the policy configs.
- Full CPU suite green: `pytest -m "not gpu and not network" -n auto
  tests/policies` → 751 passed, 2 skipped; plus the affected config/dataset
  tests.
- GPU suite (`pytest -m "gpu"`) run on the GPU dev box for the affected policy
  test files across all three backbone families (PaliGemma, Gemma 3, Qwen3-VL) —
  green.

```bash
pytest -sx tests/policies/test_freeze_vision_only_helper.py
```
```bash
pytest -sx "tests/policies/test_pi0.py::TestTrainVisionEncoderOnly"
pytest -sx "tests/policies/test_pi07_cpu.py::TestTrainVisionEncoderOnly"
pytest -sx tests/policies/test_cosmos3_cpu.py -k train_vision_encoder_only
```
Enable it on a smoke run (must pair with `freeze_vision_encoder=false`):
```bash
opentau-train --config_path=configs/examples/pi05_training_config.json \
  --policy.train_vision_encoder_only=true --policy.freeze_vision_encoder=false
```

## Checklist

### DIFF
--- a/src/opentau/policies/cosmos3/configuration_cosmos3.py
+++ b/src/opentau/policies/cosmos3/configuration_cosmos3.py
@@ -82,6 +82,11 @@ class Cosmos3Config(PreTrainedConfig):
         freeze_vision_encoder: Freeze the Qwen3-VL vision tower. Defaults to True.
         train_expert_only: Freeze the entire backbone; train only the expert +
             projections. Defaults to True (cosmos3's intended regime).
+        train_vision_encoder_only: Mirror image of ``train_expert_only`` — train ONLY the
+            Qwen3-VL vision tower (its ``merger`` / ``deepstack_merger_list`` projector lives
+            inside the tower, so it trains too) and freeze the LLM backbone, the action
+            expert, and all projections. Requires ``freeze_vision_encoder=False`` and
+            ``train_expert_only=False`` (they are mutually exclusive). Defaults to False.
         gradient_checkpointing: Checkpoint the expert decoder layers to trade
             compute for activation memory. Defaults to False.
         dropout: Dropout probability inside the expert. Defaults to 0.1.
@@ -156,6 +161,7 @@ class Cosmos3Config(PreTrainedConfig):
     attention_implementation: str = "sdpa"
     freeze_vision_encoder: bool = True
     train_expert_only: bool = True
+    train_vision_encoder_only: bool = False
     gradient_checkpointing: bool = False
     dropout: float = 0.1
 
@@ -193,6 +199,16 @@ class Cosmos3Config(PreTrainedConfig):
         """Validate the configuration."""
         super().__post_init__()
 
+        if self.train_vision_encoder_only and self.train_expert_only:
+            raise ValueError(
+                "`train_vision_encoder_only=True` and `train_expert_only=True` are mutually exclusive. "
+                "cosmos3 defaults `train_expert_only=True`; set it to False to train the vision tower only."
+            )
+        if self.train_vision_encoder_only and self.freeze_vision_encoder:
+            raise ValueError(
+                "`train_vision_encoder_only=True` requires `freeze_vision_encoder=False` — the vision "
+                "encoder cannot be both frozen and the only trained component."
+            )
         if self.n_action_steps > self.chunk_size:
             raise ValueError(
                 "The chunk size is the upper bound for the number of action steps per model "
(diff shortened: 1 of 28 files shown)


shuheng-liu:
Thanks — both addressed in b67c1a6.

1. **KI × `train_vision_encoder_only` (real interaction).** Good catch — knowledge insulation detaches the prefix KV before the expert (`modeling_pi05.py:1574`), so the flow-matching (MSE) loss can't reach the encoder; under `train_vision_encoder_only` it would train on CE alone, and a no-op/`backward()` error when CE weight is 0. Added a `__post_init__` warning + a docstring note across all five `knowledge_insulation` configs (pi05, pi05_mem, pi06, pi07_paligemma-low, and pi07-low, which reads the flag from `vlm_config`). Kept it a warning rather than an error since CE-weighted vision-only training is still a legitimate (if partial) signal.

2. **`motion_module` name contract.** Added a real integration test (`test_freeze_vision_only_keeps_real_rldx_motion_module_trainable` in `test_rldx_motion_cpu.py`) that builds an actual `RLDXVideoEncoder`, runs the helper, and asserts its `motion_module` stays trainable while a sibling projection freezes — so a rename now breaks a test instead of silently freezing the temporal block.

Verified: full CPU policies suite green (752 passed, 2 skipped); GPU suite re-run for the touched-config policies.

shuheng-liu:
Addressed in 5315754 — added `caplog` tests pinning the KI-severing warning on both code paths: the direct policy-level field (`PI06Config`) and the `vlm_config` path (`PI07LowLevelConfig`), plus the no-warning case when `knowledge_insulation=False`.

PR #492, opened 2026-07-30
feat(policies): repo_id@revision checkpoint specs + loader fixes

shuheng-liu:
## What this does

Published checkpoints are moving from **one repo per step** to **one repo per run, each saved step a git tag**, with `main` at the highest step. This adds the loading side of that: a `@<revision>` suffix on any checkpoint path.

```bash
--policy.path=TensorAuto/<runid>-<runname>@6000   # the "6000" tag
--policy.path=TensorAuto/<runid>-<runname>        # main = the latest step
```

The suffix works for `--policy.path`, `--config_path` and the `policy.pretrained_path` field of a config JSON. Parsing lives in `split_repo_revision` (`utils/hub.py`) and is applied at four loader chokepoints, so `make_policy` and the eight direct `from_pretrained` call sites need no changes.

Two behaviors worth calling out:

- **A local path is never split.** An existing path short-circuits, and a path-shaped string (leading `/`, `.`, `~`, a backslash, or more than one `/`) is left alone — so a moved checkpoint still fails as a path rather than as a bogus repo id. Checkpoint directories legitimately contain `@`.
- **Bare repo ids are unaffected**, so every checkpoint already published under the old `<run>-<step>` naming keeps resolving exactly as before.

Wiring this up surfaced four pre-existing bugs, all in the seven policies that override `from_pretrained` (pi05, pi05_mem, pi06, both pi07 low/high levels, both pi07_paligemma low/high levels — i.e. every policy in production use):

1. **The `config_version` normalization gate never ran for any of them.** The CHANGELOG describes `PreTrainedPolicy.from_pretrained` as "the single chokepoint every weight load crosses"; it was not. None of the seven peeked the version, and since `make_policy` always passes `config=`, their `if config is None` config-loading branch never ran either. **A pre-versioning checkpoint loaded through any of them was normalized with the *current* convention instead of the legacy one its weights were trained under** — silent, because the weights load cleanly and only the numbers are wrong. Extracted as `resolve_checkpoint_provenance` and called by all eight loaders immediately before `cls(config)`; the Normalize/Unnormalize modules read `config_version` at construction, so resolving it later is too late. `validate_input_resolution` was missing from the same seven and is resolved alongside it.

2. **Any load failure returned a randomly-initialized policy** with only a printed warning, so a typo'd repo or tag trained from noise. Now raises. The one tolerated case is a local directory with no `model.safetensors` — the DeepSpeed/ZeRO resume shape, where `accelerator.load_state` supplies the weights immediately after `make_policy` returns.

3. **Five of the seven read their own keyword-only parameters back out of `**kwargs`** (`revision`, `token`, `cache_dir`, `proxies`, `resume_download`, `local_files_only`), where keyword-only arguments can never appear — so all six were unconditionally `None` on every load. `revision` in particular meant an `@<step>` suffix would have been silently ignored.

4. **Published repos were not loadable by repo id at all.** They carry only `hf_config.json`, so `--policy.path=<repo>` failed on the missing `config.json` and `--config_path=<repo>` on the missing `train_config.json` — you had to fetch `hf_config.json` by hand. Both config loaders now fall back to it, which also fixes every checkpoint already on the Hub.

Weight resolution is unified in `resolve_pretrained_weights_file` and now happens **before** the model is constructed, so a bad spec fails in milliseconds instead of after a multi-billion-parameter init.

**This changes numbers for legacy checkpoints** served through those five policies: they now correctly normalize as `config_version` 0 instead of the current convention. That is the correction, but it will move outputs — see the Migration notes in `CHANGELOG.md`, including the `--policy.config_version=1` escape hatch to keep the previous behavior.

No `config_version` bump: the convention itself is unchanged, only which one gets applied.

## How it was tested

- **CPU suite green** (`pytest -m "not gpu and not network" -n auto`): 2383 passed. The only failures are pre-existing and environmental — three `tests/envs/test_factory.py` cases and two import errors, all `No module named 'libero'` in this local checkout; none touch this code.
- **GPU suite green** (`pytest -m gpu -n 0`) on an internal GPU dev box, targeted at every policy this PR changes — `test_pi05.py`, `test_pi05_mem_gpu.py`, `test_pi06.py`, `test_pi07_low_level.py`, `test_pi07_high_level_planner.py`, `test_pi07_paligemma_low_level.py`, `test_pi07_paligemma_high_level_planner.py`, `test_policies.py`: **14 passed, 0 failed, 10 skipped**. All ten skips are pre-existing unconditional `@pytest.mark.skip` decorators (memory-bound or marked slow), not related to this change.

New tests, all CPU and offline:

- `tests/utils/test_hub_revision.py` — the parse matrix: an *existing* local dir named `foo@bar`, a *non-existent* absolute path containing `@`, relative / `~` / Windows forms, `refs/pr/1`-style revisions, a trailing `@`, the conflicting-revision `ValueError`, and **idempotency** (`split(*split(s)) == split(s)`, the contract nested `from_pretrained` calls rely on).
- `tests/policies/test_pretrained_revision.py` — resolver forwarding (including the five previously-dropped kwargs) and loud failures for revision-not-found / repo-not-found / gated / entry-missing, each asserted *not* to be the tolerated error type.
- `tests/policies/test_from_pretrained_revision_overrides.py` — AST invariants over the whole policy registry. **These fail on `main`**: the `kwargs.get`-on-own-parameter check fails for 5 of 7 (six parameters each), and the shared-resolver, split, and provenance checks fail for 7 of 7. Includes a call-ordering assertion that provenance is resolved before `cls(config)`, and a completeness check that any *new* `from_pretrained` override either delegates to `super()` or gets added to the reviewed list.
- `tests/policies/test_from_pretrained_revision_behavior.py` — per-policy behavior for all seven: `@6000` reaches the Hub as `revision="6000"`, a bad tag raises instead of returning an untrained model, and `config_version` resolves to `0` for a legacy checkpoint / is inherited when tagged / is never overwritten when set explicitly.

The behavior tests run in ~0.1 s for all seven policies precisely because resolution now precedes model construction — if someone reorders that back, they become minutes-long, which is itself the signal.

```bash
git checkout claude/hf-checkpoint-upload-skill-857d35
uv sync --extra dev
```

The invariants that fail on `main` and pass here:

```bash
uv run pytest -q tests/policies/test_from_pretrained_revision_overrides.py
```

Everything else added by this PR:

```bash
uv run pytest -q tests/utils/test_hub_revision.py tests/policies/test_pretrained_revision.py tests/policies/test_from_pretrained_revision_behavior.py
```

The parse rule in isolation — note the local path with `@` in it is returned untouched:

```bash
uv run python -c "
from opentau.utils.hub import split_repo_revision as s
print(s('TensorAuto/my-run@6000'))   # ('TensorAuto/my-run', '6000')
print(s('TensorAuto/my-run'))        # ('TensorAuto/my-run', None)
print(s('/data/outputs/run@6000'))   # ('/data/outputs/run@6000', None)
"
```

## Checklist


shuheng-liu:
Both review notes addressed in 5eb40c7.

**`CHANGELOG.md:131` — overclaim (suggestion).** Agreed, and fixed. The claim now reads "a failed weight *resolution*", with an explicit **Scope** paragraph stating that the later key-remap / `load_state_dict` stage in those same seven overrides still catches broadly and returns a partially-loaded model. I left that stage alone deliberately rather than folding it in: `load_state_dict(strict=False)` already absorbs key mismatches, so what it actually hides is a genuine shape conflict, and tightening it can unmask the documented partial-load warm-start paths (e.g. pi06 warm-starting from a `lerobot/pi05` checkpoint). That belongs in its own revertible change, and the CHANGELOG now says so instead of implying this PR covered it.

**`src/opentau/configs/policies.py:801` — message vs. candidates (nit).** The message half is right and is fixed: it now names all three candidates instead of just `config.json`.

The `TypeError` half doesn't reproduce, so I haven't changed that behavior. `config_file = None` is caught by the `if config_file is None: return draccus.parse(cls, config_file, ...)` guard a few lines below, which raises `ParsingError` — `load_resolved_config_dict(None)` is never reached. That path is pinned by the pre-existing `tests/configs/test_utils_policies.py::test_from_pretrained`, which points at an empty directory and asserts `pytest.raises(ParsingError)`; it passes on this branch. Happy to be shown otherwise if you were looking at a different call path.

Unrelated but worth noting for anyone reading the checks: the earlier `check-checklist` failure was self-inflicted and not a body problem. `gh pr create` was blocked repeatedly, and the invocation that finally succeeded used `--fill-first`, so the PR briefly had the commit message as its body with no checklist. I replaced the body immediately. Because the workflow reads `context.payload.pull_request.body` from the event payload rather than fetching live, re-running that job just replays the original empty-checklist payload — only a fresh `edited`/`synchronize` event can clear it, which the push above provides.

shuheng-liu:
Correcting an overclaim of my own, found by testing this against a real consolidated repo rather than a mock.

I wrote that the `hf_config.json` fallback makes published checkpoints "loadable by repo id … which also fixes every checkpoint already on the Hub". Finding the file is not the same as decoding it:

- `--policy.path=<repo>@<step>` — **works.** Reads only the `.policy` sub-object.
- `--config_path=<repo>@<step>` — **still fails on older checkpoints.** It decodes the whole pipeline config, so a train config predating an unrelated schema change raises. Concretely, on a real 2-step consolidated repo: ``dataset_mixture.datasets.0``: The fields `grounding` are not valid for `DatasetConfig` — `grounding` was renamed to `vqa` in #124.

I am deliberately **not** adding `grounding` to `_STRIPPED_FIELDS`. That machinery exists for *removed* fields, where dropping them is lossless. `grounding` was **renamed**, so stripping it would let the config load with `vqa` at its default and silently lose a setting the run was actually trained with — a worse outcome than failing loudly. The claim was the thing that was wrong, not the behavior.

Scoped in `CHANGELOG.md` in f6005ed. Same species of overclaim as the `CHANGELOG.md:131` note earlier in this review, so thank you for tuning me to look for it.

Incidentally, that repo is a real instance of the `config_version` bug this PR fixes: its policy config carries no `config_version`, so before this change it would have been normalized under the *current* convention despite being a pre-versioning checkpoint. It now resolves to legacy `0`.

PR #497, opened 2026-08-04
feat(policies): train_state_action_representation_only finetuning mode

shuheng-liu:
## What this does

Adds an opt-in finetuning mode, `policy.train_state_action_representation_only`, for adapting a **generic VLA checkpoint to a new dataset** by training only the parameters that are actually dataset-specific, and freezing everything else.

**Trainable**
1. The discrete-action representation — `discrete_action_embedding` (input table) **and** `da_head` (logits over the FAST action vocabulary), on every policy that supports knowledge insulation. They are untied but are both pure functions of the fitted FAST vocabulary and are invalidated by the same event (a tokenizer re-fit), so they move as one unit — which is how every existing freeze flag already treats them.
2. The state/action projections — `state_proj`, `action_in_proj`, `action_out_proj`. These are the only outer projections whose shape is a function of the robot (`max_state_dim` / `max_action_dim`) rather than of the architecture.

**Frozen**: the VLM backbone (and its tied `lm_head`), the vision/video encoder **and** the multimodal projector, the action expert — plus `time_mlp_in/out` (pi0's `action_time_mlp_*`, cosmos3's `adarms_proj`), the optional modality embeddings, and pi05_mem's RLDX `motion_module`.

The time MLPs are excluded deliberately: they are hidden-size-to-hidden-size, are fed a sinusoidal embedding of flow-matching time alone, and condition the (frozen) expert. This is the same line `per_group_projection` already draws when it gives per-`(robot_type, control_mode)` copies to exactly those three projections and notes that "time_mlp_in/out stay shared".

### Where it applies

Declared on the nine policy types that have such a representation: `pi0`, `pi05` (+ `pi05_continuous_state`), `pi05_mem`, `pi06`, `pi07_low_level`, `pi07_paligemma_low_level`, `cosmos3` (+ `cosmos3_nano`).

- `pi0` and `cosmos3` have no discrete-action pathway, so the mode degenerates to "train the projections only". That is still useful, so it is allowed — with a warning naming the empty bucket.
- `pi06` (and `pi05` with `state_type="discrete"`) bin state into text tokens and have no `state_proj` at all, so there is one fewer projection to train. Also warned.
- The two high-level planners and `value` do **not** declare the field. They own no projections and no live discrete-action modules, so the flag there would train nothing and produce a flat loss curve with no error. A registry-wide test pins that set by equality in both directions.

### Notable implementation points

- **Default-deny.** Both halves freeze *everything* and then re-enable an explicit allowlist resolved **by module object, never by parameter name**. Name matching is unsafe here: `Normalize`/`Unnormalize` register per-dataset statistics as `nn.Parameter(requires_grad=False)`, so they appear in `named_parameters()` as `normalize_*.buffer_state` / `buffer_actions` — any pattern matching `state` or `action` would turn normalization statistics into learned parameters, and the optimizer factory selects on `requires_grad` alone. Default-deny also freezes the `multi_modal_projector` for free, which `freeze_vision_encoder` (tower only) does not.
- **`config.dropout` is a no-op under this flag.** The frozen trunk is pinned to `eval()`, including each wrapper's own `self.dropout` which fires inside the decoder loop. With only an embedding table and one linear head training, trunk dropout is pure variance that no frozen weight can absorb.
- **No `motion_module` carve-out.** `freeze_policy_level_params_for_vision_only` keeps pi05_mem's RLDX temporal block trainable by a name match. The new helper deliberately does not: it is the one module a generic checkpoint does not carry, so inheriting the carve-out would silently train a freshly-initialized zero-gated block in a run whose premise is that only the dataset-facing adapters move.
- **cosmos3 `run_prefix`.** Its backbone forward is gated by `torch.no_grad()` keyed on `train_expert_only`. This mode freezes the backbone too, so the predicate is extended — otherwise the tower would build and retain a full activation graph for a backward that never reaches it (no crash, just an order-of-magnitude activation-memory bill). `train_vision_encoder_only` is explicitly kept **out** of that predicate, for the opposite reason; the docstring now explains both sides.
- **pi07 plumbing.** pi07 is the only policy whose with-expert config is a serialized *field*, so the flag is declared top-level and pushed into `vlm_config` **in both directions**. A True-only push-down would leave two silent half-frozen states — most importantly, reloading an adapter-trained checkpoint with `--policy.train_state_action_representation_only=false` to fully finetune would clear the policy-level flag while the serialized `vlm_config` one stayed True, freezing the VLM and expert while the outer projections unfroze.

Knowledge insulation splits the trainable set across two losses (`state_proj` + the discrete modules are reachable only from CE; `action_in/out_proj` only from MSE). That is a warning, not an error, so a deliberate CE-only or MSE-only run stays possible.

Default `False`; every new code path is flag-gated, so the flag off is a no-op.

## How it was tested

**New tests (95, ~0.4s)**
- `tests/policies/test_state_action_representation_only_helper.py` — the four helpers against hand-rolled fakes: the trainable set, the projector freeze, the time-MLP/modality-embedding/`motion_module` exclusions, the normalization-buffer name hazard, `train()`/`eval()` pinning, and the config validator.
- `tests/policies/test_state_action_representation_only_policies.py` — the **real** wrappers (pi0, pi05, pi06, pi07) and a real `PI05FlowMatching`: observed trainable set, no `nn.Dropout` left training after `__init__` or `train(True)`, `lm_head` and projector frozen, a backward that reaches only the representation params, and pi07's two-sources-of-truth regressions.
- `tests/policies/test_state_action_representation_only_registry.py` — registry-wide invariants. The capable set is pinned by **set equality** (drift in either direction fails), the factory's branch list is parsed by AST so a policy added later cannot escape the sweep, and an AST completeness check requires every modeling file that freezes policy-level params for `train_vision_encoder_only` to handle this flag too. That last one is what would have caught the analogous `train_vision_encoder_only` gap.

**CPU suite** — `pytest -m "not gpu and not network" -n auto tests/policies/ tests/configs/ tests/test_available.py` → **1122 passed, 2 skipped**.

**GPU suite** — run on a GPU dev box, `pytest -m "gpu" -n 0` over `test_pi05.py`, `test_pi06.py`, `test_pi07_low_level.py`, `test_pi07_paligemma_low_level.py`, `test_pi05_mem_gpu.py`, `test_knowledge_insulation.py`, `test_cosmos3_gpu.py`, `test_cosmos3_nano_gpu.py`, `test_policies.py` → **21 passed, 8 skipped, 0 failed**.

**Determinism (CLAUDE.md rule 3)** — `configs/dev/ci_config.json` run twice on GPU at `seed=1000`, 25 steps: the per-step `total_loss` / `mse_loss` / `ce_loss` / `lr` / `grad_norm` series is **bit-identical** (same md5 over the extracted series).

**No-op verification** — the whole diff removes only 17 lines, of which exactly two are executable: cosmos3's `no_grad` predicate and its paired KV `.detach()`, where `backbone_is_frozen = train_expert_only or <new flag>` reduces to `train_expert_only` when the flag is off. Everything else is pure addition inside `if <flag>:` blocks, config fields defaulting `False`, docstrings and imports.

```bash
git fetch origin claude/vla-dataset-params-training-06de1e
git checkout claude/vla-dataset-params-training-06de1e
```

```bash
pytest -sx tests/policies/test_state_action_representation_only_helper.py \
           tests/policies/test_state_action_representation_only_policies.py \
           tests/policies/test_state_action_representation_only_registry.py
```

Inspect the trainable set directly:

```bash
python -c "
from opentau.policies.pi05.configuration_pi05 import PI05Config
from opentau.policies.pi05.modeling_pi05 import PI05FlowMatching
cfg = PI05Config(train_state_action_representation_only=True, state_type='continuous')
m = PI05FlowMatching(cfg, discrete_action_vocab_size=16)
print(sorted(n for n, p in m.named_parameters() if p.requires_grad))
"
```

Or enable it on any training config:

```bash
opentau-train --config_path=configs/examples/pi05_training_config.json \
  --policy.train_state_action_representation_only=true
```

## Checklist

### DIFF
--- a/src/opentau/policies/cosmos3/configuration_cosmos3.py
+++ b/src/opentau/policies/cosmos3/configuration_cosmos3.py
@@ -42,6 +42,7 @@ from opentau.optim.schedulers import (
     CosineDecayWithWarmupSchedulerConfig,
     LRSchedulerConfig,
 )
+from opentau.policies.utils import validate_state_action_representation_only_config
 
 
 @PreTrainedConfig.register_subclass("cosmos3")
@@ -87,6 +88,16 @@ class Cosmos3Config(PreTrainedConfig):
             inside the tower, so it trains too) and freeze the LLM backbone, the action
             expert, and all projections. Requires ``freeze_vision_encoder=False`` and
             ``train_expert_only=False`` (they are mutually exclusive). Defaults to False.
+        train_state_action_representation_only: Train ONLY the dataset-specific state/action
+            representation -- ``state_proj`` / ``action_in_proj`` / ``action_out_proj``, the three
+            projections whose shape is a function of the robot rather than of the architecture --
+            and freeze everything else: the Qwen3-VL backbone, the vision tower, the action expert,
+            and the flow-matching time conditioning (``time_mlp_in`` / ``time_mlp_out`` /
+            ``adarms_proj``). cosmos3 has no FAST discrete-action branch, so unlike π0.5 there is no
+            discrete-action embedding or logit head to train and the mode degenerates to the three
+            projections alone (warned about by the shared validator). Because cosmos3 defaults
+            ``train_expert_only=True``, this flag must be paired with ``train_expert_only=False``.
+            Defaults to False.
         gradient_checkpointing: Checkpoint the expert decoder layers to trade
             compute for activation memory. Defaults to False.
         dropout: Dropout probability inside the expert. Defaults to 0.1.
@@ -162,6 +173,14 @@ class Cosmos3Config(PreTrainedConfig):
     freeze_vision_encoder: bool = True
     train_expert_only: bool = True
     train_vision_encoder_only: bool = False
+
+    # Train ONLY the dataset-specific state/action projections of an otherwise generic
+    # checkpoint; the backbone, the vision tower, the action expert and the flow-matching
+    # time conditioning are all frozen. There is no discrete-action pathway here, so this
+    # is the whole trainable set. Must be paired with train_expert_only=False, which
+    # cosmos3 defaults to True. Defaults to False.
+    train_state_action_representation_only: bool = False
+
     gradient_checkpointing: bool = False
     dropout: float = 0.1
 
@@ -209,6 +228,20 @@ class Cosmos3Config(PreTrainedConfig):
                 "`train_vision_encoder_only=True` requires `freeze_vision_encoder=False` — the vision "
                 "encoder cannot be both frozen and the only trained component."
             )
+        if self.train_state_action_representation_only and self.train_expert_only:
+            # The shared validator rejects this pair too, but its message is generic and
+            # cosmos3 is the one policy that defaults `train_expert_only=True` — so the
+            # collision fires for a user who overrode nothing but the new flag, and a
+            # message about a conflict they never asked for reads like a bug. Name the fix.
+            raise ValueError(
+                "`train_state_action_representation_only=True` and `train_expert_only=True` are "
+                "mutually exclusive, and cosmos3 defaults `train_expert_only=True` — so this flag on "
+                "its own always collides. Set `train_expert_only=False` as well (e.g. "
+                "`--policy.train_expert_only=false`) to train the state/action projections only."
+            )
+        validate_state_action_representation_only_config(
+            self, policy_name=self.type, has_discrete_actions=False
+        )
         if self.n_action_steps > self.chunk_size:
             raise ValueError(
                 "The chunk size is the upper bound for the number of action steps per model "
--- a/src/opentau/policies/cosmos3/modeling_cosmos3.py
+++ b/src/opentau/policies/cosmos3/modeling_cosmos3.py
@@ -52,6 +52,7 @@ from opentau.policies.pretrained import PreTrainedPolicy
 from opentau.policies.utils import (
     PerSampleLoss,
     flow_matching_masked_mse,
+    freeze_policy_level_params_for_state_action_representation_only,
     freeze_policy_level_params_for_vision_only,
 )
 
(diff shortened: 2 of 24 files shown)


shuheng-liu:
Good catch on the gap — the Qwen3VL wrapper genuinely was the one wrapper the new tests never constructed. Fixed in b8bde28, which adds five tests to `tests/policies/test_cosmos3_cpu.py` using the existing `_tiny_cosmos3_config` harness: the observed trainable set (exactly the three projections; the wrapper ends fully frozen, which also pins that `backbone` is frozen rather than `backbone.model`, so the never-called `lm_head` cannot stay trainable), eval-pinning of the frozen trunk, gradient reachability through a real backward, and the config validation.

I mutation-checked them: replacing the `freeze_with_expert_params_for_state_action_representation_only(self)` call with `pass` fails `..._trainable_set` and `..._grad_reaches_only_projections`. So those two are real guards.

On the specific claim about the predicate, though — I have to push back, because I tried to write exactly the test you suggested and it does not discriminate:

> a revert to `self.train_expert_only` alone would regress silently (full backbone activation graph retained, KV not detached)

That failure mode does not occur. This flag freezes **every** backbone parameter, so no leaf requires grad, and autograd therefore builds no graph and retains no activations regardless of the context manager. Measured on the tiny config, toggling only the predicate:

```
no_grad (current)        key.grad_fn=None key.requires_grad=False v.requires_grad=False
nullcontext (mutated)    key.grad_fn=None key.requires_grad=False v.requires_grad=False
any backbone param requiring grad: False
```

I confirmed this the direct way too: I reverted the predicate to `self.train_expert_only` alone and re-ran the KV-detach assertion — it still passed. A test asserting `grad_fn is None` is vacuous here, so shipping it as a regression guard would have been misleading.

Two consequences, both in b8bde28:

1. The `run_prefix` docstring no longer claims the OR prevents "an order-of-magnitude activation memory bill". It now says the OR is belt-and-braces — it keeps the predicate a truthful answer to *is any backbone parameter trainable?* rather than an accident of which flags freeze what, so it stays correct if the mode's freeze set is ever narrowed. I kept the OR for that reason; it is free and self-documenting.
2. The KV test is renamed `..._prefix_kv_carries_no_graph` and its docstring states explicitly that it pins the observable contract (the expert reads the prefix as a constant) and is *not* a guard on the predicate, pointing at the trainable-set and gradient-reachability tests as the real ones.

Net: your underlying point — cosmos3 was untested — was right and is now addressed. The stated mechanism for why it mattered was not, and I would rather have the comment and the test say what is actually true.

PR #498, opened 2026-08-04
fix(datasets): make the train/val split identical on every rank

shuheng-liu:
## What this does

Fixes a silent validation-data leak: **the train/validation split was different on every rank, so the reported validation metrics were measured on data the model had been trained on.**

Two things combined:

1. `utils/random_utils.py::set_seed(seed, accelerator=...)` deliberately offsets the seed per process — `seed += accelerator.process_index * 12345` — so that per-sample draws (image augmentation, optional-key dropout, prompt substitution) are decorrelated across ranks. This is correct and intentional.
2. `datasets/factory.py::make_dataset` took the train/val `random_split` off that same **global** RNG with no explicit generator.

`scripts/train.py` calls `set_seed(cfg.seed, accelerator=accelerator)` and then `make_dataset_mixture(cfg)` on *every* rank, unguarded, so each rank drew a different `randperm` and partitioned the dataset differently.

Because every rank builds its own `Subset` views, each rank's validation frames landed in some other rank's training set. Simulating 8 ranks over a 353,222-frame dataset at a 2% val ratio:

```python
import random, numpy as np, torch

N, val_size = 353222, 7064
sets = {}
for rank in range(8):
    s = 1234 + rank * 12345
    random.seed(s); np.random.seed(s); torch.manual_seed(s)
    _, va = torch.utils.data.random_split(range(N), [N - val_size, val_size])
    sets[rank] = set(va.indices)

len(sets[0] & sets[1])            # 143 — pure chance
len(set.union(*sets.values()))    # 52818 (15% of the dataset)
len(set.intersection(*sets.values()))  # 0
```

The intersection is empty, so **100% of the frames behind the reported `Validation/*` metrics were in at least one rank's training set**. Consequences: validation loss was optimistically biased (partly-memorized frames), `running_best` checkpoint selection was driven by that contaminated metric, and val numbers were not comparable across runs with different world sizes.

### The fix

Scoped to the split — the per-rank offset in `set_seed` is left alone, since it is load-bearing elsewhere. `random_split` now takes an explicit, rank-independent generator seeded from `cfg.seed` only (`val_split_generator`).

The split is now a pure function of exactly three inputs: `cfg.seed`, `len(dataset)`, and the effective `val_split_ratio`. It is independent of `process_index`, `num_processes`, mixture composition, and how much global RNG ran beforehand. Deliberate consequences, documented on `val_split_generator`:

- **World-size change** (resuming on 4 GPUs instead of 8) leaves the split untouched, so validation stays comparable across the resume boundary.
- **Resume** reproduces the same split as long as those three inputs are unchanged. `--config_path` points at the checkpoint's `train_config.json` on resume, so they carry over automatically — but a CLI override re-partitions the data silently, and `warn_if_resumed_split_differs` warns on exactly that. It covers all three inputs: `seed`, the *effective* `val_split_ratio` per dataset (resolving the `None`-inherits-the-mixture-default rule the way `make_dataset` does), and the config-visible determinants of `len(dataset)` — `episodes`, `excluded_episodes` and `revision`, which are all serialized into the checkpoint. The one case it cannot catch is an **out-of-band** length change: the same `repo_id` at the same `revision` re-uploaded with more episodes. Nothing in the checkpoint records the length actually observed, so that is documented on `val_split_generator` rather than detected; recording it would mean touching the checkpoint format, which belongs in its own change.
- **Mixture composition is irrelevant** — adding or reordering datasets does not reshuffle the others' splits. Two datasets of equal length get the same permutation, which is what you want when the same underlying data is listed twice under different configs: the shared permutation keeps the val frames aligned instead of leaking each entry's val half into the other's train half.
- **`cfg.seed = None`** is covered too. `train.py` never calls `set_seed` in that case, so each process would otherwise split off its own entropy-seeded RNG — the same bug by a different route. A fixed `DEFAULT_VAL_SPLIT_SEED` handles it.

I swept the rest of the data path for the same class of problem (a dataset-level partition or cross-rank-consistent decision taken off the global RNG). This `random_split` was the only instance:

- `datasets/compute_stats.py::sample_indices` — deterministic `np.linspace`, no RNG.
- `datasets/delta_action_stats.py` row cap — uniform stride, no RNG.
- `datasets/dataset_mixture.py::HierarchicalSampler` — already uses an explicit generator; samples **with replacement**, so it carries no partition invariant, and per-rank divergence there is the ordinary data-sharding path rather than a leak.
- `datasets/sampler.py::EpisodeAwareSampler` — `torch.randperm` off the global RNG, but it is not wired into the training path (only tests reference it). Left alone rather than changed speculatively.
- The per-sample draws in `lerobot_dataset.py` (subgoal dropout, prompt substitution, segment offsets) are exactly what the per-rank offset exists to decorrelate — correct as-is.

Also added a `Note:` to `set_seed`'s docstring recording that the global RNG is rank-dependent by construction, so cross-rank-consistent decisions must never be taken off it.

## How it was tested

Added 19 tests in `tests/datasets/test_datasets.py`, next to the existing dataset-factory val-split tests. They drive the real `make_dataset` through the production ordering (`set_seed(cfg.seed, accelerator=...)` → build dataset), simulating `process_index` values in one process.

**Four fail on the pre-fix code and pass after it** — verified by reverting just the `generator=` argument and re-running:

```
FAILED test_val_split_is_identical_across_ranks[1000]
FAILED test_val_split_is_identical_across_ranks[None]
FAILED test_val_split_is_independent_of_world_size
FAILED test_val_split_is_independent_of_prior_global_rng_consumption
```

- `test_val_split_is_identical_across_ranks` — parameterized over `seed=1000` and `seed=None`; asserts the val index lists of 8 simulated ranks are all equal to rank 0's.
- `test_val_split_is_independent_of_world_size` — the split at world sizes 8, 4 and 1 is the same.
- `test_val_split_is_independent_of_prior_global_rng_consumption` — an unrelated `torch.rand(12345)` between the seeding and the split does not move it.
- `test_val_split_changes_with_seed` — mutation-killer: a generator hard-coded to a constant would satisfy every cross-rank assertion above while silently ignoring the run's seed.
- `test_val_split_generator_ignores_process_index` — pins the invariant at the source, including the `seed=None` fallback.
- Eight tests over `warn_if_resumed_split_differs`: it fires on a changed seed and on a changed `excluded_episodes`, stays quiet on an ordinary resume, is main-process-only (it runs on every rank), and no-ops without raising when not resuming / validation is off / the checkpoint has no `train_config.json`.
- `test_warn_if_resumed_split_differs_compares_effective_not_raw_ratios` — a per-dataset ratio moving between `None` and an explicit value equal to the mixture default is the same split and must stay silent, while a genuinely different ratio still warns. A diagnostic that cries wolf gets tuned out, which costs exactly the real detection it exists for.
- `test_warn_if_resumed_split_differs_reads_a_real_saved_train_config` — round-trips through the actual `save_pretrained` serializer rather than a hand-built dict, so a change to how the config nests those fields can't make the warning go permanently silent (a warning that never fires looks exactly like a healthy resume).

The three drift-check tests added in the second commit were likewise verified against the previous behavior: reverting just the effective-ratio resolution and the length-determining field list fails all three.

Full runs:

```
uv run pre-commit run --files <changed files>          # clean
uv run pytest -m "not gpu and not network" -n auto tests/datasets/
  → 701 passed, 7 skipped
uv run pytest -m "not gpu and not network" -n auto tests/utils tests/configs tests/scripts
  → 611 passed, 4 skipped
```

(`tests/utils/test_libero_utils.py` fails to import locally on `ModuleNotFoundError: No module named 'robosuite'` — pre-existing, reproduced on a clean checkout of `main`; this environment lacks the `libero` extra that CI installs.)

No GPU tests: this is a `datasets`-scoped fix and touches no policy, optimizer, sampler or training-loop code, so the policy/GPU checklist boxes stay unchecked.

The regression tests:

```bash
pytest -sx tests/datasets/test_datasets.py -k "val_split or resumed_split"
```

To watch them catch the original bug, revert just the generator argument in `src/opentau/datasets/factory.py` — change

```python
train_dataset, val_dataset = torch.utils.data.random_split(
    dataset, [train_size, val_size], generator=val_split_generator(train_cfg)
)
```

back to `torch.utils.data.random_split(dataset, [train_size, val_size])`, then re-run the command above: four tests fail.

The standalone repro of the underlying behavior, no OpenTau needed:

```bash
python -c "
import random, numpy as np, torch
N, val_size = 353222, 7064
sets = {}
for rank in range(8):
    s = 1234 + rank * 12345
    random.seed(s); np.random.seed(s); torch.manual_seed(s)
    _, va = torch.utils.data.random_split(range(N), [N - val_size, val_size])
    sets[rank] = set(va.indices)
print('rank0 & rank1 overlap:', len(sets[0] & sets[1]))
print('union:', len(set.union(*sets.values())))
print('intersection:', len(set.intersection(*sets.values())))
"
```

Full dataset suite:

```bash
pytest -m "not gpu and not network" -n auto tests/datasets/
```

## Checklist

### DIFF
--- a/src/opentau/datasets/factory.py
+++ b/src/opentau/datasets/factory.py
@@ -360,11 +360,13 @@ def val_split_generator(train_cfg: TrainPipelineConfig) -> torch.Generator:
       the resume boundary and across runs with different world sizes.
     * **Resume** reproduces the same split as long as those three inputs are
       unchanged. ``--config_path`` points at the checkpoint's ``train_config.json``
-      on resume, so ``seed`` and ``val_split_ratio`` carry over automatically —
-      but a CLI override of either, or a dataset whose length changed since the
-      checkpoint (re-uploaded episodes, a changed ``episodes`` /
-      ``excluded_episodes`` filter), silently re-partitions the data.
-      :func:`warn_if_resumed_split_differs` warns about the config half of that.
+      on resume, so they carry over automatically — but a CLI override of
+      ``seed``, of a ``val_split_ratio``, or of the fields that decide how many
+      frames a dataset yields (``episodes``, ``excluded_episodes``, ``revision``)
+      silently re-partitions the data. :func:`warn_if_resumed_split_differs`
+      catches all of those. The one case it cannot catch is an out-of-band
+      length change — the same repo at the same revision re-uploaded with more
+      episodes — since the checkpoint records no length to compare against.
     * **Mixture composition is irrelevant**: every dataset draws from a
       generator seeded the same way, so adding or reordering datasets does not
       reshuffle the splits of the others. Two datasets of equal length get the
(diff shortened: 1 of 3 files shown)


shuheng-liu:
Both review findings addressed in ed5959a — thanks, the first one was a real gap and it corrected a claim I'd made in the PR description.

**`episodes` / `excluded_episodes` omitted from the drift check** — agreed, and the docstring was wrong, not just incomplete. Those fields *are* serialized into the checkpoint's `train_config.json` and they determine `len(dataset)`, so a CLI override on resume re-partitioned the data while `seed` and `val_split_ratio` looked untouched. `revision` has the same property, and `repo_id` comes along so the positional per-dataset comparison is only made between like and like. The "not recorded anywhere in the checkpoint" line now says what it should have said all along: only an *out-of-band* length change is undetectable — the same repo at the same revision, re-uploaded with more episodes — because nothing records the length that was actually observed.

**Raw vs effective `val_split_ratio`** — fixed. Comparison is now on the effective ratio, resolving `None` → mixture default the way `make_dataset` does. This also subsumes the mixture-wide default, so the mirror-image false positive is gone too: changing `dataset_mixture.val_split_ratio` when every dataset overrides it no longer warns. Right that it was false-positives-only, but that's still worth fixing here — a diagnostic that cries wolf gets tuned out, which costs exactly the real detection it exists for.

While fixing both I collapsed the two hand-maintained field lists into one extractor that reads an encoded config, fed `cfg.to_dict()` on the live side and the parsed JSON on the saved side. The original shape made this class of bug easy: a key present on only one side is silently dropped from the comparison, i.e. it fails open, which is indistinguishable from a healthy resume. Long episode lists render as `[N items]` so the warning stays readable.

Three tests added, each verified to fail against the previous behavior: `..._flags_a_changed_episode_filter`, `..._compares_effective_not_raw_ratios` (asserts both that the no-op change is silent and that a genuine one still warns), and `..._summarizes_long_episode_lists`. Full dataset suite: 701 passed, 7 skipped.

On the note about not being able to run the suite in the review environment — that's fine, but for the record the `cpu_test.yml` gate does not exercise the pre-fix failure. The value is in the mutation checks I ran locally: reverting just the `generator=` argument fails 4 of the split tests, and reverting just these two drift-check changes fails the 3 new ones.

PR #503, opened 2026-08-04
fix(datasets): seed the training sampler from cfg.seed

shuheng-liu:
## What this does

Follow-up to #498, from the RNG sweep that PR prompted. `WeightedDatasetMixture.get_dataloader` constructed `HierarchicalSampler` with neither `seed` nor `generator`:

```python
sampler = HierarchicalSampler(
    dataset_lengths=ds_lengths,
    dataset_probs=self.dataset_weights,
    num_samples=num_samples_per_epoch,
)
```

despite the sampler declaring both as supported keyword-only parameters. It therefore fell back to a bare `torch.Generator()`.

**That fallback is not unseeded, and this is not a correctness bug.** PyTorch default-constructs generators with a fixed constant — verified identical across separate processes:

```bash
for i in 1 2 3; do python -c "import torch; print(torch.Generator().initial_seed())"; done
# 67280421310721
# 67280421310721
# 67280421310721
```

So the stream was deterministic and identical on every rank. Unlike the train/val split in #498, there was no cross-rank divergence and no data leak. Two things were wrong anyway:

1. **`cfg.seed` never reached the sampler.** Every run of a given mixture drew the same samples in the same order regardless of seed, so a seed sweep varied only model init and augmentation draws while feeding all its runs identical data in identical order. That is not what `seed: Random seed used for training (model initialization, dataset shuffling)` in `configs/train.py` advertises.
2. **The data order rested on a PyTorch implementation detail.** `torch.Generator()`'s default seed is not a documented guarantee; if it ever changed, every run's data order would shift silently.

### The fix

Pass `seed=self.cfg.seed`. It is the raw config value, so it stays **rank-independent** — which matters here for the same reason it does for the split. `accelerate.prepare` shards with the default `dispatch_batches=False`: every rank iterates the whole sampler stream and keeps the batches at its own offset, so the global batch is a partition of one shared sequence only while the ranks agree on it. A rank-dependent seed would make ranks re-draw each other's samples within a step. The class docstring now states that contract.

`cfg.seed = None` keeps today's behavior exactly — `HierarchicalSampler` already treats `seed=None` as "leave the default generator alone" — and that is pinned by a test comparing against an explicit bare `torch.Generator()`, so a `seed=None` run's data order is bit-unchanged by this PR.

### Behavioral change to be aware of

For any run with `cfg.seed` set, **the training data order changes**. Sampling is i.i.d. with replacement from the same weighted distribution, so this is a different draw from the same distribution rather than a distributional change — but two runs of the same config across this commit will not be step-for-step comparable, and a run resumed across it will not continue its previous stream.

### Also documented, not changed

While confirming the resume story I found that the sampler's position is never checkpointed: `accelerate.checkpointing` persists sampler state only when the sampler is its own `SeedableRandomSampler`, and `train.py` uses neither `skip_first_batches` nor `set_epoch`. A resumed run therefore rebuilds the sampler and **restarts the stream from the beginning** rather than continuing it — it replays the sample sequence already consumed before the interruption. Sampling with replacement means it is still drawing from the right distribution, so this is a data-diversity issue rather than a correctness one.

That behavior is identical before and after this PR — this change only alters *which* stream gets replayed. Fixing it is a separate design decision (skip-ahead on resume, or an epoch/step-derived seed), so this PR records it in the `HierarchicalSampler` docstring instead of quietly changing it.

## How it was tested

New `TestHierarchicalSamplerSeeding` in `tests/datasets/test_dataset_mixture.py`, unit-testing the seeding contract with no mixture build, plus one test on the wiring:

- `test_get_dataloader_seeds_the_sampler_from_cfg_seed` — the mutation-killer for this change: asserts `cfg.seed` reaches the sampler's generator. Verified it **fails** when the `seed=self.cfg.seed` line is reverted.
- `test_same_seed_reproduces_the_stream` / `test_different_seeds_give_different_streams` — a seeded run is replayable, and a seed sweep actually varies the data order.
- `test_stream_ignores_the_global_rng` — runs `set_seed(1000, accelerator=...)` at four simulated `process_index` values (the `process_index * 12345` offset) and asserts the stream is unchanged, pinning rank-independence at the source.
- `test_seed_none_preserves_the_legacy_default_generator_stream` — `seed=None` is bit-compatible with the pre-change bare `torch.Generator()`, so unseeded runs are unaffected.
- `test_explicit_generator_wins_over_seed` — precedence order.

Full runs:

```
uv run pre-commit run --files <changed files>          # clean
uv run pytest -m "not gpu and not network" -n auto tests/datasets/
  → 707 passed, 7 skipped
uv run pytest -m "not gpu and not network" -n auto tests/scripts tests/configs
  → 451 passed
```

No GPU tests: this touches only `datasets/`, no policy, optimizer or training-loop code, so the policy/GPU checklist boxes stay unchecked.

```bash
pytest -sx tests/datasets/test_dataset_mixture.py -k "HierarchicalSamplerSeeding or seeds_the_sampler"
```

To see the original behavior, revert the one line — drop `seed=self.cfg.seed` from the `HierarchicalSampler(...)` call in `src/opentau/datasets/dataset_mixture.py` — and re-run: `test_get_dataloader_seeds_the_sampler_from_cfg_seed` fails while the rest still pass, since the sampler always supported seeding and only the caller was not using it.

Confirm the default-generator constant that the old behavior depended on:

```bash
python -c "import torch; print(torch.Generator().initial_seed())"
```

Full dataset suite:

```bash
pytest -m "not gpu and not network" -n auto tests/datasets/
```

## Checklist

### DIFF
--- a/src/opentau/datasets/dataset_mixture.py
+++ b/src/opentau/datasets/dataset_mixture.py
@@ -777,14 +777,18 @@ class HierarchicalSampler(Sampler[int]):
     samples uniformly within that dataset. This avoids multinomial over a huge number of categories (over 2^24)
     by operating at the dataset level.
 
-    Seeding, in order of precedence: an explicit ``generator``, else ``seed``
-    applied to a fresh generator, else a bare ``torch.Generator()``. That last
-    fallback is **not** unseeded — PyTorch default-constructs generators with a
-    fixed constant — so it is deterministic and identical in every process, but
-    it is a PyTorch implementation detail rather than a documented guarantee and
-    it ignores the run's seed entirely. Callers on the training path pass
-    ``seed=cfg.seed`` so the sample stream is reproducible *and* controlled by
-    the run's own seed; see ``WeightedDatasetMixture.get_dataloader``.
+    Seeding. ``generator`` supplies the RNG *object* and ``seed`` sets its
+    state, so when both are given **``seed`` wins** — it is applied to the
+    passed-in generator via ``manual_seed``, re-seeding the caller's object
+    **in place**. Passing only ``generator`` uses it as-is; passing only
+    ``seed`` applies it to a fresh generator; passing neither leaves a bare
+    ``torch.Generator()``. That last fallback is **not** unseeded — PyTorch
+    default-constructs generators with a fixed constant — so it is
+    deterministic and identical in every process, but it is a PyTorch
+    implementation detail rather than a documented guarantee and it ignores the
+    run's seed entirely. Callers on the training path pass ``seed=cfg.seed`` so
+    the sample stream is reproducible *and* controlled by the run's own seed;
+    see ``WeightedDatasetMixture.get_dataloader``.
 
     Whatever the source, the seed must be **rank-independent**: `accelerate`
     shards this stream by having every rank iterate it and keep the batches at
--- a/src/opentau/datasets/dataset_mixture.py
+++ b/src/opentau/datasets/dataset_mixture.py
@@ -776,6 +776,29 @@ class HierarchicalSampler(Sampler[int]):
     r"""With-replacement sampler for a ConcatDataset that first samples a dataset according to `dataset_probs`, and then
     samples uniformly within that dataset. This avoids multinomial over a huge number of categories (over 2^24)
     by operating at the dataset level.
+
+    Seeding, in order of precedence: an explicit ``generator``, else ``seed``
+    applied to a fresh generator, else a bare ``torch.Generator()``. That last
+    fallback is **not** unseeded — PyTorch default-constructs generators with a
+    fixed constant — so it is deterministic and identical in every process, but
+    it is a PyTorch implementation detail rather than a documented guarantee and
+    it ignores the run's seed entirely. Callers on the training path pass
+    ``seed=cfg.seed`` so the sample stream is reproducible *and* controlled by
+    the run's own seed; see ``WeightedDatasetMixture.get_dataloader``.
+
+    Whatever the source, the seed must be **rank-independent**: `accelerate`
+    shards this stream by having every rank iterate it and keep the batches at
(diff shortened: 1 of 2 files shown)


shuheng-liu:
Both findings were correct — fixed in 31b4961. The docstring I added stated the precedence backwards, which is worse than having left it undocumented, since it would have sent a reader the wrong way.

**Precedence.** Verified empirically rather than re-reading the code:

```python
caller_gen = torch.Generator().manual_seed(7)
both = list(HierarchicalSampler(**kw, generator=caller_gen, seed=1000))
both == list(HierarchicalSampler(**kw, seed=1000))                                # True
both == list(HierarchicalSampler(**kw, generator=torch.Generator().manual_seed(7)))  # False
caller_gen.initial_seed()                                                        # 1000
```

So exactly as you described: `generator` supplies the RNG object, `seed` sets its state, `seed` wins when both are given, and the caller's generator is re-seeded **in place**. The docstring now says that, including the side effect.

**Test.** Right that it pinned nothing — it compared `generator=manual_seed(7)` against `seed=7` in two separate constructions, which holds under either precedence. Split in two: one for a generator supplied alone, and `test_seed_beats_generator_and_reseeds_it_in_place`, which passes both and asserts the stream matches `seed=1000` (not `seed=7`) *and* that `caller_gen.initial_seed() == 1000`. The in-place mutation gets an assertion rather than just a docstring line because it is the surprising direction — a reader reasonably expects an explicitly-constructed generator to win, so a future refactor could flip it without anything failing.

I deliberately did **not** change the behavior here. Making the both-passed case raise is defensible — it is arguably a caller error, and no production caller passes both, so nothing would break — but that is an API decision on a shared class and this PR is about wiring `cfg.seed` into one call site. Happy to do it as a follow-up if you'd prefer that over documenting it.

`tests/datasets/` 708 passed, 7 skipped; pre-commit clean.

PR #511, opened 2026-08-12
fix(configs): re-check policy action horizon after action_chunk propagation

shuheng-liu:
## What this does

(🐛 Bug) `TrainPipelineConfig` propagates the pipeline-level `action_chunk` onto `policy.chunk_size`. That `setattr` runs **after** the policy config's `__post_init__`, so the policy's own `n_action_steps <= chunk_size` invariant — enforced in every policy config that has an execution horizon — is only ever checked against the `chunk_size` the policy *declared*, never against the propagated one:

```python
from opentau.configs.train import TrainPipelineConfig
from opentau.configs.default import DatasetMixtureConfig, DatasetConfig
from opentau.policies.pi0.configuration_pi0 import PI0Config
dm = DatasetMixtureConfig(datasets=[DatasetConfig(repo_id="m", root="/tmp/m")])
c = TrainPipelineConfig(dataset_mixture=dm, policy=PI0Config(), batch_size=8, action_chunk=10)
print(c.policy.chunk_size, c.policy.n_action_steps)   # -> 10 50, silently
```

`n_action_steps=50 > chunk_size=10` is exactly what `PI0Config.__post_init__` raises on when it appears in a JSON config, but propagation reached it with nothing raising. Every in-repo config under `configs/` dodges this only by setting `policy.n_action_steps` equal to `action_chunk` by hand; a config that sets top-level `action_chunk` and leaves `n_action_steps` at its default hits it. The same dodge covered the two sibling invariants keyed on `chunk_size` (delay knob `<= chunk_size`, and shortened-horizon-vs-non-zero-delay).

This is independent of #510, which fixed *which* fields propagate (`max_action_state` → `max_action_dim`); `chunk_size` propagated correctly all along. It builds directly on #510's `_propagate_shape_fields_to_policy` — nothing in that helper's copy loop changes here, the re-validation is appended to it.

**Approach — re-validate after propagating, and fail loudly.** Not clamping: silently rewriting `n_action_steps` would change the execution horizon under the user and persist into the saved config.

* New `PreTrainedConfig.validate_action_horizon()` holds the three `chunk_size`-keyed checks, replacing the copy that lived in each of the seven policy configs (pi0, pi05, pi05_mem, pi06, pi07/low_level, pi07_paligemma/low_level, cosmos3 — and so cosmos3_nano / pi05_continuous_state by inheritance). It is a no-op for configs that declare no execution horizon (`value` has `chunk_size` but no `n_action_steps`; the high-level planners have neither).
* Each policy `__post_init__` now calls it, and so does `_propagate_shape_fields_to_policy` once its copy loop is done — which covers both call sites for free, including the `validate()` one that matters on its own (`--policy.path=...` replaces `self.policy` wholesale after `__post_init__` already ran).
* The propagation site re-raises with the pipeline field named, since `policy.chunk_size` reads as untouched in the user's JSON:

  > The chunk size is the upper bound for the number of action steps per model invocation. Got 50 for `n_action_steps` and 10 for `chunk_size`. `policy.chunk_size` was overwritten with the pipeline-level `action_chunk` (10), so the conflict is between `action_chunk` and the policy fields it does not propagate. Set `policy.n_action_steps` to `action_chunk` (or below it, for a deliberately shortened execution horizon), or change `action_chunk` — a policy-level `chunk_size` is ignored either way.

* pi0 names its knob `safety_buffer` where the others name it `max_delay`. The shared validator reads the name off an `action_horizon_delay_field` ClassVar that pi0 overrides — so pi0 also **gains** the `safety_buffer <= chunk_size` bound its siblings already had (its `select_action` refills whenever `len(queue) <= safety_buffer`, and the queue never holds more than `chunk_size`, so a larger value silently re-queries every step).

Three notes for reviewers:

1. **A config with `n_action_steps > action_chunk` now fails at config time.** No in-repo config trips it (checked all of `configs/**` and `tests/artifacts/configs/`), but an external checkpoint could: the model only ever decoded `chunk_size` actions and `select_action` sliced `min(n_action_steps, chunk_size)`, so such a run was *effectively* running at `action_chunk`. Setting `n_action_steps = action_chunk` reproduces what it was doing.
2. **#510's `_make_cfg` test helper needed a one-line change** — `PIPELINE_WIDTHS` sets `action_chunk=13` deliberately disagreeing with every policy default, which includes the default 50-step horizon, so the fixtures were themselves building the illegal pair. The helper now carries `n_action_steps` along when the policy declares it, which is exactly what a real config setting `action_chunk` has to do. Nothing in that module asserts on `n_action_steps`, so no existing pin is weakened.
3. The unified messages name the field (`` `max_delay` must be <= the chunk size ``) instead of prose ("The max delay must be..."). Two existing tests matched on the prose and were updated to match the field name, which is the more robust pin.

## How it was tested

`pre-commit run --all-files` (all hooks pass), then `pytest -m "not gpu and not network" -n auto`: **2637 passed, 13 skipped, 2 failed**, neither from this change. `tests/utils/test_libero_utils.py::test_libero2torch` fails on a missing local LIBERO `init_files` path and fails identically on this branch's base. `tests/utils/test_utils_benchmark.py::test_exception_handling` asserts a 0.05 s `sleep` lands within ±0.01 s of wall clock, which is a coin flip on a machine running the suite under `-n auto`; it passes in isolation, on a clean tree, and on re-run.

Added to `tests/configs/test_shape_field_propagation.py` (#510's module for this helper, reusing its registry sweep and fixtures):

- `test_action_chunk_below_the_policy_execution_horizon_is_rejected` — the reproduction above; asserts the message names both `n_action_steps` and `action_chunk`.
- `test_action_chunk_matching_the_execution_horizon_propagates` — the legal case still propagates.
- `test_the_delay_knob_is_rechecked_against_the_propagated_chunk_size` — re-validation is not just the one check: a `PI05Config(chunk_size=10, n_action_steps=10, max_delay=5)` that is **valid standalone** becomes illegal once `action_chunk=50` widens `chunk_size` underneath it.
- `test_a_policy_without_an_execution_horizon_is_left_alone` — `value` (has `chunk_size`, no `n_action_steps`) still receives the width instead of tripping a guard.
- `test_the_propagated_width_is_rechecked_for_every_policy` and `test_the_policy_config_still_enforces_the_horizon_itself` — both swept over the registry via the module's existing `_all_policy_configs()`, so a policy added later is covered automatically. Currently 9 types each: cosmos3, cosmos3_nano, pi0, pi05, pi05_continuous_state, pi05_mem, pi06, pi07_low_level, pi07_paligemma_low_level.
- `test_validate_rechecks_the_horizon_after_the_pretrained_reload` — pins the `validate()` path via the real `--policy.path` route, with an assertion on the artifact so the test can't quietly stop exercising the conflict.

Added to `tests/policies/test_pi0.py`: `test_guard_rejects_safety_buffer_above_chunk_size`, which also pins pi0's `action_horizon_delay_field` override — without it the validator would read a `max_delay` pi0 does not have and skip both delay checks.

Mutation-checked the new tests rather than trusting them: replacing the `validate_action_horizon()` call at the propagation site with `pass` fails 12 of them (and leaves the legal-case ones passing, as they should).

**GPU suite** (`pytest -m "gpu" -n 0` on an internal GPU dev box, single GPU, `dev` + `libero` extras): **25 passed, 144 skipped, 2 failed**. Both failures are missing sim assets on that box, not this change: a LIBERO `.bddl` file absent under the configured `LIBERO_CONFIG_PATH`, and `robocasa is not installed` (the `robocasa` extra was not synced). Both fail identically on this PR's base commit, checked by re-running the two files at the base in the same worktree. Nothing under `envs/` is touched here.

That run was on this branch's pre-rebase commit, before it was rebased onto #510. The rebase changed only `configs/train.py` (the propagation helper became #510's, with the re-validation appended) and the placement of the new CPU tests: **every file under `src/opentau/policies/` is byte-identical between the two commits**, and the only GPU-marked module that references `TrainPipelineConfig` at all is `tests/envs/test_robocasa.py`, whose GPU test is one of the two already failing for the missing extra.

**Regression tests were not run.** They are the nightly `regression_test.yml` job on an AWS g6.12xlarge; this change adds config-time validation only — no forward, normalization, or data-path code — so it cannot move a loss curve, and the run costs real money. Both checklist boxes below are therefore left unchecked rather than claiming a run that did not happen; say the word and I will dispatch `regression_test.yml` against this branch and check them.

## How to checkout & try?

```bash
pytest -sx tests/configs/test_shape_field_propagation.py
```

```bash
pytest -sx tests/policies/test_pi0.py -k safety_buffer
```

```bash
python -c "
from opentau.configs.train import TrainPipelineConfig
from opentau.configs.default import DatasetMixtureConfig, DatasetConfig
from opentau.policies.pi0.configuration_pi0 import PI0Config
dm = DatasetMixtureConfig(datasets=[DatasetConfig(repo_id='m', root='/tmp/m')])
TrainPipelineConfig(dataset_mixture=dm, policy=PI0Config(), batch_size=8, action_chunk=10)
"
```

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -177,6 +177,25 @@ carry a concrete `config_version` (and an informational `opentau_version`).
   current working directory for a Hub checkpoint (which an `@<step>` suffix would
   have turned into a directory named after a git ref). Hub exports go to
   `<output_dir>/onnx/<spec>/`; local checkpoint dirs are unchanged.
+* **`action_chunk` is re-checked against the policy's execution horizon.**
+  `TrainPipelineConfig` copies its pipeline-level `action_chunk` onto
+  `policy.chunk_size` by `setattr`, which lands *after* the policy config's
+  `__post_init__` has run — so the policy's own `n_action_steps <= chunk_size`
+  invariant (and the shortened-horizon-vs-`max_delay`/`safety_buffer` pair) only
+  ever saw the `chunk_size` the policy *declared*, never the propagated one. A
+  config setting `action_chunk` and leaving `n_action_steps` at its default
+  therefore reached, silently, exactly the combination the very same policy
+  config rejects when it appears in JSON. Every in-repo config dodges it only by
+  setting the two equal by hand. The checks are now one shared
+  `PreTrainedConfig.validate_action_horizon` — replacing a copy in each of the
+  seven policy configs that carried it — called from both policy `__post_init__`
+  and the end of `_propagate_shape_fields_to_policy`, and the error names
+  `action_chunk` as the cause.
+  **A config with `n_action_steps > action_chunk` now fails at config time**; the
+  model always decoded only `chunk_size` actions, so set `n_action_steps` to
+  `action_chunk` (what such a run was effectively doing) to keep its behavior.
+  As a side effect of sharing the check, pi0 gains the `safety_buffer <=
+  chunk_size` bound its `max_delay` siblings already enforced.
 
 ### Migration notes
 
--- a/src/opentau/configs/policies.py
+++ b/src/opentau/configs/policies.py
@@ -28,7 +28,7 @@ import tempfile
 import warnings
 from dataclasses import dataclass, field
 from pathlib import Path
-from typing import Type, TypeVar
+from typing import ClassVar, Type, TypeVar
 
 import draccus
 from huggingface_hub import hf_hub_download
@@ -434,6 +434,14 @@ class PreTrainedConfig(draccus.ChoiceRegistry, HubMixin, abc.ABC):
     action_decoder_latency_lower: float = 0.0
     action_decoder_latency_upper: float = 0.0
 
+    #: Name of this policy's "keep executing past the current chunk" knob — the field a
+    #: shortened execution horizon (``n_action_steps < chunk_size``) cannot be combined
+    #: with. Every policy that models real-time inference delay calls it ``max_delay``;
+    #: pi0 predates that path and calls its overlap-refill knob ``safety_buffer``, so it
+    #: overrides this. Read by :py:meth:`validate_action_horizon`; a policy that names the
+    #: knob something else must override this or its knob silently stops being checked.
+    action_horizon_delay_field: ClassVar[str] = "max_delay"
+
     def __post_init__(self):
         """Initialize post-creation attributes.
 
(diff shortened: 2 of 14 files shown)


shuheng-liu:
Thanks — both nits addressed:

**2 (docstring accuracy)** — fixed in 26fb903. It now reads "Every policy config that declares an execution horizon calls this from its `__post_init__` (`value` and the high-level planners do not, since for them it is the no-op described below)". The original sentence was flatly wrong about three of the twelve registered configs.

**1 (`safety_buffer == chunk_size`)** — leaving as `>`, deliberately. Tightening to `>=` would have to apply to `max_delay` too, or pi0 and its siblings would disagree on where the same knob's bound sits — and for `max_delay` that is a real behavior change to a knob currently in use, not a new guard on an unused one. The `==` case is degenerate but not incoherent, and this PR's job is to stop propagation from *bypassing* the existing bounds, not to move them. Worth its own change if someone wants it.

shuheng-liu:
Fixed in 41a53cd rather than deferred — the remediation no longer prescribes a fix that can't apply. It now reads "Bring the policy field(s) named above into range of `action_chunk` (`n_action_steps` is normally set equal to it), or change `action_chunk`", which is true for all three failures while keeping the common case concrete.

One thing that fell out of checking the delay-knob case: at the *propagation site* it is unreachable from a standalone-valid policy config. Making `max_delay > chunk_size` fire there needs `chunk_size_new < max_delay <= chunk_size_old`, but a non-zero `max_delay` is only legal standalone when `n_action_steps == chunk_size_old`, so `n_action_steps > chunk_size_new` holds too and the first check pre-empts it. It stays reachable from `__post_init__` (a JSON config with `max_delay > chunk_size`), which is where the existing per-policy tests exercise it. The reachable third case — shortened horizon plus a non-zero delay knob — is covered by `test_the_delay_knob_is_rechecked_against_the_propagated_chunk_size`, and I re-read all three messages end to end against the new wording.

PR #512, opened 2026-08-12
feat(policies): add accel, a cost-free flow-matching uncertainty proxy

shuheng-liu:
## What this does

Implements **denoising acceleration** (`accel`), the cost-free flow-matching uncertainty proxy from [arXiv:2607.27933](https://arxiv.org/abs/2607.27933), read off the samplers' *existing* Euler loop.

A maximally *certain* conditional-flow-matching field is an affine-isotropic contraction toward a point mass, so its denoising trajectory is a straight line at constant velocity. Any bend in that trajectory betrays non-zero posterior covariance. Over the first `p` of `T` Euler steps:

```
accel_p = ( p * sum_{t=1}^{p-1} ||v_t - v_{t-1}|| ) / ( sum_{t=0}^{p-1} ||v_t|| )
```

`v_t` is the velocity the sampler already evaluates to take its Euler step, so this needs **no extra network evaluations, no resampling, no training, and no auxiliary probe** — the whole cost is two vector norms per denoise step.

**Off by default, and inert when off.** `accel_prefix` defaults to `None`; with it unset every added line inside a sampler is dead, so the traced/compiled graph is byte-identical to today. **No `config_version` bump** — nothing about what the model sees or produces changes.

### What's here

- **`opentau/policies/accel.py`** — the estimator (`AccelMeter`), the action-dim mask, `AccelProvenance`, and `configure_accel`, the single enable knob shared by every entry point.
- **`opentau/utils/accel_detector.py`** — offline one-sided CUSUM + split-conformal calibration over the per-chunk stream, fitted on **successful** rollouts only (no failure labels needed).
- **All 7 flow-matching policies wired** — pi0, pi05, pi05_mem, pi06, pi07, pi07_paligemma, cosmos3. Nothing is inherited between them, so this is seven separate edits; a registry-wide AST test pins them all.
- **`opentau-accel-diagnose`** — measures the bfloat16 rounding floor, the run-to-run spread, and the prefix profile of a checkpoint. **Run this first** (see "Known limits").
- **`opentau-accel-calibrate`** — fits a threshold from an eval run's `eval_info.json`.
- **Plumbing** — `eval.py` records a per-episode stream (done-truncated, paired with `success`); gRPC gained `optional float accel = 5`; the RoboCasa server gained an opt-in `accel` envelope; `accel_prefix` is a real field on `PreTrainedConfig`.

### Two design points worth a reviewer's attention

**The meter is passed *into* the sampler, never stashed on the `nn.Module`.** These samplers are `torch.compile`d at four entry points and ONNX-exported at one; a module-attribute write inside a traced region is a Dynamo side-effect and the construct most likely to break under `fullgraph=True`.

**Two masks are mandatory, and both are silent when wrong.** The samplers work in `max_action_dim` (32) while a real embodiment uses far fewer, and the padded tail is masked out of the *training* loss — so those columns are unsupervised network output that would otherwise land in both sums. `real_action_dim` does not exist at inference, so the mask is recovered from the normalization buffers. Separately, frozen real-time-chunking rows and rows outside the executed window carry velocities describing nothing the robot will do, and are excluded.

## How it was tested

**Unit / structural — 227 new tests.**

- `tests/policies/test_accel.py` pins the paper's two anchors numerically against hand-computed values: exactly `0.0` on a constant-velocity path (Theorem 1), strictly positive on any bend, and growth with bend magnitude. Plus the masking cases that would silently corrupt the score, and the undefined-score cases that must be `NaN` rather than `0.0` (`0.0` would read as *maximally certain* — the opposite of *no measurement*).
- `tests/policies/test_pi05_accel.py` drives the real `sample_actions` with scripted velocities: re-plan cadence, frozen-prefix exclusion, and actions bit-identical with the feature on vs off.
- `tests/policies/test_accel_wiring_registry.py` is an AST sweep over all 7 policies. It was **mutation-verified**: moving `accel.update()` after the Euler step, dropping the `if accel is not None` guard, removing the `select_action` clear, and deleting `set_row_mask` each make it fail, and the file restores byte-identically. It also discovers every denoise loop in the tree independently, so a new policy cannot silently escape it.
- `tests/utils/test_accel_detector.py` covers the CUSUM recursion, conformal rank selection, and the persistence round-trip.

**CPU suite:** 2718 passed. The only failures are 3 pre-existing `tests/envs/test_factory.py` cases (the `libero` extra is not installed in this environment); verified identical on a pristine `origin/main`. Pre-commit: all hooks pass.

**GPU suite** (`pytest -m gpu -n 0` over `tests/policies` + `tests/utils`, on an internal GPU dev box): **22 passed, 0 failed.** I additionally stripped the unconditional `@pytest.mark.skip(reason="...24GB")` markers — this card has more memory — and ran the 12 memory-guarded end-to-end tests: 10 passed, 2 failed. Both failures reproduce identically on pristine `origin/main`; they are pre-existing rot in tests that never run because the skip is unconditional, and are being fixed separately. Nightly `gpu_test.yml` will re-verify the merge commit.

**Regression suite** (the `regression_test.yml` sequence, run manually; DeepSpeed ZeRO-2 with `num_processes` reduced 4 → 1 for a single-GPU box): `opentau-train` on `configs/dev/ci_config.json` completed 25 steps. `check_nonzero_grad_norm` PASS, `check_accumulate_grad_sync` PASS, `check_loss_drop` FAIL, `check_state_keys` FAIL — both failures reproduce identically on pristine `origin/main`, and both steps are `continue-on-error` in CI. `convert_checkpoint.sh` PASS, `inference.py` PASS.

**Determinism (CLAUDE.md rule 3):** two seeded 25-step `ci_config` runs produced **bit-identical metric lines (25/25)** with distinct log checksums. Also proved by AST that `forward` is untouched in all 7 policies — only `__init__` / `reset` / `sample_actions` / `select_action` changed, so the training path is not modified at all.

**Real-checkpoint validation.** Measured on a 6-DoF SO101 pi05 checkpoint (`T = 10`, QUANTILE action norm, delta actions, `max_delay = 4`) over 8 real frames of its own training data — a configuration that exercises every guard here, which no shipped config does:

| quantity | value |
|---|---|
| bfloat16 rounding floor | 0.0056 |
| redrawn-noise spread | 0.0094 |
| between-observation spread (the signal) | 0.097 |
| **signal / floor** | **17** |

The prefix sweep on that run reproduces the paper's terminal singularity: step-over-step growth decelerates monotonically (122% → 62% → 69% → 38% → 41% → 33% → 22%) and then *reverses* to +48% at the final step — which is why `default_prefix` returns `T - 1` and not `T`.

The real-time-chunking path was verified on real weights: actions bit-identical with `accel` on vs off, re-plan cadence exactly `n_action_steps - max_delay`, and frozen prefix rows genuinely excluded (180 → 156 scored elements at `delay = 4`).

```bash
pytest -sx tests/policies/test_accel.py tests/policies/test_pi05_accel.py tests/policies/test_accel_wiring_registry.py tests/utils/test_accel_detector.py
```

Measure the noise floor of a checkpoint before trusting any score from it:

```bash
opentau-accel-diagnose --config_path=configs/examples/pi05_libero_eval_config.json --policy.path=<repo_id>@<step> --policy.device=cuda
```

Collect a calibration set from an eval run, then fit a threshold:

```bash
opentau-eval --accelerate-config configs/examples/accelerate_ddp_config.yaml --config_path=configs/examples/pi05_libero_eval_config.json --policy.accel_prefix=auto --eval.n_episodes=60 --env.max_parallel_tasks=1
```

```bash
opentau-accel-calibrate outputs/eval/<run>/eval_info.json --alpha 0.1 --out outputs/accel
```

## Known limits (please read before relying on a score)

These are the method's, not implementation gaps, and they are documented in the module and the CHANGELOG:

- **Measure the floor per checkpoint.** `accel`'s numerator is a difference of nearly-equal vectors, and the serving path runs the action projection in bfloat16, so there is a positive-biased rounding floor beneath every score that compresses exactly the low-uncertainty end the method's premise depends on. The 17× signal-to-floor above is one checkpoint's number, not a constant.
- Serving precision is **fixed in code, not chosen by the caller**: the pi0/pi05 family routes its embedding path through two module-level `_preferred_dtype()` hooks that return bfloat16 unconditionally, so `policy.to(torch.float32)` alone does not produce a float32 forward.
- It is a **local** statement about the action expert's posterior at one observation. It is blind to confident-but-wrong VLM-level errors (a cleanly misread instruction produces a low-curvature, confidently-wrong chunk) and to precision-sensitive failures, and it degrades on undertrained models. **It is a monitor, not a guarantee of correctness.**
- The detector is only free at *inference*; calibration needs ~50 successful rollouts per task, which at a 20-60% success rate is 80-250 episodes — several times a default `eval.n_episodes = 16` run.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,56 @@ The format is loosely based on [Keep a Changelog](https://keepachangelog.com/).
 
 ## [Unreleased]
 
+### Added — `accel`, a cost-free flow-matching uncertainty proxy — **opt-in, no config_version bump**
+
+Denoising acceleration (`accel`, [arXiv:2607.27933](https://arxiv.org/abs/2607.27933))
+read off the flow-matching samplers' existing Euler loop. A maximally *certain* CFM field
+is an affine-isotropic contraction, so its denoising trajectory is a straight line at
+constant velocity; the normalized total variation of that velocity over the first `p` of
+`T` Euler steps is a proxy for the endpoint posterior's spread. It needs **no extra
+network evaluations, no resampling, no training, and no auxiliary probe** — the whole cost
+is two vector norms per denoise step, on velocities the sampler already computed.
+
+**No `config_version` bump.** Nothing about what the model sees or produces changes.
+`accel` is off by default (`policy.accel_prefix is None`), and with it off every added
+line in a sampler is dead, so the traced/compiled graph is unchanged.
+
+- `opentau.policies.accel` — the estimator (`AccelMeter`), the action-dim mask recovered
+  from the normalization buffers, `AccelProvenance`, and the shared `configure_accel`
+  enable knob (`OPENTAU_ACCEL_PREFIX=auto`, or an explicit prefix `>= 2`).
+- `opentau.utils.accel_detector` — offline one-sided CUSUM + split-conformal calibration
+  over the per-chunk score stream, fitted on **successful** rollouts only.
+- `opentau-accel-diagnose` — measures the bfloat16 rounding floor, the run-to-run spread,
+  and the prefix profile of a checkpoint. **Run this first**: `accel`'s numerator is a
+  difference of nearly-equal vectors, so bf16 puts a positive-biased floor under the score
+  that compresses exactly the low-uncertainty end the method's premise depends on.
+
+Measured on one real 6-DoF SO101 pi05 checkpoint (`T = 10`, QUANTILE action norm, delta
+actions, `max_delay = 4`) over 8 real frames of its own training data: rounding floor
+**0.0056**, redrawn-noise spread **0.0094**, between-observation spread **0.097** — a
+signal-to-floor ratio of **17**, i.e. the score is measuring field geometry rather than
+bf16 arithmetic on that checkpoint. Re-measure per checkpoint; this is not a constant.
+The prefix sweep on the same run reproduces the paper's terminal singularity: step-over-step
+growth in `accel_p` decelerates monotonically (122% → 62% → 69% → 38% → 41% → 33% → 22%)
+and then *reverses* to +48% at the last step, which is why `default_prefix` returns `T - 1`
+and not `T`.
+
+Note that the serving precision is fixed in code rather than chosen by the caller: the
+pi0/pi05 family routes its embedding path through **two** module-level `_preferred_dtype()`
+hooks (one in `modeling_pi05`, one in `paligemma_with_expert`) that return bfloat16
+unconditionally, so `policy.to(torch.float32)` alone does not produce a float32 forward.
+- `opentau-accel-calibrate` — fits a threshold from an eval run's `eval_info.json`.
+- `eval.py` records a per-episode `accel` stream (done-truncated, paired with `success`);
+  the gRPC `ActionChunkResponse` gained an `optional float accel = 5` (explicit presence,
+  so unset is distinguishable from a maximally-confident `0.0`); the RoboCasa server sends
+  an `accel` envelope only to clients that set `"include_accel": true`.
+
+Known limits, all inherited from the method: it is a *local* statement about the action
+expert's posterior at one observation, blind to confident-but-wrong VLM-level errors and
+to precision-sensitive failures, and it degrades on undertrained models. It is a monitor,
+not a guarantee of correctness.
+
+
 ### Changed — openpi-faithful normalization (zero-range dims + epsilon) — **breaking, gated**
 
 `config_version` **0 → 1.** Two openpi-parity normalization changes are bundled
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -109,6 +109,8 @@ opentau-so101-record = "opentau.scripts.so101.record:main"
 opentau-so101-setup-motors = "opentau.scripts.so101.setup_motors:main"
 opentau-so101-find-port = "opentau.scripts.so101.find_port:main"
 opentau-so101-find-cameras = "opentau.scripts.so101.find_cameras:main"
+opentau-accel-diagnose = "opentau.scripts.diagnose_accel:main"
+opentau-accel-calibrate = "opentau.scripts.calibrate_accel:main"
 
 [project.optional-dependencies]
 dev = ["pre-commit>=3.7.0",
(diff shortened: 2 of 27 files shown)


shuheng-liu:
All four review findings are fixed in `1a19055` — no pushbacks, they were all real.

**gRPC misattribution race** (`grpc/server.py`) — correct and worth catching; the eval path warned about exactly this shape and the servicer did not. `sample_actions` and the `last_accel` read are now atomic under a lock that is *only* taken when accel is enabled, so the default serving path keeps its existing concurrency unchanged.

**Fail-open provenance comparison** (`calibrate_accel.py`) — this was the nastiest of the four, because the failure mode is a guard silently removing itself: an unlabelled calibration makes `detect()` skip the comparability check altogether. Now compares only `AccelProvenance.COMPARABLE_FIELDS`, and returns the shared provenance with the per-sample fields cleared rather than carrying one arbitrary task’s values on a calibration fitted over all of them. Verified both directions: tasks differing only in `dataset_index`/`num_scored_dims` now yield a labelled calibration; a genuine `num_steps` difference still refuses the label.

**Missing eval coverage** — fair, and it was the one part of the change with no committed test. Added `TestRolloutAccelCollection` and `TestEvalPolicyAccelPairing` to `tests/scripts/test_eval.py`: per-env done-truncation, per-env column identity, provenance recorded once, the width-mismatch drop, the no-accel no-op, and the stream-to-episode pairing. Both guards are mutation-verified — removing the `chunk_step <= done_index` truncation and removing the width check each make exactly the corresponding test fail.

**Stale docstring** — correct, and self-inflicted: this PR added `PreTrainedConfig.accel_prefix` and the docstring still said no config field existed. Updated.

One note on the review’s own framing: the two suspicions it chased and dismissed (`next(policy.parameters()).dtype` resolving to a stat param, and the `diagnose_accel` dtype restore) were the right things to be suspicious about — the SigLIP pinning makes "the policy dtype" genuinely ambiguous in this codebase.

shuheng-liu:
Fixed in `4d0678b` — the new suggestion was right, and it is the same class as the eval-coverage finding from the previous round: I had verified the provenance fix in a throwaway script and never committed the check, so a revert to whole-dict comparison would have failed nothing.

`tests/scripts/test_calibrate_accel.py` (18 tests) now covers the driver, and pins the guard in **both** directions — because the fix could degrade either way:

- tasks differing only in the per-sample `dataset_index`/`num_scored_dims` must still share a label (the fail-open the commit fixed), and the shared label must have those fields cleared;
- a genuine disagreement on any `COMPARABLE_FIELDS` entry must still drop the label — otherwise the fix could quietly become "never disagree", restoring fail-open by another route.

Both mutations verified: reverting to whole-dict comparison fails exactly the two per-sample tests, and forcing the disagreement branch off fails the parametrized genuine-disagreement tests. The file restores byte-identically after each.

Also covered while there: outcome pairing (`accels[i]` ↔ `successes[i]`, and the length-mismatch drop), the too-few-successes refusal, and `load_eval_info` input handling.

On the review caveat about not being able to execute the tests — noted, and reasonable. For what it is worth these were run locally (384 passed across `tests/scripts` + the accel suites) and `cpu_test.yml` is the executable check on this commit.

PR #513, opened 2026-08-12
test(policies): replace unconditional GPU skips with VRAM-gated ones

shuheng-liu:
## What this does

Seven pi06 / pi07 integration tests carried
`@pytest.mark.skip(reason="Requires too much memory, does not fit on RTX 3090 24GB")`.
Because the skip was unconditional they ran **nowhere** — not on CI, not on a dev
box, not on a 80 GB A100 — and two of them had quietly rotted. This makes the
skip conditional on available VRAM and repairs the rot.

**Tests-only change.** `git diff --stat -- src/` is empty; no production
behaviour changes.

### The two rotted tests

Both failed identically on a pristine `main` once the marker was stripped, so
this is decay, not a regression.

1. **`test_pi07_low_level.py::...::test_complete_pi07_low_level_pipeline`** —
   `einops.EinopsError` in `prepare_videos`, 6-D video `[1, 2, 2, 3, 448, 448]`.
   `select_action` takes **one** observation per step and builds the temporal
   window itself (`_build_history_batch` buffers frames and emits
   `obs_history_is_pad`). The test handed it the already-stacked *training*
   tensors, so they got stacked a second time. The sibling pi07_paligemma test
   already does this correctly. Fixed the fixture to pass 4-D cameras / 2-D
   state.

2. **`test_pi07_paligemma_low_level.py::...::test_complete_pi07_low_level_pipeline`** —
   `TypeError: ~ (operator.invert) is only implemented on integer and
   Boolean-type tensors` on `frame_keep = ~obs_history_is_pad`. The test cast
   the whole batch to bfloat16, turning the bool mask into a float; only
   `action_is_pad` was restored by hand.

   **This is not a latent production bug.** I traced every source of
   `obs_history_is_pad`: `datasets/lerobot_dataset.py` builds it bool, both
   `_build_history_batch` implementations build it bool, and
   `scripts/benchmark_inference.py` builds it bool. No entry point casts batch
   tensors wholesale — `policy.to(bfloat16)` never touches the batch. So no real
   caller can reach the `~` with a non-bool, and the fix belongs in the fixture.
   The hand-maintained "restore this one key to bool" list is replaced with a
   `to_cuda_bf16` helper that preserves bool dtype by construction.

Fixing (2) revealed three further stale expectations in the same test, all
fixture-side and all confirmed against deliberate production behaviour:

- `NUM_SUBGOAL_CAMERAS = 1` → `NUM_CAMERAS`. `prepare_subgoal_images`
  deliberately pads to one slot per camera so prefix length cannot depend on
  which `subgoal{k}` keys a batch carries; the test now supplies `subgoal1` too,
  matching the file's own `_subgoal_batch` helper.
- `_verify_position_ids` / `_verify_action_expert_attention_mask` subtracted only
  the discrete-action span. `num_cross_att_tokens` drops every
  `exclude_from_cross_attention` item, which is the trailing `"Action: "`
  indicator **and** the span. The pi07 sibling test already had this right.
- `INFER_STATE_TOKENS = 1` → `N_OBS_STEPS`. It predates `_build_history_batch`,
  which restores the full `T` window at inference.

### Conditional skips

All seven sites now use `require_vram_gib(...)` from `tests/utils.py`, following
that file's existing `require_cuda` decorator convention. Floors are per-file and
set from measured peaks rather than one blanket number — the pi07_low_level tests
peak at 0.71 GiB, so a single high floor would have left them nearly as dead as
before:

| file | measured peak (fresh process) | floor |
|---|---|---|
| `test_pi07_low_level.py` | 0.71 GiB | 4 GiB |
| `test_pi07_paligemma_low_level.py` | 13.9 GiB | 18 GiB |
| `test_pi07_high_level_planner.py` | 17.3 GiB | 24 GiB |
| `test_pi06.py` | 19.6 GiB | 24 GiB |

A call-time decorator rather than `pytest.mark.skipif`: the condition needs
`get_device_properties`, which initializes a CUDA context, and the gating CPU run
imports every test module in all `-n auto` workers — on a GPU host that would
spawn one context per worker for tests that get deselected anyway.

The heaviest two files are floored above a 24 GB card. Their peaks (19.6 / 17.3
GiB) would probably fit, but I could not verify that they clear a 24 GB budget
*after* the rest of the suite has run, because no uncontended >24 GB card was
available. The floors are deliberately conservative, and each comment records the
measured peak and the value to lower it to once someone can confirm on a quiet
24 GB card.

### Cross-test memory accumulation

Running the four files in one process OOM'd even though no single test exceeds
19.6 GiB. PyTorch drops its cached blocks on OOM but never runs a Python GC, so
the reference cycles in each finished test's `nn.Module` graph stay resident and
peaks sum instead of overlapping. Added a `release_cuda_memory_after_gpu_test`
autouse fixture in `tests/conftest.py` (marker-gated, so it is a no-op for the
~2600 CPU tests). With it, all 12 pass in one process. This is most likely what
the original "does not fit on 24GB" note was actually about.

Effect on the GPU suite's ceiling on a 24 GB card: the heaviest **pre-existing**
GPU test peaks at 11.6 GiB; the heaviest test this PR newly enables there peaks
at 13.9 GiB. So the suite's requirement rises by roughly 2 GiB, well inside
budget. The two files floored at 24 GiB do not run there at all, leaving CI's
memory profile for them unchanged.

### Coverage gap closed

`prepare_subgoal_images`' padding loop exists for the *mixed* present/missing
subgoal-key case, but only the all-present and no-key cases were tested — both
stay correct if the loop is deleted. Added
`test_partial_subgoal_keys_pad_out_to_one_slot_per_camera`. Verified it is a real
pin: deleting the padding loop fails this test and no other in the class.

## How it was tested

- **GPU, targeted** — `pytest -m gpu -n 0` over the four touched files on a 32 GB
  dev GPU: **12 passed**, including all seven previously-dead tests. Also
  confirmed they pass together in a single process on a ~24 GiB budget, which is
  what exposed the accumulation issue above.
- **Skip machinery** — with the device reporting 22 GiB, the three tests floored
  at 24 GiB skip with `requires >= 24.0 GiB VRAM (device has 22.0 GiB)` and the
  rest still run.
- **Per-test VRAM peaks** measured with a temporary plugin (not committed) to set
  the floors from data rather than guesswork; the pre-existing GPU suite's own
  ceiling (11.6 GiB) was measured the same way.
- **Mutation check** on the new subgoal test: with the padding loop removed it
  fails and the other 15 tests in the class pass; restored via `git checkout`.
- **CPU** — `pytest -m "not gpu and not network" -n auto`: 2593 passed. Two
  failures are pre-existing and unrelated — `test_libero2torch` fails identically
  on a pristine `main` (local LIBERO asset paths), and `test_exception_handling`
  is a timing test that passes in isolation.
- **Pre-commit** — clean over all files.

Regression tests were not run; this PR changes no policy code.

```bash
git fetch origin claude/confident-allen-c1fdb8 && git checkout claude/confident-allen-c1fdb8
```

On a CUDA box, the tests that used to be dead everywhere:

```bash
pytest -m gpu -n 0 tests/policies/test_pi07_low_level.py tests/policies/test_pi07_paligemma_low_level.py tests/policies/test_pi07_high_level_planner.py tests/policies/test_pi06.py
```

The new CPU-only subgoal test:

```bash
pytest -sx tests/policies/test_pi07_paligemma_low_level.py::TestPI07PaligemmaLowLevelSubgoalEmbedding
```

To see the skip reason on a card below the floor, or to confirm nothing is
silently dead any more:

```bash
pytest -m gpu -n 0 -rs tests/policies/test_pi06.py
```

## Checklist

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -15,9 +15,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import gc
 import os
 
 import pytest
+import torch
 from huggingface_hub import HfApi
 
 from opentau.datasets.standard_data_format_mapping import DATA_FEATURES_NAME_MAPPING
@@ -75,6 +77,30 @@ def patch_builtins_input(monkeypatch):
             print(text)
 
     monkeypatch.setattr("builtins.input", print_text)
+
+
+@pytest.fixture(autouse=True)
+def release_cuda_memory_after_gpu_test(request):
+    """Drop a finished GPU test's CUDA memory before the next one allocates.
+
+    `pytest -m gpu -n 0` runs every GPU test in one process. PyTorch frees its
+    own cached blocks when an allocation fails, but it never runs a Python GC,
+    so the reference cycles inside a just-finished test's `nn.Module` graph keep
+    whole models resident. Peaks that should overlap add up instead. Measured:
+    the four pi06/pi07 policy files OOM on a 24 GiB budget without this sweep
+    and pass with it, though no single test in them exceeds 19.6 GiB.
+
+    The marker check keeps this off the ~2600 CPU tests, where it would be pure
+    overhead. On a failing test pytest still holds the frame for the traceback,
+    so the sweep only reliably reclaims after a pass — which is the case that
+    matters for the tests that follow.
+    """
+    yield
+    if request.node.get_closest_marker("gpu") is None:
+        return
+    gc.collect()
+    if torch.cuda.is_available():
+        torch.cuda.empty_cache()
 
 
 @pytest.fixture(scope="session")
--- a/tests/policies/test_pi06.py
+++ b/tests/policies/test_pi06.py
@@ -44,6 +44,17 @@ from opentau.policies.pi06.modeling_pi06 import (
     resize_with_pad,
 )
 from opentau.policies.utils import PerSampleLoss, ce_per_sample
+from tests.utils import require_vram_gib
+
+# VRAM floor for the end-to-end tests below, which build the full Gemma 3 4B
+# backbone on CUDA. Measured peak in a fresh process on an RTX 5090: 19.6 GiB
+# reserved — by far the heaviest tests in the GPU suite (everything else tops
+# out at 11.6 GiB). The floor is deliberately set above a 24 GB card rather
+# than just above the peak: these run late in `pytest -m gpu -n 0`, and whether
+# 19.6 GiB plus the suite's residue clears a 24 GB card could not be verified
+# (the only >24 GB card available was under contention from other workloads).
+# Lower it to ~21 once someone can confirm on a quiet 24 GB card.
+MIN_VRAM_GIB = 24.0
 
 # Block-causal attention mask (pi05 / π0.6 prefix-LM pattern)
 
@@ -956,7 +967,7 @@ class TestPi06GradCkptEquivalence:
 # End-to-end integration — guarded because Gemma 3 4B is huge.
 
 
-@pytest.mark.skip(reason="Requires too much memory, does not fit on RTX 3090 24GB")
+@require_vram_gib(MIN_VRAM_GIB)
 @pytest.mark.gpu
 @pytest.mark.slow
 def test_complete_pi06_pipeline_integration_smoke(lerobot_dataset_metadata):
@@ -1028,7 +1039,7 @@ def test_complete_pi06_pipeline_integration_smoke(lerobot_dataset_metadata):
     assert all(v.isfinite() for v in loss.values())
 
 
-@pytest.mark.skip(reason="Requires too much memory, does not fit on RTX 3090 24GB")
+@require_vram_gib(MIN_VRAM_GIB)
 @pytest.mark.gpu
 @pytest.mark.slow
 def test_pi06_loc_tokens_extend_vocab_and_resize_embeddings(lerobot_dataset_metadata):
--- a/tests/policies/test_pi07_high_level_planner.py
+++ b/tests/policies/test_pi07_high_level_planner.py
@@ -23,6 +23,17 @@ from opentau.policies.pi07.high_level_planner.modeling_pi07_high_level import (
     PI07HighLevelPlannerPolicy,
     make_att_2d_masks,
 )
+from tests.utils import require_vram_gib
+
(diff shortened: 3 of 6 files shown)


shuheng-liu:
Taken in `5ec6c22` — the suggestion was right, and checking it against the dataloader made it more than a docstring fix.

`prepare_metadata`'s inputs come off the dataset as `torch.long` (`speed`, `quality`) and `torch.bool` (`mistake`) — see `datasets/lerobot_dataset.py:1223`. So the blanket cast wasn't merely harmless-for-now: it was already changing what the model sees, stringifying `Speed: 50.0` where production produces `Speed: 50`. Keying the cast on `is_floating_point()` fixes that too, and keeps a future token-id tensor safe from the same rounding. Docstring now matches the implementation.

Verified on CPU that the metadata block is unchanged in the ways these tests assert on — with ints preserved vs. the old blanket cast, `prepare_metadata` yields the same padded width (52), the same real-token count (24), and a contiguous mask either way — so the prefix-layout assertions are unaffected. A GPU re-run of the four touched files is still pending; I'll confirm it before taking this out of draft.

shuheng-liu:
### GPU verification status before merge

Recording this plainly so it isn't lost in the thread.

**Verified on a 32 GB dev GPU at `30e4e4f`:** `pytest -m gpu -n 0` over the four touched files — **12 passed**, including all seven tests that this PR takes off the unconditional skip. Also confirmed they pass together in a single process on a ~24 GiB budget, which is what surfaced the cross-test memory accumulation the conftest fixture now handles.

**Not re-run at `5ec6c22`.** That commit is the review fix and touches only `tests/utils.py` (`git diff 30e4e4f..5ec6c22` is confined to it) — the bfloat16 cast is now keyed on `is_floating_point()` so integer fields keep their dtype. The GPU host became unreachable before it could be re-run.

What was checked on CPU instead: the only observable consequence of that dtype change is what `prepare_metadata` stringifies (`Speed: 50` vs `Speed: 50.0`). With ints preserved versus the old blanket cast, it yields the same padded width (52), the same real-token count (24), and a contiguous mask either way — so the prefix-layout assertions the GPU tests make are unaffected.

That is good evidence but it is not the GPU run. The nightly `gpu_test.yml` is the confirmation; if it reddens on any of these four files, this commit is the first place to look.

PR #520, opened 2026-08-17
feat(accel): measure the best prefix instead of assuming the paper's

shuheng-liu:
## What this does

`accel`'s prefix `p` was chosen by inheritance, not by measurement. `default_prefix` returns `T - 1` because that is what the paper's online detector used, and the shipped diagnostic's prefix sweep *looked* like it confirmed that — it reports mean `accel_p` per `p`, and the previous CHANGELOG entry read its shape as reproducing the paper's terminal singularity.

**That inference does not hold.** Both of `accel_p`'s sums accumulate over the prefix, so its magnitude — and the shape of its growth — rise with `p` whether or not the extra Euler steps carry any posterior information. The sweep would produce a similar curve on data with no signal at all, so its peak cannot select a prefix. It is descriptive, and this PR stops it from being read as more than that.

The criterion that *can* select a prefix is the rank correlation between `accel_p` and an independent measure of posterior spread. That reference has to be bought the expensive way — `K` resampled denoises of one observation, whose disagreement *is* the posterior width — which is why it lives in an offline diagnostic and nowhere near the serving path.

**Measured on a real 6-DoF SO101 pi05 checkpoint** (`T = 10`, QUANTILE action norm, delta actions, 24 real frames of its own training data at stride, `K = 32`):

| p | 2 | 3 | 4 | 5 | 6 | 7 | 8 | **9** | 10 |
|---|---|---|---|---|---|---|---|---|---|
| **rho** vs resampled spread | +0.703 | +0.743 | +0.752 | +0.728 | +0.770 | +0.763 | +0.814 | **+0.833** | +0.812 |
| mean `accel_p` | 0.009 | 0.017 | 0.023 | 0.034 | 0.044 | 0.061 | 0.088 | 0.131 | **0.228** |

`rho` peaks at `p = T - 1` and falls at `p = T`. So `default_prefix` is right on this checkpoint and the terminal step really is the weaker one — the same conclusion as before, now resting on evidence that supports it. The mean sweep's argmax is `p = 10`, which `rho` ranks *below* `p = 9`.

Two things worth carrying forward: every prefix scored +0.70…+0.83, so prefix choice is not very consequential here (which also means it will not confound a downstream detection result); and the paper's mid-schedule optimum (`p/T ~ 0.4-0.5`, i.e. `p = 4-5`, `rho` +0.75) does **not** transfer. A `T = 5` schedule (pi06/pi07) is unexamined, and there the two candidate answers are far apart.

### Supporting changes

None of these alter a served score.

- **`AccelMeter` gained an opt-in per-step trace and `value_at(p)`.** `accel_p` is a prefix statistic of one velocity sequence, so every prefix is recoverable from **one** denoise pass rather than `T - 1` of them. That is what makes the resample study affordable, and it makes the descriptive sweep `T - 1` times cheaper as a side effect. Off in serving, where the meter allocates exactly what it did before.
- **`record_traces()`** hands the meters built inside it back to a diagnostic, so no policy has to expose its internals or widen a signature. A `ContextVar` rather than a module global: the gRPC server calls `sample_actions` from several threads, and a global would splice one thread's diagnostic run into another's serving traffic.
- **`action_dim_scale()`** converts a raw-unit spread into the normalized units `accel` is computed in, sharing one buffer extractor with `resolve_action_dim_mask` so the two cannot drift apart when a normalization mode is added.
- **Diagnostic observations are drawn across every dataset in the mixture**, not just the first. The correlation is taken *across* observations, so their variety is the study's resolution; silently sampling one member of a mixture narrowed that without saying so. Default count raised to 24 — at 8 points Spearman needs |rho| >= 0.74 to clear p < 0.05, too blunt to separate neighbouring prefixes.
- **The float32 leg of the dtype-floor measurement is skippable** (`OPENTAU_ACCEL_MEASURE_DTYPE_FLOOR=0`). It holds a second copy of the weights and is the first thing to fail on a GPU shared with other work; the floor it measures is fixed per checkpoint. Skipping reports `NaN` rather than a fabricated ratio.

## How it was tested

**New: `tests/scripts/test_diagnose_accel.py` (18 tests).** The central one is `test_the_study_follows_rank_correlation_not_the_rising_mean`: it scripts a policy where the two criteria give *different* answers — early curvature tracks the true posterior spread, late curvature is built to fight it — and pins that the study follows the correlation. It also asserts the mean sweep peaks at the *worst* prefix, because without that manufactured disagreement the test would pass under either implementation.

**Mutation-tested both load-bearing pins**, since a pin whose passing does not depend on the property it names is not a pin:

| mutation | result |
|---|---|
| selection criterion swapped from `rho` back to the mean sweep | 2 failed, incl. the central test |
| trace snapshot moved before the numerator update (off-by-one) | 2 failed, incl. the bit-identity test |

Both restored byte-identically afterwards.

`value_at(p)` is asserted **bit-identical** (`torch.equal`, not a tolerance) to a meter independently constructed at prefix `p`, for every `p` — a tolerance-based check would wave through exactly the off-by-one above.

Also pinned: the delta-action invariance the study relies on (a constant offset across resamples cancels, so no delta inverse is needed), per-dim standardization undoing a unit change, mask exclusion of padded dims and rows past the executed window, and `record_traces()` restoring the serving configuration on exception.

**Full CPU suite** for the affected trees: `pytest tests/policies/ tests/scripts/ tests/utils/ -m "not gpu and not network" -n auto` — 1627 passed. The one failure, `tests/utils/test_libero_utils.py::test_libero2torch`, is a pre-existing local-environment issue (a stale `LIBERO_CONFIG_PATH`); I confirmed it fails identically on a pristine `origin/main` checkout, so it is not from this change.

**On the GPU checkbox, so the reviewer is not misled:** the `-m "gpu"` suite was **not** run ad hoc for this PR — GPU coverage is deliberately delegated to the nightly `gpu_test.yml` and regression workflows. The policy box is checked because `AccelMeter.update` executes inside the sampler's Euler loop, which is the conservative reading even though no `modeling_*.py`, `optim/`, sampler, or `train.py` file is touched. What *was* exercised on GPU is the end-to-end run below, plus a `pi07_paligemma` eval that loaded the policy and armed the estimator (`accel enabled: prefix=9 of num_steps=10`) before I stopped it.

No determinism run: CLAUDE.md rule 3 covers `train.py`, `modeling_*.py`, `optim/`, and the samplers, none of which this touches, and `accel` is inference-only (never reached from `forward`).

**End-to-end on the real checkpoint**, which is where the numbers in the table come from — 24 real training frames, 768 denoise passes, on the GPU dev box.

```bash
pytest -sx tests/scripts/test_diagnose_accel.py
```

The claim that the mean sweep cannot select a prefix is worth checking directly — this is the test that fails if the criterion is swapped back:

```bash
pytest -sx tests/scripts/test_diagnose_accel.py::test_the_study_follows_rank_correlation_not_the_rising_mean
```

On a GPU box with a flow-matching checkpoint, the study itself. It is inference-only and starts no training:

```bash
OPENTAU_ACCEL_OBSERVATIONS=24 OPENTAU_ACCEL_RESAMPLES=32 \
python src/opentau/scripts/diagnose_accel.py \
    --config_path=<a config carrying the checkpoint's training data> \
    --policy.path=<repo>@<step> --policy.device=cuda
```

Set `OPENTAU_ACCEL_RESAMPLES=0` to skip the study (it is the only expensive part), or `OPENTAU_ACCEL_MEASURE_DTYPE_FLOOR=0` if the GPU is shared.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,48 @@ The format is loosely based on [Keep a Changelog](https://keepachangelog.com/).
 
 ## [Unreleased]
 
-_Nothing yet._
+### Changed — `accel`'s prefix is now *measured* rather than assumed
+
+`default_prefix` returns `T - 1` because that is the prefix the paper's online detector
+used. The 0.13.0 notes below cite the prefix sweep as confirming it, via the step-over-step
+growth in `accel_p` reversing at the last step. **That inference does not hold**, and this
+entry supersedes it: both of `accel_p`'s sums accumulate over the prefix, so its magnitude
+— and the shape of its growth — rise with `p` whether or not the extra Euler steps carry
+posterior information. The sweep is descriptive; it would produce a similar curve on data
+with no signal at all, so its peak cannot select a prefix.
+
+The criterion that *can* is the rank correlation between `accel_p` and an independent
+measure of posterior spread, and that reference has to be bought the expensive way — `K`
+resampled denoises of one observation, whose disagreement is the posterior width.
+`opentau-accel-diagnose` now measures it (`OPENTAU_ACCEL_OBSERVATIONS`, default 24;
+`OPENTAU_ACCEL_RESAMPLES`, default 32; `0` skips the study) and reports `rho` per prefix
+with a recommendation.
+
+Re-measured on the same 6-DoF SO101 pi05 checkpoint (`T = 10`, 24 real frames, `K = 32`):
+`rho` climbs to **+0.833 at `p = 9`** and falls to **+0.812 at `p = 10`**. So `T - 1` is
+right on this checkpoint and the terminal step really is the weaker one — the same
+conclusion as before, now resting on evidence that supports it. Every prefix scored
++0.70…+0.83, so prefix choice is not very consequential here; the paper's mid-schedule
+optimum (`p/T ~ 0.4-0.5`, i.e. `p = 4-5`, `rho` +0.75) does **not** transfer. Re-measure
+per checkpoint — a `T = 5` schedule (pi06/pi07) is unexamined, and there `T - 1` and the
+paper's mid-schedule choice are far apart.
+
+Supporting changes, none of which alter a served score:
+
+- `AccelMeter` gained an **opt-in** per-step trace and `value_at(p)`, so every prefix is
+  recoverable from **one** denoise pass instead of `T - 1` — that is what makes the study
+  affordable, and it makes the descriptive sweep `T - 1` times cheaper too. Off in serving.
+- `record_traces()` hands the meters built inside it back to a diagnostic, so no policy has
+  to expose its internals. A `ContextVar`, not a global: the gRPC server samples from
+  several threads, and a global would splice a diagnostic into serving traffic.
+- `action_dim_scale()` converts a raw-unit spread into the normalized units `accel` is
+  computed in, sharing one buffer extractor with `resolve_action_dim_mask`.
+- Diagnostic observations are drawn across **every** dataset in the configured mixture
+  rather than only the first — the correlation is taken across observations, so their
+  variety is the study's resolution.
+- The float32 leg of the dtype-floor measurement can be skipped
+  (`OPENTAU_ACCEL_MEASURE_DTYPE_FLOOR=0`). It holds a second copy of the weights and is the
+  first thing to fail on a shared GPU; skipping reports `NaN` rather than a fabricated ratio.
 
 ## [0.13.0] - 2026-08-17
 
--- a/src/opentau/policies/accel.py
+++ b/src/opentau/policies/accel.py
@@ -350,11 +350,10 @@ def _fit_dim_width(per_head: Tensor, max_action_dim: int, pad_value: float | boo
     return per_head[:, :max_action_dim] if width > max_action_dim else per_head
 
 
-def _gather_heads(per_head: Tensor, dataset_index: Tensor) -> Tensor:
-    """Select one ``(dim,)`` row of a per-head quantity for each sample."""
+def _head_index(dataset_index: Tensor, per_head: Tensor) -> Tensor:
+    """Return the clamped ``(B,)`` row index selecting a norm head per sample."""
     index = dataset_index.to(device=per_head.device, dtype=torch.long).reshape(-1)
-    index = index.clamp_(0, per_head.shape[0] - 1)
-    return per_head.index_select(0, index)
+    return index.clamp_(0, per_head.shape[0] - 1)
 
 
 def action_dim_scale(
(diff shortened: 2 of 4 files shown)


shuheng-liu:
Both findings confirmed and fixed in eeb5276 — thanks, the first one is a real miss on my part.

**`diagnose_accel.py:667` (blocking) — fixed.** Confirmed exactly as described: NaN is truthy and `np.isfinite(NaN)` is False, so the guard fell through to `inf`, and every `<` against `inf` is False so the thresholds landed on "OK". Worth stating plainly — my own run on a shared GPU printed `dtype floor nan / signal / floor inf / OK — real signal dominates the rounding floor`, and I read past it. The docstring and CHANGELOG claim were describing intended behaviour, not actual.

`_signal_to_floor` now distinguishes the two cases that were being collapsed:

- floor **never measured** (leg skipped, or no finite pair) → `NaN`
- floor **measured at exactly zero** (float32 and the serving dtype agreed on every observation) → `inf`, which is a real infinite ratio and should still read "OK"

`format_report` gets an explicit NaN branch *ahead* of the thresholds, printing `NOT MEASURED` plus how to obtain the missing measurement. `test_a_floor_measured_at_exactly_zero_is_still_infinite` guards the second case so the correction doesn't overreach.

**`diagnose_accel.py:841` (suggestion) — fixed, both directions.** Also confirmed: `[100, 0]` with `count=24` drew 12, and `count=2` over 5 datasets drew 4. Agreed this matters more than a typical allocation nit, for the reason you gave — `count` is the study's sample size, which is the thing this PR argues decides whether neighbouring prefixes are separable.

One wrinkle worth flagging: capping by the running remainder at every position is *not* sufficient on its own. A single pass can only push a shortfall forward, so a mixture whose **last** member is the short one (`[100, 5]`, `count=24`) still comes up short — the datasets with spare frames have already been offered their slice by then. `_allocate_draws` therefore re-passes over the members that still have capacity until the remainder is placed or no progress is possible. The caller warns when the final tally differs from `count`.

Verified with a property check over 20,000 random `(sizes, count)` combinations: `sum(quotas) == min(count, capacity)` and no dataset is ever over-drawn.

**Mutation-checked**, since a regression test that passes either way isn't one:

| mutation | fails |
|---|---|
| restore the original `signal_to_floor` expression | `test_an_unmeasured_floor_is_not_an_infinite_ratio` |
| drop the NaN branch from `format_report` | `test_the_report_refuses_to_certify_a_floor_it_never_measured` |
| revert to the single-pass allocator | `test_a_short_dataset_is_made_up_by_the_others` |

Nine new tests; suite is 29 in `test_diagnose_accel.py`, 188 across the accel files.

PR #523, opened 2026-08-18
fix(grpc): run compiled sample_actions on one dedicated thread

cenxiaoxinxin-dotcom:
## What this does
Fixes gRPC `GetActionChunk` crashing with an empty `Inference error:` after `torch.compile(mode="max-autotune")`. CUDA graphs are thread-local; warmup ran on the main thread while RPCs ran on the gRPC pool. Route warmup and every `sample_actions` through one dedicated infer thread, and include the exception type in INTERNAL errors.

## How it was tested
- Reproduced on tuner-dev: HealthCheck succeeded, `GetActionChunk` returned `INTERNAL: Inference error:` (empty `AssertionError` from cudagraph TLS).
- Added `TestInferenceThread` in `tests/scripts/test_grpc_planner_server.py`, pinning both halves of the new contract: `sample_actions` runs on the servicer's own infer worker, and consecutive requests land on that same thread. The assertions compare thread *objects*, not names or idents — a fresh per-call pool reuses the `policy-infer` name, and CPython recycles idents of dead threads, so neither would fail on a regression.
- Mutation-tested those two tests locally: inlining the call (back to the gRPC handler thread) and rebuilding the executor per call each make both tests fail; the current implementation passes.
- `pytest -m "not gpu and not network" tests/scripts/ -n auto` → 344 passed. `pre-commit run --all-files` clean.
- The real `torch.compile` + cudagraph path needs a CUDA box and was not re-run here; the CPU tests pin the threading contract only.

```bash
pytest -sx tests/scripts/test_grpc_planner_server.py::TestInferenceThread
```

On a CUDA box, start the gRPC server, then HealthCheck followed by GetActionChunk on a thread-pool RPC (not the process that constructed the servicer).

## Checklist

### DIFF
--- a/src/opentau/scripts/grpc/server.py
+++ b/src/opentau/scripts/grpc/server.py
@@ -56,7 +56,7 @@ import traceback
 from concurrent import futures
 from dataclasses import asdict
 from pprint import pformat
-from typing import Callable, Iterator
+from typing import Any, Callable, Iterator
 
 import numpy as np
 import torch
@@ -132,9 +132,7 @@ class RobotPolicyServicer(robot_inference_pb2_grpc.RobotPolicyServiceServicer):
         # every later sample_actions must share one dedicated thread — otherwise
         # the first real request dies in cudagraph_trees with a bare
         # AssertionError (empty str → "Inference error: ").
-        self._infer_executor = futures.ThreadPoolExecutor(
-            max_workers=1, thread_name_prefix="policy-infer"
-        )
+        self._infer_executor = futures.ThreadPoolExecutor(max_workers=1, thread_name_prefix="policy-infer")
 
         logger.info(f"Initializing policy on device: {self.device}")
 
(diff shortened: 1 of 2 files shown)


cenxiaoxinxin-dotcom:
Addressed all five findings in 16f689f. Two of them deliberately deviate from the suggested implementation — rationale below so it is on the record.

**Test pinning the dedicated-thread contract (suggestion 3).** The suggested `assert seen_threads == ["policy-infer_0"]` does pin the thread *name*, but I mutation-tested it and it does not kill the whole contract: replacing `self._infer_executor.submit(...)` with a fresh `ThreadPoolExecutor(max_workers=1, thread_name_prefix="policy-infer")` per call reproduces the same name, so the "one shared thread" half still passes. Falling back to `threading.get_ident()` does not help either — CPython recycles idents of dead threads, and that mutation kept passing too. `TestInferenceThread` therefore records thread *objects* and compares identity against the thread the servicer's own executor runs work on:

```python
@staticmethod
def _infer_worker(servicer) -> threading.Thread:
    return servicer._infer_executor.submit(threading.current_thread).result()
```

Both mutations now fail both tests (verified locally): inlining `_run()` puts sampling on `MainThread`, and a per-call pool yields a different object than the servicer's worker.

**`_sample_actions_sync` typing (nit 5).** Annotated `batch: dict[str, Any]` rather than the suggested `dict[str, torch.Tensor]`: `_prepare_observation` also puts `prompt` (a `list[str]`) and the `robot_type` / `control_mode` / `dataset_repo_id` norm-head strings into the batch, so `dict[str, torch.Tensor]` would be a contract the implementation does not honor (CLAUDE.md rule 8). `_prepare_observation`'s own return annotation has the same inaccuracy; left untouched here since it predates this PR and is unrelated to the fix.

The other three are taken as suggested: ruff-format applied (`pre-commit run --all-files` clean), the stale concurrency comment now scopes the lock to sample + `last_accel` atomicity and credits the executor with serializing the sampling, and the error detail uses the lossless `f"{type(e).__name__}: {e}" if str(e) else type(e).__name__`.

Verification: `pre-commit run --files src/opentau/scripts/grpc/server.py tests/scripts/test_grpc_planner_server.py` all green, `pytest -m "not gpu and not network" tests/scripts/ -n auto` → 344 passed. The real `torch.compile` + cudagraph path is still only reproducible on a CUDA box; no GPU here, so `-m gpu` was not run.

shuheng-liu (approved):

PR #525, opened 2026-08-18
feat(policies): best-of-N action-chunk sampling for pi05 and pi06

shuheng-liu:
## What this does

Adds **best-of-N action-chunk sampling** to pi05 and pi06 (🗃️ Feature).

A flow-matching policy maps one Gaussian draw to exactly one action chunk, deterministically. So drawing `n_candidates` noises yields `n_candidates` distinct chunks, and an `ActionChunkCritic` picks the one that reaches the robot.

**The candidates share a single VLM prefix pass.** The prefix KV cache is never written again during denoising (`fill_kv_cache=False` in every `denoise_step`), so expanding it across candidates lets one prefix pass serve all N. Only the action expert's Euler loop widens — and at batch 1 that loop is occupancy-bound rather than FLOP-bound, so widening it rides largely free.

Measured on an RTX 5090 (bf16, sdpa, 3 cameras, 1024 prefix tokens, chunk 50, 10 Euler steps, batch 1), fused against naive replication of the observation across the batch — one prefix pass per candidate, same result:

| N | fused | naive | speedup | overhead vs own N=1 | fused peak | naive peak |
|---|---|---|---|---|---|---|
| 1 | 152.9 ms | 154.3 ms | 1.01x | — | 6.41 GiB | 6.42 GiB |
| 2 | 153.2 | 187.7 | 1.23x | +0.1% | 6.41 | 6.56 |
| 4 | 154.8 | 256.1 | 1.65x | +1.2% | 6.41 | 6.84 |
| 8 | 155.5 | 392.1 | 2.52x | +1.7% | 6.45 | 7.41 |
| 16 | 191.2 | 698.2 | 3.65x | +25% | 6.62 | 8.54 |
| 32 | 277.8 | OOM | — | +82% | 6.95 | — |

**Eight candidates cost under 2% of wall clock.** The knee is between 8 and 16. Peak memory is nearly flat because the only per-candidate allocation is the prefix KV cache at 18 MiB (MQA: one KV head, head_dim 256, 18 layers); naive replication OOMs at N=32 where the fused path does not.

### The default path is unchanged

`n_candidates` defaults to `1`, and a policy wrapper is constructed with `1` regardless of what its config says — `configure_candidates` is the only writer, so a config travelling with a checkpoint cannot self-arm best-of-N in a script that never opted in. Same posture as `accel_prefix`.

At N=1 no critic is loaded, no candidate code runs, and the sampler is the one that shipped before. Verified at full size on GPU: N=1 after the change reproduces the pre-change output digest bit-for-bit (`5a0bf2ed20abbcc2`), from an identically-seeded model (weights hash `144933185b9ff63f`, same venv, same worktree, only the source differing).

**No `config_version` bump.** The field gates nothing about training and does not change what a given noise draw decodes to; `config_version` gates normalization conventions only.

### Scope and the things it deliberately refuses

- **Only pi05 and pi06 are wired.** Every other family — pi0, pi05_mem, both pi07 low levels, cosmos3, cosmos3_nano, the two high-level planners and `value` — raises on `n_candidates > 1` rather than accepting the field and ignoring it, which is the failure an operator would read as "the critic never fires".
- **Serving-only.** The serving entry points arm it; `scripts/eval.py` refuses, rather than multiplying its default batch of 16 by N into a memory profile nobody has measured. Every script that builds a policy from `cfg.policy` and reaches the sampler now either arms or refuses; a registry test fails on any new one that does neither.
- **The only critic is `"medoid"`**, a parameter-free consensus selector that keeps the candidate closest to all the others. It exists so best-of-N is testable and benchmarkable end to end before a trained critic does. It requires `n_candidates >= 3` and raises below that — at N=2 the single pairwise distance is shared, every score ties, and selection would always fall back to candidate 0. It is **not** a quality model: on a genuinely multimodal task it prefers the more populated mode rather than the better one. A learned critic is a follow-up.
- **All N candidates share one greedy reasoning trace.** The autoregressive `predict_response` decode is argmax and therefore noise-independent, and the fan-out sits after it. This diversifies flow-matching noise only, not reasoning — so a wrong reasoning trace is wrong in all N chunks. Documented rather than refused.

### Two consequences worth a reviewer's attention

1. **Both new fields are written into every saved `config.json`.** draccus encodes the whole dataclass with no default filtering, so a legacy checkpoint that is fine-tuned, resumed, or converted and re-uploaded gains `n_candidates` and `action_chunk_critic_path` on the way through. A pinned older install reading such a config rejects it on the unknown fields; the checkpoint convert/upload path is where that bites.
2. **`AccelProvenance` gained `n_candidates`, and it is in `COMPARABLE_FIELDS`** — an accel calibration fitted at N=1 now refuses to apply to an N>1 stream, because best-of-N conditions the emitted chunk on a critic. Calibration JSON written before this change still loads and defaults to 1.

Also fixes the `past_key_values` type annotations in pi05/pi06, which claimed `list[dict[str, Tensor]]` for what is a dict keyed by layer index — the expansion helper depends on knowing which.

## How it was tested

**CPU suite** (`pytest -m "not gpu and not network" -n auto`), including four new test modules:

- `tests/policies/test_candidates.py` — 24 tests over the module itself: `(b n)` ordering pinned by exact values (a transposed convention fails it), the einops stride-0 aliasing at B==1 asserted in both directions, NaN/tie/all-non-finite selection, medoid outlier ranking and `row_mask` honouring, and the arming precedence conflict exercised with **both** inputs present.
- `tests/policies/test_candidate_policy_wiring.py` — 43 tests over pi05 and pi06: N=1 never enters the candidate path (patching the per-module bindings, and re-proving each raiser fires at N=2), fused candidate *i* equals a standalone run with that noise for B in {1,2,3}, the prefix pass runs exactly once regardless of N, frozen real-time-chunking rows identical across candidates, cross-rank selection stability, and `last_accel` staying length B.
- `tests/policies/test_candidate_entry_point_wiring.py` — 27 tests; AST-scans `scripts/` and fails on any entry point that builds a policy and reaches the sampler but neither arms nor refuses.
- `tests/policies/test_accel_wiring_registry.py` — extended with registry pins for the sampler parameter position and for exactly-the-wired-policies declaring the attribute.

Tests were **mutation-checked, not just run green**: transposing the fan-out to `(n b)` turns six equivalence and frozen-prefix cases red (B=1 correctly stays green, which is why the equivalence test is parametrized over B); replacing `select_candidate` with a random tie-break breaks the cross-rank pin; removing an entry point from the arm allowlist turns the registry test red.

Full CPU run: **2922 passed**. Six non-passes were traced to the environment, not this change — five were `ModuleNotFoundError: No module named 'libero'` from syncing without the `libero` extra (all pass once it is installed), and `test_utils_benchmark.py::test_context_manager_usage` passes when run serially. One residual failure, `test_libero_utils.py::test_libero2torch`, is a missing LIBERO data asset.

**GPU**, `pytest -m "gpu" -n 0` over the eight affected test files on a single CUDA device: **2 passed, 2 failed**. Both failures are `torch.OutOfMemoryError` loading Gemma3-4B in `tests/policies/test_pi06.py`, on a card that was sharing ~19 GiB with unrelated work. They are **not caused by this change** — verified by stashing the entire diff and re-running, which reproduces the same two OOMs on the baseline. The nightly regression suite runs in CI; it was not run by hand here.

Separately, on the same device with a real trained 6-DoF pi05 checkpoint carrying two norm heads:

- N=1 before vs after the change: bit-identical output digest.
- Full policy path at N=4 through `configure_candidates` -> `sample_actions` -> unnormalize -> critic -> selection: return stays `(B, chunk, action_dim)` with no candidate axis leaking to callers, the critic ranked all four distinctly, and disarming back to N=1 reproduces the original chunk exactly.
- The latency and memory table above.

```bash
pytest -sx tests/policies/test_candidates.py tests/policies/test_candidate_policy_wiring.py
```

```bash
pytest -sx tests/policies/test_candidate_entry_point_wiring.py tests/policies/test_accel_wiring_registry.py
```

Benchmark the fan-out on a GPU box (flags ride on env vars in this script, matching its existing convention; note that 2 is rejected, so sweep 1/4/8/16):

```bash
BENCH_N_CANDIDATES=8 python src/opentau/scripts/benchmark_inference.py --config_path=configs/examples/pi05_training_config.json
```

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,88 @@ The format is loosely based on [Keep a Changelog](https://keepachangelog.com/).
 
 ## [Unreleased]
 
+### Added — best-of-N action-chunk sampling — **opt-in, default `1`, no `config_version` bump**
+
+A flow-matching policy maps one Gaussian draw to exactly one action chunk, deterministically,
+so drawing `n_candidates` noises yields `n_candidates` distinct chunks and an
+`ActionChunkCritic` (`opentau.policies.candidates`) picks the one that reaches the robot.
+
+**Why it is nearly free.** The VLM prefix pass runs once per *observation*, and the KV cache it
+fills is never written again during denoising — so expanding that cache across candidates lets
+one prefix pass serve all N. Only the action expert's Euler loop widens, and at batch 1 that
+loop is occupancy-bound rather than FLOP-bound, so widening it rides largely free. Measured on
+an RTX 5090 (bf16, sdpa, 3 cameras, 1024 prefix tokens, chunk 50, 10 Euler steps, batch 1, a
+randomly-initialized 3.36B-parameter pi05), fused against naive replication of the observation
+across the batch — one prefix pass per candidate, same result:
+
+* **N=1** — 152.9 ms fused, 154.3 ms naive
+* **N=2** — 153.2 / 187.7 ms, i.e. **1.23x**; +0.1% over the fused path's own N=1
+* **N=4** — 154.8 / 256.1 ms, **1.65x**; +1.2%
+* **N=8** — 155.5 / 392.1 ms, **2.52x**; +1.7%
+* **N=16** — 191.2 / 698.2 ms, **3.65x**; +25%
+* **N=32** — 277.8 ms fused; naive is out of memory
+
+Peak memory is nearly flat for the same reason: **6.41 GiB at N=1 → 6.95 GiB at N=32**, the
+dominant per-candidate allocation being the prefix KV cache at **18 MiB** (MQA — one KV head,
+head_dim 256, 18 layers). Naive replication runs 6.42 → 8.54 GiB by N=16 and OOMs at N=32. The
+useful envelope is therefore **N ≤ 8**, where the feature costs under 2% of wall clock; N=16 is
+where the Euler loop stops riding free.
+
+**`n_candidates` defaults to `1`, and stays off unless an entry point arms it.** At `1` no
+critic is loaded, no candidate code runs, and the sampler is the one that shipped before —
+verified at full size on GPU, where N=1 after the change reproduces the pre-change output
+digest bit-for-bit (`5a0bf2ed20abbcc2`). `configure_candidates` is the only writer of
+`policy.n_candidates`, so a config travelling with a checkpoint cannot self-arm best-of-N in a
+script that never opted in, the same posture as `accel_prefix`. The serving entry points arm
+it; `scripts/eval.py` **refuses** `n_candidates > 1` outright rather than multiplying its
+default batch of 16 by N.
+
+**Only pi05 and pi06 are wired.** Every other family — pi0, pi05_mem, both pi07 low levels,
+cosmos3, cosmos3_nano, the two high-level planners and `value` — raises on `n_candidates > 1`
+rather than accepting the field and ignoring it, which is the failure an operator would read as
+"the critic never fires".
+
+**The only critic available is `"medoid"`**, a parameter-free consensus selector that keeps the
+candidate closest to all the others. It exists so best-of-N is testable and benchmarkable end
+to end before a trained critic does. It needs `n_candidates >= 3` and **raises** below that: at
+N=2 the one pairwise distance is shared, so every score ties and selection would always fall
+back to candidate 0 — best-of-N that silently is not. And it
+**is not a quality model**: it cannot tell a confidently-wrong mode from a correct one, and on a
+genuinely multimodal task it prefers the more *populated* mode rather than the better one. A
+learned critic is a follow-up; an `action_chunk_critic_path` pointing anywhere else raises
+rather than silently falling back.
+
+**All N candidates share one greedy reasoning trace.** The autoregressive `predict_response`
+decode is argmax and therefore noise-independent, and the candidate fan-out sits after it
+(expanding earlier would be overwritten by the response loop's own `fill_kv_cache=True`
+writes). This feature diversifies flow-matching noise only, not reasoning. Documented rather
+than refused — but it does bound what best-of-N can recover from: a wrong reasoning trace is
+wrong in all N chunks.
+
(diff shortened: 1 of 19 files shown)


shuheng-liu:
All four findings were correct and are fixed in `fc9325b`. Thanks — the blocking one was a real startup regression.

**blocking, `configure_candidates` wiring probe.** Confirmed exactly as described: `PreTrainedConfig.n_candidates` defaults to `1`, not `None`, so the `requested is None` early return never fires for a real config and the probe ran unconditionally. The probe now sits after the `n == 1` return, and that branch only writes `policy.n_candidates` when the attribute already exists — so it cannot create it on an unwired policy and defeat the probe on a later call, which was the follow-on you flagged.

**suggestion, `MedoidCritic` row_mask broadcast.** Also confirmed. Now `rearrange(row_mask, "b c -> b 1 c 1")` with an explicit `ndim == 2` guard, so the batch axis can no longer land on the candidate axis.

**nit, "and the caller warns".** The claim was false; rather than drop it I made it true. `warn_if_no_candidate_scored` reads the host-side scores the policy already syncs for `last_candidate_scores`, so the all-non-finite fallback is observable at no extra device round trip, logged once per process.

**nit, the always-writes docstring.** The `requested is None` path now folds into the same `n == 1` branch, so the resolved value really is written whenever the policy is wired. The docstring also now states the unwired case explicitly.

On the two "the test cannot fail on this" points — both were right, and one of them caught a second gap. I added `test_an_unwired_family_is_a_noop_at_the_default_n_of_one`, `test_configure_never_creates_n_candidates_on_an_unwired_family`, and `test_medoid_honours_a_per_observation_row_mask`, then mutation-checked each by reverting the fix. The first two go red as expected. The row_mask test initially only went red for `B != N` — my first version reused one mask across the batch, so a transposed axis produced identical numbers and the `B == N` silent case still passed. It now uses a distinct mask per observation and compares against the per-observation `(1, chunk)` path, and both parametrizations go red under the old rearrange.

PR #528, opened 2026-08-20
fix(tests): compare candidate arming by execution order, not file offset

shuheng-liu:
## What this does

Fixes the gating CPU test currently failing on `main`.

`test_candidates_are_armed_before_the_first_sampler_call` (added in #525) checked that
`configure_candidates` runs before the sampler by comparing **raw line numbers**:

```python
assert arm[0].lineno < first_sample   # first_sample = min(lineno) over every .sample_actions(...)
```

File offset stops tracking execution order the moment a sampler call is factored into a
helper. #523 did exactly that — it extracted `_sample_actions_sync` in `grpc/server.py` and
placed it *above* `_load_policy` — so the matcher read a correct server as sampling before it
armed:

```
AssertionError: grpc/server.py: configure_candidates() at line 357 runs after the first
sample_actions()/select_action() call at line 328
```

The runtime order was never wrong. `__init__` calls `_load_policy`, which arms at 357 and only
then warms up at 388-389 via `self._sample_actions_sync(...)`; the call at line 328 lives in
that helper's body and does not execute until 388 calls it.

Neither PR could have caught this alone — #523 introduced the helper, #525 introduced the
test, they merged 2h46m apart, and `main` does not require branches to be up to date before
merging, so the combination was never run.

**The fix:** resolve sampling to its *execution* site. The matcher now finds the function that
arms, computes the set of functions that reach the sampler (to a fixed point, so helper chains
resolve), and counts a call to any of them at its **call site**. It also fails loudly instead
of vacuously if arming and the warmup are ever split across functions, rather than quietly
checking nothing.

No source change — the four serving entry points were already correct.

## How it was tested

`tests/policies/test_candidate_entry_point_wiring.py`: 33 passed (was 30, 1 failing).
Full `tests/policies/` CPU subset: **1220 passed, 2 skipped**.

Six tests added, each constructing a case the matcher could get wrong:

- `test_a_helper_defined_above_the_arming_call_samples_at_its_call_site` — the regression
  itself, in the shape #523 gave `grpc/server.py`.
- `test_arming_after_the_warmup_is_still_caught` — same source, arming moved below the warmup,
  so the fix cannot quietly become "never disagree".
- `test_a_chain_of_helpers_resolves_to_the_outermost_call_site` — pins the fixed-point walk.
- `test_an_arming_function_that_never_samples_is_reported_not_passed` — pins the vacuity guard.
- `test_a_warmup_closure_inside_the_arming_function_samples_at_its_call_site` — the same bug
  one nesting level down, raised in review of this PR: `ast.walk` descends into nested `def`
  bodies, so a warmup factored into a closure *inside* the arming function read as
  `arm=7, sample=5` on correct code.
- `test_a_warmup_closure_called_before_arming_is_still_caught` — the fail-open direction for
  that fix.

Both directions were mutation-verified rather than assumed (CLAUDE.md rule 8 — a pin that
stays green under the reverted bug is not a pin):

| Mutation | Goes red |
|---|---|
| Drop helper resolution (back to file-offset behavior) | the regression pin, the chain pin, and the real `grpc/server.py` case |
| Make the matcher a tautology | the arming-after-warmup pin and the vacuity guard |
| Revert the nested-def scan to `ast.walk` | the nested-closure pin |
| Exclude nested defs from the reaching set (fail-open) | both nested-closure pins |

```bash
gh pr checkout 528 && pytest -sx tests/policies/test_candidate_entry_point_wiring.py
```

Confirm it reproduces on `main` first:

```bash
git checkout main && pytest -sx tests/policies/test_candidate_entry_point_wiring.py -k armed_before
```

## Checklist

### DIFF
--- a/tests/policies/test_candidate_entry_point_wiring.py
+++ b/tests/policies/test_candidate_entry_point_wiring.py
@@ -171,13 +171,42 @@ def _sampler_reaching_functions(tree: ast.AST) -> set[str]:
         reaching = grown
 
 
+def _walk_skipping_nested_defs(fn: ast.AST) -> list[ast.AST]:
+    """Nodes under ``fn``, without descending into nested ``def`` bodies.
+
+    A nested function's body does not run where it is written, so its sampler calls belong
+    to its call site — which :func:`_sampler_reaching_functions` already resolves by name,
+    nested defs included. Descending would put the sampling back at the definition and
+    reintroduce the file-offset bug one level down.
+
+    ``lambda`` bodies *are* descended into: an anonymous function carries no name for that
+    resolution to key on, so treating it as executing in place fails loud rather than open.
+
+    Args:
+        fn: The function whose body is scanned.
+
+    Returns:
+        Every descendant node of ``fn`` except those inside a nested named function.
+    """
+    out: list[ast.AST] = []
+    stack = list(ast.iter_child_nodes(fn))
+    while stack:
+        node = stack.pop()
+        out.append(node)
+        if isinstance(node, ast.FunctionDef | ast.AsyncFunctionDef):
+            continue
+        stack.extend(ast.iter_child_nodes(node))
+    return out
+
+
 def _first_sampling_line_within(fn: ast.AST, sampler_reaching: set[str]) -> int | None:
     """Line inside ``fn`` at which sampling first actually *happens*.
 
     A ``sample_actions(...)`` written in a helper does not execute where it is written —
     it executes where that helper is **called**. So a call to a sampler-reaching helper
     counts at its call site, which is the whole difference between this and comparing raw
-    file offsets.
+    file offsets. That holds for helpers nested inside ``fn`` too, hence
+    :func:`_walk_skipping_nested_defs`.
 
     Args:
         fn: The function whose body is scanned.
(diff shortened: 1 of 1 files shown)


shuheng-liu:
Good catch — confirmed and fixed in 7d1e82d.

I reproduced your case before changing anything, and it lands exactly where you said:

```
arm=7  first_sampling=5  -> assertion arm<first is False
```

`ast.walk(fn)` descends into nested `def` bodies, so a warmup factored into a closure inside the arming function put the sampling back at the closure's *definition* — the same file-offset bug this PR is named after, one nesting level down.

The direct scan now goes through `_walk_skipping_nested_defs`, which stops at nested named defs. Their call sites already resolve through `_sampler_reaching_functions` (which includes nested defs), so nothing is lost. One deliberate asymmetry: `lambda` bodies are still descended into, because an anonymous function carries no name for that resolution to key on — treating it as executing in place fails loud rather than open.

Pinned in both directions, and both pins are mutation-verified:

| Mutation | Goes red |
|---|---|
| Revert the scan to `ast.walk` | `test_a_warmup_closure_inside_the_arming_function_samples_at_its_call_site` |
| Exclude nested defs from the reaching set (fail-open) | both nested-closure pins |

Worth noting my first attempt at that second mutation was a no-op — it removed a node that was never counted, so the suite stayed green and briefly looked like the pin was dead. The real fail-open lever is the reaching set, not the walker.

`tests/policies/`: 1220 passed, 2 skipped.

PR #552, opened 2026-09-11
feat(xr1): port the Xiaomi-Robotics-1 policy (Qwen3-VL-4B + 36-layer DiT)

shuheng-liu:
## What this does

Adds **`xr1`**, an OpenTau port of [Xiaomi-Robotics-1](https://huggingface.co/XiaomiRobotics/Xiaomi-Robotics-1-RoboCasa365) (Apache-2.0) — the top entry on the RoboCasa365 leaderboard. A Qwen3-VL-4B vision-language backbone whose per-layer key/value cache is read by a 36-layer DiT flow-matching action head, one DiT layer per VLM layer.

The port reuses the structure `cosmos3` established (stock `transformers` backbone as a black box, only its KV cache crossing into a hand-written head), so nothing about the 4.4B backbone is reimplemented. What is genuinely different from every existing policy here, and what most of the new code exists to get right:

- **Observation history as video.** Four frames per camera at stride 2 reach the backbone as three two-frame *videos*, not stills. The prompt labels cameras positionally — Left / Right / **Wrist last** — so camera order is load-bearing.
- **Ascending flow time.** The Euler loop integrates τ 0 → 1 with `dt = +1/num_steps`. π₀.₅ and cosmos3 both descend; copy-pasting either sampler inverts the flow silently.
- **Identity normalization throughout.** The reference's RoboCasa365 statistics are mean 0 / std 1, and the state adapter consumes *raw* quaternions — normalizing one before the axis-angle conversion is silently catastrophic, so the config refuses any other mapping.

**Parity is bit-identical, not approximate.** Against the reference checkpoint, `torch.equal` holds for the tokenized prompt, all 72 prefix key/value tensors, every Euler-step velocity, the full decoded chunk, and a hash over all 1121 parameters; a 16-chunk open-loop replay stays within p99 2e-3.

In simulation the five-rung parity ladder measured (ours / published):

| Task | Ours | Published | Accept (Wilson 95%) |
|---|---|---|---|
| CloseFridge, n=10 | 10/10 | 94 % | ≥ 5/10 |
| CloseFridge, n=50 | 98 % | 94 % | 42–50 |
| TurnOnMicrowave, n=50 | 48 % | 56 % | 21–34 |
| OpenDrawer, n=50 | 96 % | 94 % | 42–50 |
| CloseBlenderLid, n=50 | 40 % | 36 % | 12–25 |

All five inside their intervals, deviations +4 / +4 / −8 / +2 / +4 pp — no directional bias.

Also ships two example configs and the golden fixtures + parity ladder recipe under `tests/artifacts/policies/xr1/`.

**Depends on #550**, which corrected the RoboCasa flat action layout. That fix was found *by* this port and landed separately so it could be reviewed on its own; this branch is rebased on it and is now purely additive — new `xr1` files plus one `CHANGELOG` section.

### Two traps worth knowing if you extend this

- The RoboCasa365 datasets store `observation.state` **EE-first** while the simulator's `agent_pos` is **base-first**. Both are 16 wide and both carry two unit quaternions, so the wrong one yields a plausible pose and no error. `state_adapter` names which layout a config reads.
- `F.scaled_dot_product_attention` is **not rank-invariant** — a `(B, 1, H, L, D)` call and a `(B, H, L, D)` call select different kernels and disagree bitwise. The training-time `training_repeat` axis therefore folds into the batch (the reference's own scheme) rather than becoming an extra dim, or training and inference would sit on different numerics.

## How it was tested

**Reference parity (GPU, against the real checkpoint).** `tests/policies/test_xr1_gpu.py` — 41 tests. The default gate is `torch.equal`, not a tolerance; where a tolerance appears the docstring names the operation responsible. Covers the strict-coverage state-dict load, per-parameter weight identity, the prefix KV cache, per-Euler-step velocities, the full chunk, and a replay trace. Run on a GPU dev box against the real checkpoint: **40 passed, 1 skipped** — the skip is a 40 GiB-gated full-finetune gradient test declining a 24 GiB card, which is the gate working. They read a local reference checkpoint and a heavy-fixture bundle via `XR1_REFERENCE_CHECKPOINT` / `XR1_FIXTURE_DIR`, and **skip loudly** when those are unset or when the torch/CUDA/device triple differs from the capture — bit-identity is kernel-dependent, so loosening to a tolerance there would silently turn the ladder into a rounding check.

**CPU suite.** `tests/policies/test_xr1_cpu.py` — 98 tests on a tiny random `Qwen3VLConfig`, no network, no downloads. Beyond shapes, these pin the specific ways a plausible implementation goes silently wrong: the flow integrating the wrong way, the action mask hoisted out of the Euler loop, a bidirectional instead of causal DiT query block, the MRoPE offset collapsed across axes, history zero-padded instead of clamped, `use_cache` silently disabled by gradient checkpointing. Several are rule-8 conflict tests that construct the case that would break the claim — e.g. the timestep sinusoid is checked against the cosmos3 helper it must *not* equal.

**Sim.** The five-rung ladder above, run on the reference's own per-episode scene seeds via `eval.seed_list` so both implementations roll identical scenes.

**Determinism.** Ten seeded optimizer steps, twice, compared bit-for-bit. The shape-static async prefix is what makes that check meaningful — the reference draws one Python-level prefix length per batch and slices the DiT query, which would make sequence length a per-rank random variable.

**Gates.** `pytest -m "not gpu and not network" -n auto` → 3412 passed; the only failures are pre-existing ones from LIBERO not being installed locally. `pre-commit run --all-files` clean.

```bash
git checkout claude/sleepy-babbage-b29fd7
```

The CPU suite needs nothing but the repo — no GPU, no network, no checkpoint:

```bash
pytest -sx tests/policies/test_xr1_cpu.py
```

The registry/AST sweeps that a new policy has to satisfy:

```bash
pytest -sx tests/policies/test_accel_wiring_registry.py \
          tests/policies/test_state_action_representation_only_registry.py \
          tests/policies/test_from_pretrained_revision_overrides.py \
          tests/test_available.py
```

Reference-parity gates need a local copy of the reference checkpoint plus the captured fixtures (see `tests/artifacts/policies/xr1/README.md`, which documents the capture script and the ladder):

```bash
XR1_REFERENCE_CHECKPOINT=<path to Xiaomi-Robotics-1-RoboCasa365> \
XR1_FIXTURE_DIR=<path to captured fixtures> \
pytest -m gpu -n 0 tests/policies/test_xr1_gpu.py
```

## Checklist


shuheng-liu:
Thanks — four of the five are fixed in the commit above. Pushing back on one.

**The action-queue eviction (`modeling_xr1.py:985`) is not reachable.** The scenario needs `n_action_steps <= chunk_size - delay`, i.e. `n_action_steps < chunk_size` while a delay is in play — and `PreTrainedConfig.validate_action_horizon` refuses exactly that pair:

> A shortened execution horizon (n_action_steps < chunk_size) is not yet supported together with a non-zero `max_delay`

So whenever a delay exists, `n_action_steps == chunk_size`, and the refill slice `actions[d : d + n_action_steps]` yields `chunk_size - d` rows for the `d` free slots. It fits exactly at every reachable delay:

| runtime delay | queued | appended | total | maxlen | evicted |
|---|---|---|---|---|---|
| 0 | 0 | 16 | 16 | 16 | 0 |
| 1 | 1 | 15 | 16 | 16 | 0 |
| 2 | 2 | 14 | 16 | 16 | 0 |
| 3 | 3 | 13 | 16 | 16 | 0 |
| 4 | 4 | 12 | 16 | 16 | 0 |

You were right that the path had **no test**, though, so I added `test_replanning_with_a_frozen_prefix_never_evicts_a_queued_action`. It asserts both halves — that the validator refuses the shortened-horizon pair, and that the arithmetic fits at every delay — because the invariant this rests on lives in a *different file*, so a future relaxation of that validator would make the queue silently wrong here with nothing failing.

On the three suggestions/nits that were right:

- **Choice-head state pooling** — real, and worse than reported: the reference's `state_shape` is `(1, 60)`, so its choice head conditions on a *single* row, not a pooled window. Now takes the last frame.
- **`freq_loss_weight`** — the config default (0.5) was the wrong one; the reference's `freq_coefficient` is 1.0. Fixed the default rather than the config, so the fine-tune example no longer needs an override at all.
- **`build_history_batch`** — correct, that name is a per-policy method, not the shared helper. Docstring now names `history_slot_indices` and says why the two differ.

